PR #16, opened 2022-03-24
ENH: photometric redshift model

ntessore:
This PR adds a generator `glass.galaxies.gal_phz_gausserr` for photometric galaxy redshifts. It uses the same simple Gaussian error model as `glass.observations.tomo_nz_gausserr`.

An example is provided as well.


arthurmloureiro (approved):
Both documentation and example ran.

Definitions all correct, all good to go!

PR #17, opened 2022-03-25
ENH: generators groups

ntessore:
This PR adds the possibility to group generators. This is useful if multiple populations of objects run similar generators, since it removes the need to manually specify distinct input and output names. Defining two groups:

```py
generators = [
    ...,
    glass.sim.group('group-a', [
        ...,  # generators in group-a
    ]),
    glass.sim.group('group-b', [
        ...,  # generators in group-b
    ]),
    ...,
]
```

The group names provide a nested namespace for the same outputs, which would otherwise be overwritten (unless their names were changed manually):

```py
for it in glass.sim.generate(generators):
    it['group-a'][...]  # nested namespace for group-a
    it['group-b'][...]  # nested namespace for group-b
```

An example is provided.


examples/1_basic/groups.py
```diff
@@ -0,0 +1,78 @@
...
+# %%
+# Plots
+# -----
+# Plot the two distributions together with the expected inputs.
+
+norm = glass.util.ARCMIN2_SPHERE*(z[-1] - z[0])/40
+
+for zz, nz in (low_z, dndz_low), (high_z, dndz_high):
+    plt.hist(zz, bins=40, range=(z[0], z[-1]), histtype='stepfilled', alpha=0.5)
+    plt.plot(z, norm*nz, '-k', lw=1, alpha=0.5)
```
arthurmloureiro, line 75: I'd add some legend to the plot :)

arthurmloureiro (commented):
Just one suggestion to add labels to the example plot :)

arthurmloureiro (approved):
Perfect!

PR #23, opened 2022-08-10
ENH: shells cover full redshift range

ntessore:
This PR changes the shell definition `zspace` and `xspace` such that when `dz` or `dx` are specified, the resulting shells **cover** the given redshift range.  Before, they would work like numpy's `arange` in that the shells are **included** in the given redshift range.

For example, when using `xspace(c, 0., 2., dx=200)` the last shell has an upper redshift of 2.013991, thus covering the entire 0 to 2 range.

Tagging reviewers so that they can let me know if this is a problem in their applications.

### DIFF
--- a/glass/sim.py
+++ b/glass/sim.py
@@ -40,7 +40,7 @@ def zspace(zmin, zmax, *, dz=None, num=None):
     if (dz is None) == (num is None):
         raise ValueError('exactly one of "dz" or "num" must be given')
     if dz is not None:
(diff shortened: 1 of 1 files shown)


arthurmloureiro (approved):
Oh, good fix!

I always assumed that if I requested a range of z=[0., 2.8] that the bin ending at 2.8 would be included. 

I tested here and it's all good to go from my side.

PR #24, opened 2022-08-23
ENH: intrinsic normal galaxy ellipticity distribution

ntessore:
This PR adds the intrinsic normal distribution for galaxy ellipticity. It is described in the upcoming GLASS paper.


arthurmloureiro (approved):
All good from this side :D

PR #29, opened 2022-09-29
ENH: support for matter weight function

ntessore:
This PR introduces support for user-defined matter weight functions. A number of schemes are provided in `glass.matter`, and it is easy to add new ones.

The matter weight is given for each shell as a tuple of redshifts and unnormalised weights, stored as `wz`.

All examples have been updated.


arthurmloureiro (commented):
Awesome!

PR #31, opened 2022-09-29
DOC: move examples to separate repository

### LINKED ISSUE #30: move examples to own project
The sphinx-gallery examples in GLASS are more difficult than elsewhere because they are computationally expensive.  Therefore, they are built and committed to the repo. This adds a lot of unnecessary volume to PRs, such as in #29.

I propose we move the examples to their own `glass-examples` project.  That way, we can build them asynchronously whenever necessary, and commits don't clutter the main repository.

Using the subprojects feature on ReadTheDocs, we should still be able to serve the examples under the GLASS documentation.

ntessore:
Closes #30 by moving examples out of the repository. I have kept a (now external) link to the examples in the header bar, and (unrelated to this PR) highlighted it for better visibility.

Linking to examples in the GLASS documentation is easy using the intersphinx `` :doc:`examples:basic/photoz` `` syntax.

A preview will be generated in the actions below.


ntessore:
NB I want to get this out of the way before making some of the big changes because it decouples the examples from the library.

PR #45, opened 2023-02-04
TYP: add type hints to library functions

ntessore:
Add type hints to library functions. These are all fairly concrete and not (yet) ready for full static type checker support, which will probably require some thought about dealing with numpy and scalar types.

The PR is still work in progress:

- [X] fields.py
- [X] galaxies.py
- [X] lensing.py
- [X] matter.py
- [X] observations.py
- [ ] points.py
- [ ] shapes.py
- [ ] user.py


ntessore:
Merging this even though some modules are still missing, so that other things can happen in the meantime.

PR #68, opened 2023-03-03
API: universal support for array types

ntessore:
This change introduces a `glass.array` module which can be imported as the array namespace:

    import glass.array as xp

By default, that loads the numpy array namespace, so that `xp` is mostly equivalent to the usual `np`.

However, the `glass.array` module has a function `use` for switching to a different array namespace internally:

    xp.use('numpy')  # default
    xp.use('cupy')
    xp.use('torch')

This can also be done locally using the context manager-form of the same command:

    with xp.use('torch'):
        ...

Internally, the `array-api-compat` package is used to load the various array namespaces into modules `glass._array_api.<name>`.  These modules can be extended with functionality for GLASS, such as e.g. a spherical harmonic transform that supports the array type.

The `glass._array_api` module should also be a namespace package, so that we can have extensions which add support for other array types, e.g. `glass-jax`.

### DIFF
--- a/glass/_array_api/cupy.py
+++ b/glass/_array_api/cupy.py
@@ -1,0 +1,3 @@
+'''cupy array namespace for GLASS'''
+
+from array_api_compat.cupy import *  # noqa
--- a/glass/_array_api/numpy.py
+++ b/glass/_array_api/numpy.py
@@ -1,0 +1,3 @@
+'''numpy array namespace for GLASS'''
+
+from array_api_compat.numpy import *  # noqa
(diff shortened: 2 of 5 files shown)


ntessore:
Implementing a couple of functions just now, I don't even think we need the global state of the array namespace at all. I was mainly worried about situations where we create new arrays and wouldn't know which type to use. But the only situations where there isn't an array input seem to involve random sampling, where we have the array type implicitly via the `rng`.

ntessore:
Closing this in favour of a more comprehensive compatibility layer.

Issue #98, opened 2023-06-10
Iterative sampling of points


ntessore:
At the moment, the point sampling in `glass.points` is done by constructing a single output array. This can become huge for serious simulations, even if shells are thin.

However, there is no reason to create this array in memory; in fact, it is a huge waste of time. Internally, the sampling is already iterative: it processes batches of pixels, and builds up the (huge) output array piece by piece.

I suggest to change the `positions_from_delta()` and `uniform_positions()` functions to instead return an iterator over `lon, lat, count`, in the same format as not. Every subsequent function would stay exactly the same. The only difference is that this:

```py
lon, lat, count = positions_from_delta(...)
...
```

would become this:
```py
for lon, lat, count in positions_from_delta(...):
    ...
```

where now every `lon`, `lat` has at most a decent size (e.g. 1 million rows).

PR #99, opened 2023-06-10
API(points): iterative sampling of positions

ntessore:
Changes the point sampling functions `positions_from_delta()` and `uniform_positions()` to return an iterator over batches of samples.

Closes: #98
Changed: The point sampling functions `positions_from_delta()` and `uniform_positions()` now return an iterator


ntessore:
Holding off on merging until the examples are updated as well.

ntessore:
The examples were updated by @arthurmloureiro, merging this.

Issue #118, opened 2023-07-21
Triangular radial window functions


ntessore:
Implement the triangular window function that corresponds to linear interpolation.

PR #119, opened 2023-07-30
ENH(shells): radial windows for linear and cubic interpolation

ntessore:
Adds window functions `linear_windows()` and `cubic_windows()` that correspond to linear and cubic spline interpolation, respectively.

Documentation is updated to give an overview of available window functions.

Closes: #118 
Added: New `linear_windows()` and `cubic_windows()` window functions for shells.


ntessore:
Plot of the new window functions from the docs:
image

Issue #156, opened 2024-03-26
Add FITS catalogue writer tool


ntessore:
Add the async FITS catalogue writer as a user tool.

PR #158, opened 2024-04-30
ENH(user): Add FITS catalogue writer tool #156 

ucapbba:
Update the user functions with a new `write_catalog` method that allows a user to easily write a FITS catalogue.

Testing against Python 3.7 removed as it does not accept positional arguments with `/`, which are used in the "write" function.

Closes: #156

### DIFF
--- a/glass/shells.py
+++ b/glass/shells.py
@@ -6,7 +6,7 @@ Shells (:mod:`glass.shells`)
 
 .. currentmodule:: glass.shells
 
-The :mod:`glass.shells` module provides functions for the definition of
+The :mod:`glass.shells` super awesome module provides functions for the definition of
 matter shells, i.e. the radial discretisation of the light cone.
 
 
--- a/glass/shells.py
+++ b/glass/shells.py
@@ -6,7 +6,7 @@ Shells (:mod:`glass.shells`)
 
 .. currentmodule:: glass.shells
 
-The :mod:`glass.shells` super awesome module provides functions for the definition of
+The :mod:`glass.shells` module provides functions for the definition of
 matter shells, i.e. the radial discretisation of the light cone.
 
 
--- a/glass/test/test_fits.py
+++ b/glass/test/test_fits.py
@@ -11,9 +11,8 @@ import glass.user as user
 import numpy as np
 
 
-@pytest.mark.skipif(not HAVE_FITSIO, reason="test requires fitsio")
 def _test_append(fits, data, names):
-    """Write routine for FITS data."""
+    '''Write routine for FITS test cases'''
     cat_name = 'CATALOG'
     if cat_name not in fits:
         fits.write_table(data, names=names, extname=cat_name)
@@ -36,7 +35,7 @@ def test_basic_write(tmp_path):
     filename_gfits = "gfits.fits"  # what GLASS creates
     filename_tfits = "tfits.fits"  # file create on the fly to test against
 
-    with user.write_context(d / filename_gfits, ext="CATALOG") as out, fitsio.FITS(d / filename_tfits, "rw", clobber=True) as myFits:
+    with user.write_catalog(d / filename_gfits, ext="CATALOG") as out, fitsio.FITS(d / filename_tfits, "rw", clobber=True) as myFits:
         for i in range(0, my_max):
             array = np.arange(i, i+1, delta)  # array of size 1/delta
             array2 = np.arange(i+1, i+2, delta)  # array of size 1/delta
@@ -59,7 +58,7 @@ def test_write_exception(tmp_path):
     d.mkdir()
 
     try:
-        with user.write_context(d / filename, ext="CATALOG") as out:
+        with user.write_catalog(d / filename, ext="CATALOG") as out:
             for i in range(0, my_max):
                 if i == except_int:
                     raise Exception("Unhandled exception")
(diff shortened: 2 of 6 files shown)


glass/test/test_fits.py
```diff
@@ -0,0 +1,110 @@
+import pytest
+
+# check if fitsio is available for testing
+try:
+    import fitsio
+except ImportError:
+    HAVE_FITSIO = False
+else:
+    del fitsio
+    HAVE_FITSIO = True
```
ntessore, line 10:
```suggestion
# check if fitsio is available for testing
import importlib.util
if importlib.util.find_spec("fitsio") is not None:
    HAVE_FITSIO = True
else:
    HAVE_FITSIO = False
```

glass/test/test_fits.py
```diff
@@ -0,0 +1,110 @@
...
+
+import glass.user as user
+import numpy as np
+
+
+@pytest.mark.skipif(not HAVE_FITSIO, reason="test requires fitsio")
```
ntessore, line 16: not a test

glass/test/test_fits.py
```diff
@@ -0,0 +1,110 @@
...
+    if cat_name not in fits:
+        fits.write_table(data, names=names, extname=cat_name)
+    else:
+        hdu = fits[cat_name]
+        hdu.write(data, names=names, firstrow=hdu.get_nrows())
+
+
+delta = 0.001  # Number of points in arrays
+myMax = 1000  # Typically number of galaxies in loop
+exceptInt = 750  # Where test exception occurs in loop
```
ntessore, line 29: snake case

glass/test/test_fits.py
```diff
@@ -0,0 +1,110 @@
...
+            arrays = [array, array2]
+            names = ['RA', 'RB']
+            _test_append(myFits, arrays, names)
+
+    from astropy.io import fits
+    with fits.open(d / filename_gfits) as g_fits, fits.open(d / filename_tfits) as t_fits:
+        glass_data = g_fits[1].data
+        test_data = t_fits[1].data
+        assert glass_data['RA'].size == test_data['RA'].size
+        assert glass_data['RB'].size == test_data['RA'].size
```
ntessore, line 55: Why go through two files, instead of creating an array of data, writing it in batches, then reading and comparing after?

ntessore: Also, we should use fitsio consistently here, not astropy.fits

glass/test/test_fits.py
```diff
@@ -0,0 +1,110 @@
...
+        out.write()
+    assert 1 == 1
+
+
+@pytest.mark.skipif(not HAVE_FITSIO, reason="test requires fitsio")
+def test_write_yield(tmp_path):
+    d = tmp_path / "sub"
+    d.mkdir()
+    with user.write_context(d / filename, ext="CATALOG") as out:
+        assert type(out) is user.FitsWriter
```
ntessore, line 110: this is testing unimportant implementation details

glass/user.py
```diff
@@ -45,3 +51,52 @@ def load_cls(filename):
         values = npz['values']
         split = npz['split']
     return np.split(values, split)
+
+
+class FitsWriter:
```
ntessore, line 56: let's mark this class as internal with a leading underscore

glass/user.py
```diff
@@ -45,3 +51,52 @@ def load_cls(filename):
...
+            self._append(data)
+
+        # if keyword arguments are given, treat them as names and columns
+        if columns:
+            names, values = list(columns.keys()), list(columns.values())
+            self._append(values, names)
+
+
+@contextmanager
+def write_context(filename, *, ext=None):
```
ntessore, line 92: we could call this function `fits_catalog()` or similar

ntessore (changes requested):

glass/shells.py
```diff
@@ -6,7 +6,7 @@
 
 .. currentmodule:: glass.shells
 
-The :mod:`glass.shells` module provides functions for the definition of
```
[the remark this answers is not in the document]
ucapbba: Ha, sorry, I was testing the documentation and this test ended up in my commit!

glass/test/test_fits.py
ucapbba: Maybe we can discuss this.  As we're using fitsio for our new functionality, I thought it would be a good test to use a different API to read the written files and compare them.  I also don't see how the suggested alternative implementation improves the test case.  We can check each data point easily enough if required.

glass/user.py
```diff
@@ -10,15 +10,17 @@
 library.
 
 
-Input and output
+Input and Output
 ----------------
 
 .. autofunction:: save_cls
 .. autofunction:: load_cls
+.. autofunction:: write_context
```
ntessore, line 18: Needs to reflect new name.

glass/user.py
```diff
@@ -45,3 +47,52 @@ def load_cls(filename):
...
+            self._append(values, names)
+
+
+@contextmanager
+def write_catalog(filename, *, ext=None):
+    '''Context manager for a FITS catalogue writer.  Calls class FitsWriter.
+
+    ext is the name of the HDU extension
+
+    '''
```
ntessore, line 93: I was going to merge this, but then I saw this in the rendered docs. Could we create a slightly more user-friendly docstring?

```suggestion
    """
    Write a catalogue into a FITS file, where *ext* is the optional
    name of the extension.  To be used as a context manager::

        # create the catalogue writer
        with write_catalog("catalog.fits") as out:
            ...
            # write catalogue columns RA, DEC, E1, E2, WHT with given arrays
            out.write(RA=lon, DEC=lat, E1=eps1, E2=e2, WHT=w)

    """
```

ntessore (approved):

PR #190, opened 2024-09-11
gh-206: tidy up and apply `toml-sort` to `pyproject.toml`

paddyroddy:
Add `toml-sort` to `pre-commit`. https://github.com/pappasam/toml-sort is a tool to alphabetically sort TOML files. IMO this makes things more readable and removes the thought process of where to put something.

I also like the syntax where each `[table]` has its own section, i.e. `[project]`, `[tool.hatch]` etc.

Reviewed-by: Nicolas Tessore


Saransh-cpp (approved):

ntessore (approved):
Not a fan of removing `[project.urls]` and `[project.optional-dependencies]` in favour of a flat table but presumably that choice was made for us by toml-sort.

paddyroddy:
> Not a fan of removing `[project.urls]` and `[project.optional-dependencies]` in favour of a flat table but presumably that choice was made for us by toml-sort.

It actually wasn't, it doesn't force that. I've reinstated them as separate things. But the `[tool.hatch.build.hooks.vcs]` being separate to `[tool.hatch]` I've combined them. To me, it makes it a lot more obvious what is grouped together. `[project]` has to be there anyway, so I don't mind that one so much.

PR #194, opened 2024-09-11
gh-184: Change copyright range in documentation to be automated

### LINKED ISSUE #184: Expand copyright range in documentation etc.
Can even automate for it to update like this: https://github.com/paddyroddy/.github/tree/main/actions/licence

paddyroddy:
Rather than hardcoding the year, we can compute it as we go. I've also removed the duplication of the `author` field.

I've also manually updated the year range in the licence file, I have an action we could use to update that if we want https://github.com/paddyroddy/.github/tree/main/actions/licence

Fixes #184


Saransh-cpp (approved):

docs/conf.py
```diff
@@ -7,11 +7,14 @@
...
 
+import datetime
+import pytz
 from importlib import metadata
 
 project = 'GLASS'
-copyright = '2022-2024 Nicolas Tessore'
+year = datetime.datetime.now(tz=pytz.timezone("GMT")).year
 author = 'Nicolas Tessore'
+copyright = f'2022-{year} {author}'
```
ntessore, line 17:
```suggestion
author = 'GLASS developers'
copyright = f'2022-{datetime.date.today().year} {author}'
```

ntessore (commented):
Nice idea! How about the suggestion without `pytz`? Since GLASS has a strict "no work on new year's eve" policy, it shouldn't ever make a difference.

docs/conf.py
paddyroddy: Sure, I like that policy!

Issue #189, opened 2024-09-11
Change dependencies to `~-` syntax rather than `>=`


paddyroddy:
My concern is for example `NumPy` now has v2 (#163), the current syntax would allow that when the code doesn't necessarily support it yet.

Saransh-cpp:
Would make sense to do this for NumPy, but I'd prefer not doing it for other dependencies - [Should You Use Upper Bound Version Constraints?](https://iscinumpy.dev/post/bound-version-constraints/)

paddyroddy:
Addressed in #207, can close for now as this issue itself isn't that important

Issue #163, opened 2024-06-25
NumPy 2 compatibility (+ drop `3.9` support)


ntessore:
Ensure compatibility with the NumPy 2 release.

paddyroddy:
Going to start this with the help of https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy

paddyroddy:
- [ ] when done can remove the associated `# type: ignore[attr-defined]` comments for #308

paddyroddy:
Have updated the title to reflect the fact that we will need to drop `3.9` to support `NumPy v2.1`

PR #195, opened 2024-09-11
MNT: pin NumPy to v1 (until v2 is supported)

paddyroddy:
Fixes (the necessary) bit of #189.

Can adjust once #163 is sorted.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "numpy>=1.20.0",
+    "numpy~=1.0",
     "healpix>=2022.11.1",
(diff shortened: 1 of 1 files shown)


ntessore:
Hi, just catching up. The code seems to run fine with numpy v2 beyond the deprecation warning due to trapz vs trapezoid. Can we fix that import instead of excluding anyone who's running with numpy v2, such as me 😄

paddyroddy:
> Hi, just catching up. The code seems to run fine with numpy v2 beyond the deprecation warning due to trapz vs trapezoid. Can we fix that import instead of excluding anyone who's running with numpy v2, such as me 😄

Sure. I ran with the tests locally with `v2` and it passed, but wasn't sure if that was due to the tests being non-exhaustive. Will close this.

PR #199, opened 2024-09-11
STY: add `ruff-format` and remove `flake8`

paddyroddy:
The maintainer of `flake8` can be a bit difficult, so nice to remove. I will add `ruff` the linter in a separate PR, but would be nice to get formatting done first. `ruff-format` is basically faster `black` (but some subtle differences).


.github/workflows/test.yml
```diff
@@ -10,12 +10,13 @@ on:
...
-    steps:
-    - uses: actions/checkout@v4
-    - run: pipx run flake8 glass
+# TODO: replace with a more general pre-commit linter
+#   style:
+#     name: Style
+#     runs-on: ubuntu-latest
+#     steps:
+#     - uses: actions/checkout@v4
+#     - run: pipx run flake8 glass
```
Saransh-cpp, line 19: Unrelated to this PR, but maybe we can add pre-commit.ci to the repository?

Saransh-cpp (changes requested):
This will mess up [`git blame`](https://git-scm.com/docs/git-blame)'s output. A good practice is to add the commit where `ruff` made the formatting changes to a new `.git-blame-ignore-revs` file. For example, PyBaMM's `.git-blame-ignore-revs` file - https://github.com/pybamm-team/PyBaMM/blob/develop/.git-blame-ignore-revs

.github/workflows/test.yml
paddyroddy: The problem is the changes invalidate `flake8` so it. kind of had to go

paddyroddy: I've never been a fan of the CI as it only works in public repos, but seeing as its public I guess we could

paddyroddy:
> This will mess up [`git blame`](https://git-scm.com/docs/git-blame)'s output. A good practice is to add the commit where `ruff` made the formatting changes to a new `.git-blame-ignore-revs` file. For example, PyBaMM's `.git-blame-ignore-revs` file - [pybamm-team/PyBaMM@`develop`/.git-blame-ignore-revs](https://github.com/pybamm-team/PyBaMM/blob/develop/.git-blame-ignore-revs?rgh-link-date=2024-09-11T17%3A29%3A09Z)

Wasn't aware of this idea to combat `git blame`, have added the file

.github/workflows/test.yml
Saransh-cpp: Also could use the pre-commit or the pre-commit-lite action here.

Saransh-cpp (approved):
Thanks!

.github/workflows/test.yml
paddyroddy: Will do in a different PR

ntessore:
The commit where ruff made the changes will get squashed, so we should add that file not here but in a PR immediately after.

Also, why not add the linter in this PR as well, so we only incur one massive changeset?

Saransh-cpp:
> The commit where ruff made the changes will get squashed, so we should add that file not here but in a PR immediately after.

Ah, that too. This makes more sense then.

paddyroddy:
> The commit where ruff made the changes will get squashed, so we should add that file not here but in a PR immediately after.

Yeah I wasn't sure on this

paddyroddy:
> Also, why not add the linter in this PR as well, so we only incur one massive changeset?

Because formatting changes and linting changes are different. I think it will hide a lot more changes that way.

ntessore (approved):

paddyroddy:
I have updated from `main`. I am going to merge, and then update the commit for the `git blame` file

paddyroddy:
Have merged directly https://github.com/glass-dev/glass/commit/e01b40ec11278be4aa02d52c9ad76ce09bc19a6a

Issue #203, opened 2024-09-12
rm pull request review action


Saransh-cpp:
https://github.com/glass-dev/glass/blob/main/.github/workflows/pull_request_review.yml

From slack:
> I need to kill that action off because it cannot work with PRs coming from forks, so people are justly complaining a lot about it

PR #205, opened 2024-09-12
gh-203: remove PR review action

Saransh-cpp:
The PR review action does not work with forks and is a bit redundant, given that commits are squashed and the squashed commit directly references the PR it originates from.

Closes: #203
Removed: PR review action has been discarded


paddyroddy:
FRI can link issues manually like this:
image
Not sure if the `gh-xxx` syntax alone is sufficient

ntessore (approved):

Issue #198, opened 2024-09-11
Should we drop `3.8-3.9` support?


paddyroddy:
According to [SPEC0](https://scientific-python.org/specs/spec-0000) it currently recommends dropping everything below `3.10`, [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) agrees similarly. Can read more about [the motivation here](https://scientific-python.org/specs/spec-0000/#motivation). Any good reason to support `3.8-3.9`?

paddyroddy:
> should only be dropped

I feel this is personal choice, many in the scientific community are leaning towards more aggressive dropping old Python support. We had this discussion in our Python Tooling recently https://github.com/UCL-ARC/python-tooling/issues/362.

paddyroddy:
Looks like we _have_ to drop `3.8` for NumPy v2 migration https://github.com/glass-dev/glass/actions/runs/10833030831/job/30058811483?pr=207

paddyroddy:
Think we have to drop `3.9` too https://github.com/numpy/numpy/releases/tag/v2.1.1

PR #207, opened 2024-09-12
gh-163: full support of `NumPy v2`

### LINKED ISSUE #163: NumPy 2 compatibility (+ drop `3.9` support)

### LINKED ISSUE #189: Change dependencies to `~-` syntax rather than `>=`

paddyroddy:
Fixes #163. Think the only change is required is `np.trapz` to `np.trapezoid`, verified with [NPY201](https://docs.astral.sh/ruff/rules/numpy2-deprecation). Also fixes #189.


paddyroddy:
Have had to drop `3.8-3.9` support for full v2 support. Fixes #198.

Saransh-cpp (commented):
Looks good! But should this be 2 PRs? Or at least should be 2 separate CHANGELOG notes (dropping Python versions should go in breaking changes?)?

paddyroddy:
> Looks good! But should this be 2 PRs? Or at least should be 2 separate CHANGELOG notes (dropping Python versions should go in breaking changes?)?

Yeah I guess the Python changes (#198) should be done first, then NumPy. I hadn't realised they were necessary at the time.

ntessore:
AFAICT, this would make numpy v2 _required_, not merely compatible without deprecations. I find that a bit of a strong requirement when the code works perfectly well with numpy<2 as well. Why not conditionally import `trapezoid` or `trapz as trapezoid` somewhere in our utility modules, then use that?

paddyroddy:
> AFAICT, this would make numpy v2 _required_, not merely compatible without deprecations. I find that a bit of a strong requirement when the code works perfectly well with numpy<2 as well. Why not conditionally import `trapezoid` or `trapz as trapezoid` somewhere in our utility modules, then use that?

Yep, I had thought about that. I think this will become a piece of frustrating maintenance burden to be both API compatible, whilst still maintaining `numpy<2`. What do you think? Happy to do as you suggest, but I think it will quickly become a bit of a barrier

ntessore:
I would probably keep it as long as it's one single change.

To be evaluated again if we start implementing the Array API, since NumPy 2 should be compatible with that out of the box, so we could forego the compatibility layer. Not sure if that's desirable, though, so probably something to discuss soon.

paddyroddy:
> I would probably keep it as long as it's one single change.
> 
> To be evaluated again if we start implementing the Array API, since NumPy 2 should be compatible with that out of the box, so we could forego the compatibility layer. Not sure if that's desirable, though, so probably something to discuss soon.

Let's discuss this on Monday. Making sure I understand your POV.

paddyroddy:
Closing in favour of #326. Can be revisited when we are happy to drop `3.9` support (#198).

Issue #176, opened 2024-09-10
Split the release workflow into 2 jobs - build and publish


Saransh-cpp:
The release workflow should ideally create + upload (to GH), and download + release (to PyPI) artifacts in 2 jobs with a few checks on the built SDist.

Issue #187, opened 2024-09-10
Make the repository Scientific Python compliant


Saransh-cpp:
See https://learn.scientific-python.org/development/guides/repo-review/?repo=glass-dev%2Fglass&branch=main for all the missing things. Not all of them _must_ be added, but most of them should be.

Saransh-cpp:
This can be closed now. The only failing checks from repo-review are -

1. [PY007](https://learn.scientific-python.org/development/guides/tasks#PY007): Supports an easy task runner (nox or tox)
  We don't want this as our testing infrastructure is simple enough.
2. [GH212](https://learn.scientific-python.org/development/guides/gha-basic#GH212): Require GHA update grouping
  Getting rid of dependabot in #174 
3. [PC140](https://learn.scientific-python.org/development/guides/style#PC140): Uses a type checker, [MY100](https://learn.scientific-python.org/development/guides/style#MY100): Uses MyPy (pyproject config), and related checks...
  Being tracked in #280 and #246
4. [PC160](https://learn.scientific-python.org/development/guides/style#PC160): Uses a spell checker
  We don't want one permanently in our repository.

paddyroddy:
> This can be closed now. The only failing checks from repo-review are -
> 
>     1. [PY007](https://learn.scientific-python.org/development/guides/tasks#PY007): Supports an easy task runner (nox or tox)
>        We don't want this as our testing infrastructure is simple enough.

I'm not sure if I agree with this, `tox` is very simple to set up (not used `nox` myself). But I agree it is not urgent either way.

Saransh-cpp:
I can create an issue for this for later.

PR #209, opened 2024-09-13
gh-176: revamp publishing workflow

Saransh-cpp:
The developers will now be able to download wheels as artifacts from GH Actions without actually publishing a release. This helps in checking if everything is working well or if you want to distribute a pre-release wheel. Anyways, it is a good practice to keep these 2 jobs separate.

Closes: #176
Refs: #187

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,20 +1,49 @@
 name: Release
 
 on:
+  workflow_dispatch:
   release:
     types:
       - published
 
 jobs:
-  publish:
-    name: Publish on PyPI
+  dist:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build SDist and wheel
+        run: pipx run build
+
+      - name: Check metadata
+        run: pipx run twine check dist/*
+
+      - uses: actions/upload-artifact@v4
+        with:
+          path: dist/*
+
+  publish:
+    needs: [dist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
     environment:
       name: publish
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,20 +1,49 @@
...
     runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
     environment:
       name: publish
       url: https://pypi.org/p/glass
     permissions:
       id-token: write
+
     steps:
-      - uses: actions/checkout@v4
```
Saransh-cpp, line 18: I'm not sure how `checkout` without setting `fetch-depth: 0` was working (or maybe it was silently failing)? Given that glass uses VCS for versioning, it becomes important to pull in the entire git history (not a bad thing) to make the versions/tags work.

.github/workflows/release.yml
```diff
@@ -1,20 +1,49 @@
 name: Release
 
 on:
+  workflow_dispatch:
```
Saransh-cpp, line 4: Can be manually triggered now to build wheels.

.github/workflows/release.yml
ntessore: It should work since it's the release tag itself that is being cloned to a depth of 1 here:

```console
% git clone --branch v2024.1 --depth 1 https://github.com/glass-dev/glass.git glass-v2024.1
% cd glass-v2024.1 
% git describe --tags
v2024.1
```

Saransh-cpp: Ah, I see, the workflow is triggered when a release is published so the it `checkout`s on the tag and not the branch. That makes sense, thanks!

paddyroddy (changes requested):
Why are there testing files in this PR? Seems unrelated

Saransh-cpp:
Ah, I should look at my git history more closely while jumping between branches. Thanks for pointing out!

paddyroddy (approved):
Looks good

PR #212, opened 2024-09-13
gh-211: cancel intermediate builds

### LINKED ISSUE #211: Auto-cancel workflows on repeated PRs/commits
### What is the Topic of Your Question

Other

### Add Your Question Below

Will help save CI resources

Saransh-cpp:
It is a good practice to cancel intermediate builds (old builds when a new build on the same branch/PR starts) to save CI resources. The practice is also more environment friendly 😄 

Closes: #211
Reviewed-by: Nicolas Tessore

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,8 @@ jobs:
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
     - uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
@@ -42,6 +44,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
     - run: pipx run build
 
   docs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -38,17 +42,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
```
Saransh-cpp, line 42: We don't need to fetch the entire git history in any of these jobs.

ntessore: You do because you need the git-rev-parse to work for the hatch-vcs version.

ntessore: You could argue that we don't need it for all jobs, but then we end up with weird glass-0.1something versions in some tests, which I would find confusing. (As a matter of fact, I only recently fixed the fetch-depth because I couldn't understand why the action didn't test the latest version.)

Saransh-cpp: > We do because we need the git-rev-parse to work for the hatch-vcs version.

I meant that we don't need hatch-vcs to do the versioning work for these job, but I see, it can be confusing when reading the workflow logs. I'll revert it!

ntessore (approved):

Issue #46, opened 2023-02-05
Code coverage


ntessore:
Set up codecov or similar to track coverage of unit tests. This should give us a good push to start adding tests, now that we can have them!

ntessore:
This would be nice, and is easy to do, but marking it "future" because it's also not essential for now.

Saransh-cpp:
Hi, @ntessore! Do you have any strong preference for which coverage application to use? I was thinking Codecov because I really like its UI. IIRC, you'll need to add an API token (CODECOV_TOKEN) in the repository secrets on GitHub and enable a few settings (PR comments, annotations, badge) in Codecov to get everything working. I'll create a PR updating the workflow and config files.

ntessore:
No preference for codecov or coveralls from my side, so I am happy to enable codecov right now for this repository.

paddyroddy:
I have a mild preference for `coveralls` (I find `codecov` can be a bit noisy and false-positivey). However, we as a group, recommend both http://github-pages.arc.ucl.ac.uk/python-tooling/pages/ci.html#coverage-monitoring

Saransh-cpp:
I'll go with coveralls, haven't used it before so would be a good learning opportunity

Saransh-cpp:
Sent a coveralls request to organisation owners for approval ✅

Issue #180, opened 2024-09-10
Review markdown badges


paddyroddy:
Standard practice is to have various badges on the README to showcase/show off, such as
- [x] test passing status
- [x] test coverage
- [x] documentation
- [x] python versions supported
- [x] licence
- [x] DOIs (zenodo)
- [x] PyPI version
etc.

PR #213, opened 2024-09-13
gh-46: add infrastructure for coveralls (coverage)

Saransh-cpp:
Add code coverage in workflow using pytest-cov and coveralls.

Closes: #46
Ref: #180

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,10 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: pip install -c .github/test-constraints.txt '.[test]'
-    - run: pytest -v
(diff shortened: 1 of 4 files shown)


ntessore:
Looking at the action, the coverage sees none of the lines run in `glass`?

(Also, the tests in `glass/core/test` are still in the source tree, but that's for another issue.)

ntessore:
This is still installing `glass` in the `pip install` step, but gathering the coverage for the `glass` folder in the source tree.

Saransh-cpp:
Works now! `pytest-cov` requires editable build/install. I will merge this after #217 is merged.

ntessore (approved):
Excellent!

Issue #200, opened 2024-09-12
Find a way to enforce `gh-xxx description` for PR titles


paddyroddy:
There's a desire to have PR titles similar to https://github.com/heracles-ec/heracles, there must be a way to enforce that

PR #218, opened 2024-09-16
gh-200: update CONTRIBUTING.md to follow new title convention

Saransh-cpp:
The CONTRIBUTING guideline should reflect the recent change in PR title convention.

Ref: #200


paddyroddy (approved):
I assume this is ready to be reviewed. Nice catch!

Issue #107, opened 2023-06-14
Functions in `glass` module


ntessore:
Make all public GLASS functionality under the `glass` module name.

This needs a new structure for the documentation, and is therefore a bigger task than just importing the modules into `glass/__init__.py`.

ucapbba:
Hi @ntessore , as you mention, it is straightforward to import into __init__.py - I don't understand the issue on documentation - "Import glass" as well as the existing "Import glass.module" will work in the test cases and examples so documentation could remain the same, because the full path can still be used in the imports.

ntessore:
> the full path can still be used in the imports

That's right, but there should only be one canonical path for each object. Once we import everything into the top-level module, then everything should be documented as accessing the symbols from `glass`. The submodules would effectively become an implementation detail to organise the code.

Once everything is in the top-level module, I think we should always use functions with the module prefix, e.g. `glass.partition()` instead of `from glass import partition; partition()`.

But, as you point out, the full path can still be used, so we would not break anyone's code :)

Saransh-cpp:
It will also be useful to add the `__all__` variable in every module to differentiate public and private API.

Another thing to discuss - the repository has relative imports everywhere at the moment, but PEP8 recommends using absolute imports wherever possible -

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)
> ...
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

All of these issues are interconnected, but please let me know if they should be separate issues on the tracker.

ntessore:
> It will also be useful to add the `__all__` variable in every module to differentiate public and private API.

Indeed, although if only the main `glass` namespace contains public API going forward, we probably don't need to have `__all__` in every module.

> Another thing to discuss - the repository has relative imports everywhere at the moment, but PEP8 recommends using absolute imports wherever possible -

This is mostly a holdover from when GLASS was more modular and namespace-based. We can get rid of the relative imports.

> All of these issues are interconnected, but please let me know if they should be separate issues on the tracker.

This issue seems as good a place as any to discuss.

> Also, `glass.core` is not documented in the [API reference](https://glass.readthedocs.io/stable/reference/index.html). Is it only supposed to be used internally and not by the users directly?

[...]

PR #221, opened 2024-09-16
gh-107: add all public functions/classes under glass namespace

Saransh-cpp:
- All public or user-facing GLASS functionality now falls under the `glass` module name.
- Use absolute imports instead of relative imports within the package (PEP8)

Closes: #107
Changed: All the public functions and classes are now under the `glass` namespace


examples/1-basic/cls.npy
ntessore, line 1: This file probably shouldn't change?

ntessore (changes requested):
A quick `grep -nRI 'glass.' glass` shows the following additional locations:

```
glass/lensing.py:9:The :mod:`glass.lensing` module provides functionality for simulating
glass/lensing.py:43:    from glass.shells import RadialWindow
glass/lensing.py:386:    shells : list of :class:`~glass.shells.RadialWindow`
glass/user.py:9:The :mod:`glass.user` module contains convenience functions for users of the
glass/shapes.py:9:The :mod:`glass.shapes` module provides functionality for simulating the
glass/points.py:9:The :mod:`glass.points` module provides functionality for simulating point
glass/points.py:52:    w : :class:`~glass.shells.RadialWindow`
glass/points.py:126:        the :mod:`~glass.points` module, e.g. ``'linear'`` for
glass/fields.py:9:The :mod:`glass.fields` module provides functionality for simulating random
glass/observations.py:9:The :mod:`glass.observations` module provides functionality for simulating
glass/shells.py:9:The :mod:`glass.shells` module provides functions for the definition of
glass/shells.py:95:        >>> from glass.shells import RadialWindow
glass/galaxies.py:9:The :mod:`glass.galaxies` module provides functionality for simulating galaxies
glass/galaxies.py:46:    w : :class:`~glass.shells.RadialWindow`
glass/galaxies.py:212:    glass.observations.tomo_nz_gausserr :
```

glass/__init__.py
```diff
@@ -1,4 +1,68 @@
 try:
-    from ._version import __version__, __version_tuple__  # noqa: F401
+    from glass._version import __version__, __version_tuple__
```
paddyroddy, line 2: Isn't version one place where you don't do an absolute import? At least in my experience, but can't find a resource to back this up

Saransh-cpp: I use absolute for my projects and it works fine, but I can update this if relative is recommended for versions imports.

Saransh-cpp:
> A quick grep -nRI 'glass.' glass shows the following additional locations:

Ah, I thought the modules should still import other modules explicitly so that developers know where the functions are coming from? They can look at the `__init__` file too, but it should be fine to have long imports internally. I'll be happy to switch it to the new style tho.

examples/1-basic/cls.npy
Saransh-cpp: Ooh, no idea how that happened. Thanks for pointing out!

ntessore:
> > A quick grep -nRI 'glass.' glass shows the following additional locations:
> 
> Ah, I thought the modules should still import other modules explicitly so that developers know where the functions are coming from? They can look at the `__init__` file too, but it should be fine to have long imports internally. I'll be happy to switch it to the new style tho.

Modules should import from the true location. Those examples are all user-facing text in the docs.

glass/__init__.py
Saransh-cpp: Looked around a bit - [`setuptools_scm` suggests using `importlib.metadata`](https://setuptools-scm.readthedocs.io/en/latest/usage/#python-package-metadata) to obtain version in `__init__` file, but scientific-python/cookie suggests doing [`from ._version import version as __version__`](https://github.com/scientific-python/cookie/blob/99acb847958802c719e05cca26ffcb3ddc9e47e1/%7B%7Bcookiecutter.project_name%7D%7D/src/%7B%7Bcookiecutter.__project_slug%7D%7D/__init__.py#L11) (probably because this works with majority of the build-backends).

Saransh-cpp:
> Those examples are all user-facing text in the docs

Oh yes, I missed a few spots. Thank you!

glass/__init__.py
paddyroddy: I'd probably lean toward the `scientif-python` one, I must have found that convention somewhere!

paddyroddy (approved):
Much simpler, just change that version `__init__.py` thing and good to go I reckon

ntessore (approved):
Great work, this had been stuck in the queue for ages!

paddyroddy:
I've merged this as will need for #232

PR #223, opened 2024-09-17
Don't make second box mandatory for feature request

paddyroddy:
Especially for small issues this can be a bit unnecessary so may as well remove it


Saransh-cpp (approved):
Can we maybe also enable blank issues? I _sometimes_ find it hard to fit things inside these templates / they are _sometimes_ intimidating for people new to GitHub.

paddyroddy:
> Can we maybe also enable blank issues? I _sometimes_ find it hard to fit things inside these templates / they are _sometimes_ intimidating for people new to GitHub.

Sure, please raise an issue

Issue #170, opened 2024-09-06
Update pre-commit setup


ntessore:
Use a modern pre-commit setup with ruff etc.

Saransh-cpp:
All Scientific Python recommended (and a few extras as well) pre-commit hooks have been added. The only major one missing is `mypy`, but that is being tracked in #280. Closing this as completed!

PR #230, opened 2024-09-17
gh-170: add `yamllint` and run over the repo

paddyroddy:
Still working towards #170. The options seem like a lot but these are ones I've used and I've found great for readability. Plus, for example, removing Python versions makes diffs easier when doing something like
```yaml
python-version:
  - "3.8"
  - "3.9"
  - "3.10"
  - "3.11"
```
rather than
```yaml
python-version: ["3.8", "3.9", "3.10", "3.11"]
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+---
 ci:
   autoupdate_commit_msg: "MNT: update pre-commit hooks"
   autofix_commit_msg: "STY: pre-commit fixes"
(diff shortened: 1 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -1,4 +1,51 @@
+---
 repos:
+  - repo: https://github.com/adrienverge/yamllint
+    rev: v1.35.1
+    hooks:
+      - id: yamllint
+        args:
```
Saransh-cpp, line 7: Is it possible to add these args as config in pyproject.toml? Looks good otherwise!

ntessore (approved):

.pre-commit-config.yaml
ntessore: Can always add "move some of the pre-commit config to pyproject" as a to-do item.

Saransh-cpp:
Taking over 🫡

.pre-commit-config.yaml
paddyroddy: https://github.com/glass-dev/glass/issues/267

paddyroddy: Sorry this is my bad. These options have been tried and tested, but I have some hacky "global" pre-commit hooks, where I don't want to have the same config everywhere https://github.com/paddyroddy/.github/blob/main/precommit/general/general-hooks.yaml

Saransh-cpp: Ah, thanks for looking into this, keeping it as it is should be fine!

PR #236, opened 2024-09-18
gh-170: add traditional pygrep hooks

Saransh-cpp:
Adds several misc `pygrep` hooks recommended by scientific python.

Refs: #170, #187


paddyroddy (approved):
Nice, not used these before

Issue #204, opened 2024-09-12
Only test examples if something in source folder changed


ntessore:
I noticed we run a lot of expensive tests of the examples for infrastructure changes. IIRC we can restrict runs of the `examples` job to cases where something in `glass` has changed.

paddyroddy:
There's a bunch of other stuff we could do like cancelling ones in progress per branch, only run on non-draft PRs, etc.
![image](https://github.com/user-attachments/assets/f92b1478-9e51-4378-bccb-73f169556cb9)

paddyroddy:
One disadvantage of only selectively testing is that you can used enforced status checks https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks. We're not currently using these (as far as I can tell, don't have full access to the settings), they can be handy if a PR is ready to go, and you can [enable automerge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request) and just let CI run in the background after approval

ntessore:
No strong feelings from my side either way, but I always found enforced status checks a bit useless (either you have the time to wait for status checks, so you do, or you don't, so you don't). Feel free to close this issue as wontfix.

PR #237, opened 2024-09-18
gh-204: split examples job and run it only on code changes

Saransh-cpp:
Makes example tests a standalone job that executes only if something is changed in `glass/`.

Closes: #204


paddyroddy (approved):
Noice

Issue #208, opened 2024-09-12
Remove legacy random number generation


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

See explanation here https://docs.astral.sh/ruff/rules/numpy-legacy-random, have verified on the codebase with `ruff check --select NPY002 .`, there are 12 errors at the moment. These had support dropped from `numpy==1.16`

### Describe the Solution You'd Like

Use syntax like this instead
```python
rng = np.random.default_rng(1337)
rng.normal()
```

ntessore:
We generally have `rng` support for all user functions, so these should be "safe" to update in examples and tests only.

Random number generation more widely will be one of the issues to solve for generic array support, so might be good to keep that in mind as well.

PR #241, opened 2024-09-18
gh-208: remove legacy random number generation

Saransh-cpp:
Gets rid of legacy `RandomState` and replaced it with the new random number `Generator` for rng calls.

Fixes: #208

### DIFF
--- a/tests/core/test_algorithm.py
+++ b/tests/core/test_algorithm.py
@@ -17,8 +17,9 @@ def test_nnls():
 
     from glass.core.algorithm import nnls as nnls_glass
 
-    a = np.random.randn(100, 20)
-    b = np.random.randn(100)
+    rng = np.random.default_rng(seed=42)
+    a = rng.standard_normal((100, 20))
+    b = rng.standard_normal((100,))
 
     x_glass = nnls_glass(a, b)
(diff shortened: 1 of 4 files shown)


paddyroddy (changes requested):
I think it would be good to have the random seed as a test constant or `pytest.fixture`

ntessore:
> I think it would be good to have the random seed as a test constant or `pytest.fixture`

Agree, I usually make `rng` itself the fixture, which can be helpful for testing with a mock RNG.

tests/core/test_algorithm.py
```diff
@@ -10,15 +10,22 @@
     HAVE_SCIPY = True
 
 
+@pytest.fixture
+def rng():
+    import numpy as np
+
+    return np.random.default_rng(seed=42)
+
+
```
ntessore, line 19: Move this to conftest.py so it only needs to be defined once?

ntessore (changes requested):

tests/core/test_algorithm.py
Saransh-cpp: Sounds better, thanks!

ntessore (approved):

PR #242, opened 2024-09-18
gh-204: examples run twice on PRs

Saransh-cpp:
Minor fix. The examples run twice on PRs right now as `branches` is not specified under `on`.

Refs: #204

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -4,9 +4,13 @@ on:
   push:
     paths:
       - "glass/**"
+    branches:
+      - main
   pull_request:
     paths:
       - "glass/**"
+    branches:
+      - main
 
 concurrency:
   # Skip intermediate builds: always.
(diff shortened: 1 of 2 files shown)


.github/workflows/examples.yml
```diff
@@ -4,9 +4,13 @@ on:
   push:
     paths:
       - "glass/**"
+    branches:
+      - main
   pull_request:
     paths:
       - "glass/**"
+    branches:
+      - main
```
paddyroddy, line 13: In the event we are merging to non-`main` branches, are we sure we want this line?
```suggestion
```

paddyroddy (changes requested):
Good catch

.github/workflows/examples.yml
Saransh-cpp: The main test workflow skips PRs on non-main branches too -

https://github.com/glass-dev/glass/blob/d7bd1e2f76bea37c67050c6acdae06fb18cfc71f/.github/workflows/test.yml#L7-L9

I guess it will make sense to remove it from there as well.

paddyroddy: Agreed

paddyroddy (approved):
Nice, merging

Issue #222, opened 2024-09-17
Following #178, add to the deployment workflow something involving `TestPyPI`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

We now have `TestPyPI` so we can add to the deployment workflow to make sure it is working

### Describe the Solution You'd Like

-

Saransh-cpp:
Should wait for #176 to go in

PR #244, opened 2024-09-18
gh-222: add infrastructure for testpypi uploads

Saransh-cpp:
The workflow now makes a PyPI upload if a release is created on GitHub, or if the workflow is triggered manually with an explicit "pypi" input. For default manual triggers, it will now upload to TestPyPI.

This might require configuring release environment in TestPyPI project.

Closes: #222


paddyroddy (approved):
Nice idea, I like it

Issue #226, opened 2024-09-17
Add `prettier` for non-Python files


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Add https://prettier.io which is a great code formatter for a variety of config options. Allows for consistency everywhere, streamlining PRs

PR #247, opened 2024-09-18
gh-226: pre-commit's mirror of prettier is obsolete

Saransh-cpp:
[pre-commit's mirror of prettier](https://github.com/pre-commit/mirrors-prettier) was [archived](https://github.com/pre-commit/mirrors-prettier?tab=readme-ov-file#archived) a few months back -

> # archived
>
> prettier made some changes that breaks plugins entirely

Scientific Python recommends using https://github.com/rbubley/mirrors-prettier?tab=readme-ov-file now.

Refs: #226

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
           - --in-place
           - --spaces-indent-inline-array=4
           - --trailing-comma-inline-array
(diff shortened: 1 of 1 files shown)


paddyroddy (approved):
Okay cool, I was aware it was archived, :/ good to know Scientific Python have a recommendation. Wasn't aware of that - there have been many made since.

Issue #239, opened 2024-09-18
`test_fits.py::test_out_filename` is accessing private functions from the `fitsio` library


paddyroddy:
### Describe the Bug

This is bad practice, as the external code cannot guarantee keeping internal functions with a consistent API. Are we happy with this? https://github.com/glass-dev/glass/blob/f7cedc9e5289f0505a15aff742ca0ff8bb4c180e/tests/test_fits.py#L93-L99

### To Reproduce

```python
import glass
...
```

### Version In Use

2024.2.dev41+ga3f00f2.d20240918

ntessore:
Yeah, we can use `get_filename()` to obtain the same thing.

Saransh-cpp:
I couldn't find a `get_filename()` in fitsio. Is the function somewhere else or am I missing something?

Issue #238, opened 2024-09-18
Is `test_fits.py::test_out_filename` meant to utilise `tmp_path`?


paddyroddy:
### Describe the Bug

The test https://github.com/glass-dev/glass/blob/a101ab48d8e9fcc6be5dc3fdc491afe971beda4e/tests/test_fits.py#L93-L99 has `tmp_path` fixture but isn't using it. Is this intentional or an oversight?

### To Reproduce

```python
import glass
...
```

### Version In Use

2024.2.dev38+gd9ced9d.d20240918

ntessore:
Definitely an oversight, and probably how we ended up with `*.Fits` (with wrong case) in .gitignore. Can we fix the test, the `filename`, and `.gitignore`? :)

paddyroddy:
Of course!

PR #253, opened 2024-09-19
gh-238: utilise tmp_path correctly in tests

Saransh-cpp:
Closes: #238
Closes: #239

### DIFF
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -34,14 +34,14 @@ def test_basic_write(tmp_path):
 
     with user.write_catalog(
         tmp_path / filename_gfits, ext="CATALOG"
-    ) as out, fitsio.FITS(tmp_path / filename_tfits, "rw", clobber=True) as myFits:  # noqa: N806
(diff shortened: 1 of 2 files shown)


tests/test_fits.py
```diff
@@ -93,9 +86,9 @@ def test_write_exception(tmp_path):
...
 @pytest.mark.skipif(not HAVE_FITSIO, reason="test requires fitsio")
-def test_out_filename():
+def test_out_filename(tmp_path):
     import fitsio
 
-    fits = fitsio.FITS(filename, "rw", clobber=True)
+    fits = fitsio.FITS(tmp_path / filename, "rw", clobber=True)
     writer = user._FitsWriter(fits)  # noqa: SLF001
-    assert writer.fits._filename == filename  # noqa: SLF001
+    assert writer.fits._filename == f"{tmp_path}/{filename}"  # noqa: SLF001
```
ntessore, line 94: What is this actually testing? That `fitsio` works? In which case, none of our business here.

Saransh-cpp: Actually yes, the test is redundant.

ntessore (approved):

Issue #185, opened 2024-09-10
Add developer documentation


paddyroddy:
This makes it obvious exactly how others can contribute and can include things like the contribution guidelines, commit styles etc.
https://brainglobe.info/community/developers/index.html

paddyroddy:
This should include things like the commit lint applies to this repo

PR #255, opened 2024-09-19
gh-185: update CONTRIBUTING.md (developer guide)

Saransh-cpp:
I have updated the contributing guideline to reflect all the recent changes. I think keeping it just on GitHub should be okay, given that the new developers will look at the GitHub repository before contributing.

Closes: #185

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,143 @@
-# Contributing guidelines
+# Contributing to GLASS
 
-## Workflow
+If you are planning to develop _GLASS_, or want to use the latest commit of
+_GLASS_ on your local machine, you might want to install it from the source.
+This installation is not recommended for users who want to use the stable
+version of _GLASS_. The page below describes how to build, test, and develop
+_GLASS_.
+
+## Installation
+
+The developer installation of _GLASS_ comes with several optional dependencies -
+
+- `test`: installs `scipy`, `fitsio`, and the relevant `pytest` dependencies
+- `docs`: installs `sphinx`, `nbsphinx`, and a few other docs related
+  dependencies
+- `examples`: installs `camb`, `glass.ext.camb`, and a few notebook related
+  dependencies
+
+These options can be used with `pip` with the editable (`-e`) mode of
+installation in the following way -
+
+```bash
+pip install -e .[docs,test]
+```
+
+## Tooling
+
+### Pre-commit
+
+_GLASS_ uses a set of `pre-commit` hooks and the `pre-commit.ci` bot to format,
+lint, and prettify the codebase. The hooks can be installed locally using -
+
+```bash
+pre-commit install
+```
+
+This would run the checks every time a commit is created locally. The checks
+will only run on the files modified by that commit, but the checks can be
+triggered for all the files using -
+
+```bash
+pre-commit run --all-files
+```
+
+If you would like to skip the failing checks and push the code for further
+discussion, use the `--no-verify` option with `git commit`.
+
+## Testing
+
+_GLASS_ is tested using `pytest` and `pytest-doctestplus`. `pytest` is
+responsible for testing the code, whose configuration is available in
+[pyproject.toml](https://github.com/glass-dev/glass/blob/main/pyproject.toml).
+`pytest-doctestplus` is responsible for testing the examples available in every
+docstring, which prevents them from going stale. Additionally, _GLASS_ also uses
+`pytest-cov` (and [Coveralls](https://coveralls.io)) to calculate/display the
+coverage of these unit tests.
+
+### Running tests locally
+
+The tests can be executed using the `test` dependencies of _GLASS_ in the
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,6 +1,143 @@
...
+
+## Installation
+
+The developer installation of _GLASS_ comes with several optional dependencies -
+
+- `test`: installs `scipy`, `fitsio`, and the relevant `pytest` dependencies
+- `docs`: installs `sphinx`, `nbsphinx`, and a few other docs related
+  dependencies
+- `examples`: installs `camb`, `glass.ext.camb`, and a few notebook related
+  dependencies
```
paddyroddy, line 17: Do we want to hardcode this here? If we change dependencies, it is another thing to change. Could just say `test` is related to testing dependencies

CONTRIBUTING.md
```diff
@@ -1,6 +1,143 @@
...
+
+These options can be used with `pip` with the editable (`-e`) mode of
+installation in the following way -
+
+```bash
+pip install -e .[docs,test]
```
paddyroddy, line 23: If running in `zsh` one needs quotes, might want to include so it'll copy-and-paste fine.
```suggestion
pip install -e ".[docs,test]"
```

paddyroddy: I also wonder if should add an `"all"` option

CONTRIBUTING.md
```diff
@@ -1,6 +1,143 @@
...
+`pytest-cov` (and [Coveralls](https://coveralls.io)) to calculate/display the
+coverage of these unit tests.
+
+### Running tests locally
+
+The tests can be executed using the `test` dependencies of _GLASS_ in the
+following way -
+
+```bash
+python -m pytest --cov=glass --doctest-plus -v
```
paddyroddy, line 64: For another PR, but we can configure the coverage in `pyproject.toml` so one doesn't need to explicitly write `--cov=glass`. https://github.com/astro-informatics/sleplet/blob/782279ec1be404aa356c6e0b583bd536cf3a4789/pyproject.toml#L101-L109

CONTRIBUTING.md
```diff
@@ -1,6 +1,143 @@
...
+[reStructurredText](https://docutils.sourceforge.io/docs/user/rst/quickref.html).
+The docstrings include the description, arguments, examples, return values, and
+attributes of a class or a function, and the `.rst` files enable us to render
+this documentation on _GLASS_'s documentation website.
+
+_GLASS_ primarily uses [Sphinx](https://www.sphinx-doc.org/en/master/) for
+rendering documentation on its website. The configuration file (`conf.py`) for
+`sphinx` can be found
+[under the `docs` folder](https://github.com/glass-dev/glass/blob/main/docs/conf.py).
+The documentation is deployed on [https://readthedocs.io]()
```
paddyroddy, line 80:
```suggestion
The documentation is deployed on <https://readthedocs.io>
```

CONTRIBUTING.md
```diff
@@ -1,6 +1,143 @@
...
+If any _GLASS_ extension packages depend on the new release, new versions of
+these packages should be produced as soon as the new release is published to
+PyPI.
+
+### Versioning
+
+_GLASS_ follows [CalVer](https://calver.org). There is no difference between
+releases that increment the year and releases that increment the month; in
+particular, releases that increment the month may introduce breaking changes.
+The version is generate dynamically through VCS using `hatch-vcs`.
```
paddyroddy, line 132: Can you add a link to `hatch-vcs`?

paddyroddy (changes requested):
Thanks for doing this! Various suggestions

CONTRIBUTING.md
Saransh-cpp: I usually avoid using the `[all]` option because it brings in everything, from docs to dev to test, which makes the environment cumbersome to manage or reproduce.

CONTRIBUTING.md
Saransh-cpp: Added `-v` in the recent PR. I usually avoid adding the `cov` and the `doctest-plus` option in pyproject.toml because we don't need those always.

CONTRIBUTING.md
paddyroddy: Fair enough

CONTRIBUTING.md
paddyroddy: I agree, my suggestion is you add to it the `coverage` section. So when you ask for coverage you get it on `glass` without having to specify it - I'm not asking for the `pytest` section

Saransh-cpp: Ahh, let me quickly check that.

paddyroddy: Don't worry we can fix in another issue https://github.com/glass-dev/glass/issues/282

paddyroddy (approved):
Looks nice

Issue #257, opened 2024-09-23
Use pytest-mock


Saransh-cpp:
### Is Your Feature Request Related to a Problem? Please Describe

The tests use `mock` from unittest right now, but we should ideally be using the `pytest-mock` plugin.

PR #258, opened 2024-09-23
gh-257: use pytest-mock instead of unittest.mock

Saransh-cpp:
Replaces `unittest.mock.Mock` with pytest's mocking plugin - `pytest-mock` ( a new `tests` dependency now).

Closes: #257


paddyroddy (changes requested):
Can you add some text to the PR body?

ntessore:
Sorry, is there any benefit at all to this change? If not, why add a dependency that might stop working at any point?

Saransh-cpp:
The benefit imo - using one testing framework uniformly across the project. [pytest-mock](https://github.com/pytest-dev/pytest-mock/) is maintained by pytest itself, so there are no chances of it breaking (techincally `pytest` is a third party library/dependency too, so you could argue that it might break sometime and we should use `unittest` for everything, but the `pytest` ecosystem is well tested and is used extensively within the Python ecosystem).

ntessore:
> The benefit imo - using one testing framework uniformly across the project. [pytest-mock](https://github.com/pytest-dev/pytest-mock/) is maintained by pytest itself, so there are no chances of it breaking

Ah, excellent, I thought it was a third-party project. That makes sense!

paddyroddy (approved):

PR #264, opened 2024-09-25
gh-276: add tests for `glass.shapes`

Saransh-cpp:
Adds tests for `glass.shapes.triaxial_axis_ratio` and `glass.shapes.ellipticity_ryden04` inspired by `skypy`.
Sets up `.coveragrc` (coverage config file) to ignore unrelated files.

Closes: #276

### DIFF
--- a/glass/shapes.py
+++ b/glass/shapes.py
@@ -29,7 +29,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 
 if TYPE_CHECKING:
-    from numpy.typing import ArrayLike, NDArray  # pragma: no cover
+    from numpy.typing import ArrayLike, NDArray
 
 
 def triaxial_axis_ratio(zeta, xi, size=None, *, rng=None):
--- a/glass/shapes.py
+++ b/glass/shapes.py
@@ -29,7 +29,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 
 if TYPE_CHECKING:
-    from numpy.typing import ArrayLike, NDArray
+    from numpy.typing import ArrayLike, NDArray  # pragma: no cover
 
 
 def triaxial_axis_ratio(zeta, xi, size=None, *, rng=None):
--- a/tests/test_shapes.py
+++ b/tests/test_shapes.py
@@ -2,6 +2,56 @@ import numpy as np
 import pytest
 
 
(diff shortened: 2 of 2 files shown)


.coveragerc
```diff
@@ -0,0 +1,2 @@
+[run]
+omit = glass/_version.py
```
paddyroddy, line 2: Can we make coverage stuff a separate PR?

glass/shapes.py
```diff
@@ -31,7 +31,7 @@
 import numpy as np
 
 if TYPE_CHECKING:
-    from numpy.typing import ArrayLike, NDArray
+    from numpy.typing import ArrayLike, NDArray  # pragma: no cover
```
paddyroddy, line 34: I've said this in a separate PR, but I think we should do this https://coverage.readthedocs.io/en/latest/excluding.html#advanced-exclusion

paddyroddy (changes requested):

.coveragerc
Saransh-cpp: Yes, was about to do that.

glass/shapes.py
Saransh-cpp, line 34:
```suggestion
    from numpy.typing import ArrayLike, NDArray
```

Saransh-cpp:
Need to add some comments and docstring.

Saransh-cpp (commented):
Hi @paddyroddy, could you please let me know (and possibly link an example for) what kind of docstrings should be added to the tests?

paddyroddy:
> Hi @paddyroddy, could you please let me know (and possibly link an example for) what kind of docstrings should be added to the tests?

If you think it's overkill, feel free to skip. To the uninitiated it just seems like a bunch of random input/outputs, so just a little explanation would be helpful IMO

paddyroddy (approved):
Just seen the comments, much better. Any reason why the blank line between comment and test?

Saransh-cpp:
> Any reason why the blank line between comment and test?

Really not a fan of the extra blank lines, but again, just following the way it is done in the rest of the existing tests to maintain uniformity😅

paddyroddy:
> Really not a fan of the extra blank lines, but again, just following the way it is done in the rest of the existing tests to maintain uniformity😅

Okay makes sense!

Issue #267, opened 2024-09-25
Where possible add config options to `pyproject.toml` rather than `.pre-commit-config.yaml`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

The main advantage of this is that it means whenever one runs, say, `ruff` it will apply the config. If it is only in the `pre-commit` then it will only apply when running `ruff` via `pre-commit`.

PR #281, opened 2024-09-25
gh-267: move verbose output flag for ruff and pytest to pyproject.toml

Saransh-cpp:
Verbose output flags for both ruff and pytest are used frequently enough to be added to pyproject.toml.

Refs: #267


Saransh-cpp:
pre-commit.ci run

paddyroddy (approved):
Looks good

paddyroddy:
Failing test appears to be noise

PR #284, opened 2024-09-26
gh-267: move arguments from `pre-commit` to config files

### LINKED ISSUE #267: Where possible add config options to `pyproject.toml` rather than `.pre-commit-config.yaml`

paddyroddy:
Closes #267. The main advantage here is when running the given tool it will always work, rather than just via `pre-commit`. Previously requested in https://github.com/glass-dev/glass/pull/230#discussion_r1764036548. Have used `pyproject.toml` where possible.


Saransh-cpp (approved):
Ah, you moved it out already, should work too, no strong opinions. Thanks!

PR #289, opened 2024-09-27
gh-288: run CI when a draft PR is marked as ready

### LINKED ISSUE #288: CI should run when a draft PR gets marked as ready for review
### Describe the Bug

Some crucial bits are missing in the `on` section of the GitHub Actions workflows. So if a PR is marked as draft, the CI will never run.

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

2024.2.dev5+g27fa141.d20240911

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Fixes #288. Currently a PR is never tested the moment it gets marked as draft. The idea behind not running in draft is to limit CI use, but we need to ensure everything is ran before merging to `main`.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -10,6 +10,12 @@ on:
   pull_request:
     paths:
       - glass/**
+    # ensure the draft PRs are tested when ready
+    types:
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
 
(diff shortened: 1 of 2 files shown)


Saransh-cpp (approved):
Thanks!

Issue #302, opened 2024-10-02
Apply `actionlint`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Add https://github.com/rhysd/actionlint to lint the GitHub Action workflows

PR #287, opened 2024-09-27
gh-283: set up nox for developers and CI

Saransh-cpp:
- Adds `noxfile` to support `nox` sessions for testing, docs, and linting/formatting.
- `Nox` is now used to run everything in the CI.
- Adds documentation about the new infrastructure in contributing guidelines.

Closes: #283

PR #303, opened 2024-10-02
gh-302: add `actionlint` and fix issues

paddyroddy:
Add https://github.com/rhysd/actionlint to check through GitHub Actions workflows.

The fix was related to this error

```yaml
.github/workflows/test.yml:51:31: receiver of object dereference "key" must be type of object but got "number" [expression]
   |
51 |           nox -s doctests-${{ matrix.python-version.key || matrix.python-version
   |                               ^~~~~~~~~~~~~~~~~~~~~~~~~
.github/workflows/test.yml:55:31: receiver of object dereference "key" must be type of object but got "number" [expression]
   |
55 |           nox -s coverage-${{ matrix.python-version.key || matrix.python-version
```

Why did we have this `matrix.python-version.key` bit @Saransh-cpp? Introduced in https://github.com/glass-dev/glass/pull/287, sorry I missed it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,13 +47,9 @@ jobs:
       - name: Install nox and coverage.py
         run: pip install coverage[toml] nox
       - name: Run doctests
-        run:
-          nox -s doctests-${{ matrix.python-version.key || matrix.python-version
-          }} --verbose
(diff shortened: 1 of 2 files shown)


Saransh-cpp (approved):
Thanks, @paddyroddy! Ignore the comment above 😬

Issue #280, opened 2024-09-25
Add `mypy` for type checking


Saransh-cpp:
### Is Your Feature Request Related to a Problem? Please Describe

It will be useful for both the users and the developers to have well defined static types for everything in glass. The repository right now does have types in places but they are not tested. These can be tested by adding mypy as a pre-commit hook. Satisfying mypy will also pave the way for experimenting with [mypyc](https://mypyc.readthedocs.io/en/latest/introduction.html) compilations.

Related issues - https://github.com/glass-dev/glass/issues/187. https://github.com/glass-dev/glass/issues/246

paddyroddy:
See #358

paddyroddy:
Closing due to #359, in favour of #358

PR #305, opened 2024-10-03
gh-246: replace `numpy.ndarray` with `numpy.typing.NDArray`

### LINKED ISSUE #246: Use `numpy.typing` instead of `numpy.ndarray`

This allows greater control of the type of a specific NumPy array https://numpy.org/doc/stable/reference/typing.html

_No response_

_No response_

### Additional Context

_No response_

paddyroddy:
Closes #246. Use `numpy.typing.NDArray` as the first step towards #280. With `NDArray` we can use types, i.e. `NDArray[np.float64]`, which I'll save for future work (#306).

I have also gone throughout and made sure anything like `ArrayLike` is using the alias `import numpy.typing as npt`. Imports are generally in a `if TYPE_CHECKING` block where possible.

We are using `numpy.testing` in some places. This was being aliased as `npt`. To avoid confusion, I have removed this alias and left them as an explicit import.

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -37,12 +37,12 @@ import numpy.typing as npt
 from gaussiancl import gaussiancl
 
 # types
-Array = np.ndarray
+Array = npt.NDArray
 Size = Union[None, int, Tuple[int, ...]]
 Iternorm = Tuple[Optional[int], Array, Array]
 ClTransform = Union[str, Callable[[Array], Array]]
 Cls = Sequence[Union[Array, Sequence[float]]]
-Alms = np.ndarray
+Alms = npt.NDArray
 
 
 def iternorm(
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -37,7 +37,7 @@ import numpy.typing as npt
 from gaussiancl import gaussiancl
 
 # types
-Size = None | int | tuple[int, ...]
+Size = int | tuple[int, ...] | None
 Iternorm = tuple[int | None, npt.NDArray, npt.NDArray]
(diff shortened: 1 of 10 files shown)


glass/fields.py
```diff
@@ -29,25 +29,24 @@
...
-Size = Union[None, int, Tuple[int, ...]]
-Iternorm = Tuple[Optional[int], Array, Array]
-ClTransform = Union[str, Callable[[Array], Array]]
-Cls = Sequence[Union[Array, Sequence[float]]]
-Alms = np.ndarray
+Size = None | int | tuple[int, ...]
+Iternorm = tuple[int | None, npt.NDArray, npt.NDArray]
+ClTransform = str | Callable[[npt.NDArray], npt.NDArray]
+Cls = Sequence[npt.NDArray | Sequence[float]]
+Alms = npt.NDArray
```
Saransh-cpp, line 44: I have tried this before and the newer syntax does not work when you are making your own custom types.

paddyroddy: Well let's see. I got an error, but that looked to be from it being `None | <x>` rather `<x> | None`.

paddyroddy: Yeah, it was that. Have restored `Optional`, `Tuple`, `Union` here

Saransh-cpp (approved):
This looks good, thanks, @paddyroddy!

paddyroddy:
This is ready, but blocked by #307.

paddyroddy:
Actually thinking about it, the test failure is related to the notebooks - which aren't being touched. So actually this is okay to merge.

Issue #312, opened 2024-10-03
Fix `broadcast_first` test


Saransh-cpp:
### Describe the Bug

The test is failing with the logs -
```py
_____________________________ test_broadcast_first _____________________________

    def test_broadcast_first():
        a = np.ones((2, 3, 4))
        b = np.ones((2, 1))
    
        # arrays with shape ((3, 4, 2)) and ((1, 2)) are passed
        # to np.broadcast_arrays; hence it works
        a_a, b_a = broadcast_first(a, b)
        assert a_a.shape == (2, 3, 4)
        assert b_a.shape == (2, 3, 4)
    
        # plain np.broadcast_arrays will not work
        with pytest.raises(ValueError, match="shape mismatch"):
            np.broadcast_arrays(a, b)
    
        # arrays with shape ((5, 6, 4)) and ((6, 5)) are passed
        # to np.broadcast_arrays; hence it will not work
        a = np.ones((4, 5, 6))
        b = np.ones((5, 6))
    
        with pytest.raises(ValueError, match="shape mismatch"):
            broadcast_first(a, b)
    
        # plain np.broadcast_arrays will work
>       a_a, b_a = broadcast_first(a, b)

tests/core/test_array.py:41: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
glass/core/array.py:11: in broadcast_first
    arrays = np.broadcast_arrays(*arrays)
.nox/coverage-3-10/lib/python3.10/site-packages/numpy/lib/_stride_tricks_impl.py:558: in broadcast_arrays
    shape = _broadcast_shape(*args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (array([[[1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., ...,
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]]))

    def _broadcast_shape(*args):
        """Returns the shape of the arrays that would result from broadcasting the
        supplied arrays against each other.
        """
[...]

PR #314, opened 2024-10-03
gh-312: fix tests for broadcast_first

Saransh-cpp:
The comment was right but the code was wrong 🤦🏻‍♂️

Use `np.broadcast_arrays` for comparison.

Fixes: #312


paddyroddy (approved):
Ah lol, I see this issue

Issue #304, opened 2024-10-02
Do we want to exclude anything from the final `SDist`?


paddyroddy:
### What is the Topic of Your Question

Other

### Add Your Question Below

I have more experience in [setuptools-scm](https://pypi.org/project/setuptools-scm) than [hatch-vcs](https://pypi.org/project/hatch-vcs) we are currently using. In [setuptools](https://pypi.org/project/setuptools) land you can use a [`Manifest.in`](https://learn.scientific-python.org/development/guides/packaging-classic/#includingexcluding-files-in-the-sdist) to exclude extra stuff that would otherwise go in, see e.g. https://github.com/astro-informatics/sleplet/blob/main/MANIFEST.in. Can/should we do this here? Or am I overthinking/misunderstanding [hatchling](https://pypi.org/project/hatchling).

Saransh-cpp:
Ahh, I always thought that hatch only keeps the required and "important" things in the SDist, but that is not the case -

```
When the user has not set any [file selection](https://hatch.pypa.io/latest/config/build/#file-selection) options, all files that are not [ignored by your VCS](https://hatch.pypa.io/latest/config/build/#vcs) will be included.

The following files are always included and cannot be excluded:

- `/pyproject.toml`
- `/hatch.toml`
- `/hatch_build.py`
- `/.gitignore or /.hgignore`
- Any defined [readme](https://hatch.pypa.io/latest/config/metadata/#readme) file
- All defined [license-files](https://hatch.pypa.io/latest/config/metadata/#license)
```

Right, we should ignore some folders and file. I will start by ignoring the ones mentioned in the linked `MANIFEST.in`.

paddyroddy:
> Ahh, I always thought that hatch only keeps the required and "important" things in the SDist, but that is not the case -

Yeah, `setuptools-scm` works exactly the same. Like it includes literally everything tracked by `git`. To be fair, how would it know what is important?

PR #316, opened 2024-10-03
gh-304: exclude a few things from SDist

Saransh-cpp:
Excludes `docs/`, `tests/`, `examples/`, `noxfile.py`, and all the `.*` files (except `.gitignore`) from the SDist.

Closes: #304


paddyroddy (approved):
Thanks! Good to know how to do it in `hatch`

Issue #315, opened 2024-10-03
`size` is not inferred automatically in `ellipticity_ryden04` (wrong docstring?)


Saransh-cpp:
### Add your issue here

Refs: #299

The [docstring of `ellipticity_ryden04` states](https://github.com/glass-dev/glass/blob/643504c9166addb9dfa70aa71dc9e8b4910561cf/glass/shapes.py#L121-L122) -
```py
def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None):  # noqa: PLR0913
    """
    ...
    size : int or tuple of ints or None
        Sample size.  If ``None``, the size is inferred from the parameters.
    ...
    """
```

but `size` is not inferred automatically in a few places within the function -

https://github.com/glass-dev/glass/blob/643504c9166addb9dfa70aa71dc9e8b4910561cf/glass/shapes.py#L143-L152

Is this an intended behavior or is this a bug?

For instance, the [implementation in `skypy`](https://github.com/skypyproject/skypy/blob/ee05b5aba9118663c68fc69ffae523bc0e450e8a/skypy/galaxies/morphology.py#L265-L267) does the following -

```py
    # get size if not given
    if size is None:
        size = np.broadcast(mu_gamma, sigma_gamma, mu, sigma).shape
```

Issue #299, opened 2024-10-01
`tests/test_shapes.py::test_ellipticity_ryden04` seems flaky


paddyroddy:
### Describe the Bug

I've noticed that the `python==3.12` test fails occasionally, specifically `tests/test_shapes.py::test_ellipticity_ryden04`. Can you have a look @Saransh-cpp?

<details><summary>Logs</summary>

```
>       e2 = ellipticity_ryden04(-1.85, 0.89, 0.222, [0.056, 0.067])

tests/test_shapes.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mu = -1.85, sigma = 0.89, gamma = 0.222, sigma_gamma = [0.056, 0.067]
size = None

    def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None):  # noqa: PLR0913
        r"""
        Ellipticity distribution following Ryden (2004).
    
        The ellipticities are sampled by randomly projecting a 3D ellipsoid with
        principal axes :math:`A > B > C` [1]_.  The distribution of :math:`\log(1 -
        B/A)` is normal with mean :math:`\mu` and standard deviation :math:`\sigma`.
        The distribution of :math:`1 - C/B` is normal with mean :math:`\gamma` and
        standard deviation :math:`\sigma_\gamma` [2]_.  Both distributions are
        truncated to produce ratios in the range 0 to 1 using rejection sampling.
    
        Parameters
        ----------
        mu : array_like
            Mean of the truncated normal for :math:`\log(1 - B/A)`.
        sigma : array_like
            Standard deviation for :math:`\log(1 - B/A)`.
        gamma : array_like
            Mean of the truncated normal for :math:`1 - C/B`.
        sigma_gamma : array_like
            Standard deviation for :math:`1 - C/B`.
        size : int or tuple of ints or None
            Sample size.  If ``None``, the size is inferred from the parameters.
        rng : :class:`~numpy.random.Generator`, optional
            Random number generator.  If not given, a default RNG will be used.
    
        Returns
        -------
        eps : array_like
            Array of :term:`ellipticity` from projected axis ratios.
    
        References
        ----------
        .. [1] Ryden B. S., 2004, ApJ, 601, 214.
        .. [2] Padilla N. D., Strauss M. A., 2008, MNRAS, 388, 1321.
[...]

PR #317, opened 2024-10-03
gh-315: add broadcasting rule in ellipticity_ryden04 + tests

Saransh-cpp:
Added a broadcasting rule inspired by `skypy` to prevent crashes. 

Fixes: #315
Closes: #299 
Fixed: `ellipticity_ryden04` infers the size argument on the basis of other arguments if not explicitly passed

### DIFF
--- a/glass/shapes.py
+++ b/glass/shapes.py
@@ -138,6 +138,10 @@ def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None): 
     if rng is None:
         rng = np.random.default_rng()
 
+    # default size if not given
+    if size is None:
+        size = np.broadcast(gamma, sigma_gamma, mu, sigma).shape
+
     # draw gamma and epsilon from truncated normal -- eq.s (10)-(11)
     # first sample unbounded normal, then rejection sample truncation
     eps = rng.normal(mu, sigma, size=size)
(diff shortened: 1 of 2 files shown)


glass/shapes.py
```diff
@@ -138,6 +138,10 @@ def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None):
     if rng is None:
         rng = np.random.default_rng()
 
+    # default size if not given
+    if size is None:
+        size = np.broadcast(gamma, sigma_gamma, mu, sigma).shape
```
paddyroddy, line 143: https://github.com/skypyproject/skypy/blob/ee05b5aba9118663c68fc69ffae523bc0e450e8a/skypy/galaxies/morphology.py#L265-L267

tests/test_shapes.py
```diff
@@ -67,7 +67,13 @@ def test_ellipticity_ryden04():
     e1 = ellipticity_ryden04(-1.85, 0.89, [0.222, 0.333], 0.056)
     e2 = ellipticity_ryden04(-1.85, 0.89, 0.222, [0.056, 0.067])
     e3 = ellipticity_ryden04([-1.85, -2.85], 0.89, 0.222, 0.056)
-    assert np.shape(e1) == np.shape(e2) == np.shape(e3) == (2,)
+    e4 = ellipticity_ryden04(-1.85, [0.89, 1.001], 0.222, 0.056)
+    assert np.shape(e1) == np.shape(e2) == np.shape(e3) == np.shape(e4) == (2,)
```
paddyroddy, line 71: This seems a bit silly, is there a better way of doing this? Possibly related to #327

paddyroddy (approved):
Hopefully this helps

tests/test_shapes.py
Saransh-cpp: The only thing I can think of is using `eX.shape` 🤔 no other helpful methods in `np.testing`.

Issue #323, opened 2024-10-04
Add support for `uv` in the CI


Saransh-cpp:
### Is Your Feature Request Related to a Problem? Please Describe

At the moment, glass uses [test-constraints.txt](https://github.com/glass-dev/glass/blob/main/.github/test-constraints.txt) to specify constraints for installing `test` dependencies. The file specifies `--prefer-binary` flag which is not supported by `uv` - https://github.com/astral-sh/uv/issues/1794 - blocking `uv`'s support.

### Describe the Solution You'd Like

Supporting `uv` in `nox` would require adding -
```py
nox.options.default_venv_backend = "uv|virtualenv"
```
in `noxfile.py`.

connoraird:
@paddyroddy mentioned a possible reason for not being able to use UV is the use of `--only-binary`. I was wondering if this is an old issue that has now been solved? [These docs](https://docs.astral.sh/uv/pip/compatibility/#-only-binary-enforcement) seem to suggest it would work in essentially the same way uasing uv or pip

connoraird:
Ah `prefer` not just `only`

connoraird:
Could we use this partial solution since it is just for testing? https://github.com/astral-sh/uv/issues/1794#issuecomment-2163884060

paddyroddy:
My other question would be how much https://github.com/glass-dev/glass/blob/main/.github/test-constraints.txt actually adds @ntessore

PR #324, opened 2024-10-04
gh-323: rm broken support for uv + update constraint format

Saransh-cpp:
`Nox` will always use `virtualenv` now and the `.txt` file for constraints is in a more widely accepted format now (readable by both `uv` and plain `pip install`.

Refs: #323

### DIFF
--- a/.github/test-constraints.txt
+++ b/.github/test-constraints.txt
@@ -1,2 +1,13 @@
---only-binary numpy,scipy,healpy,healpix
+numpy
+--only-binary numpy
+
+scipy
+--only-binary scipy
+
+healpy
+--only-binary healpy
+
+healpix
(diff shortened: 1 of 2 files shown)


paddyroddy (approved):
Thanks for sorting. What does the `--only-binary` thing do? I've seen it before

Saransh-cpp:
[The `--only-binary` argument is used to restrict installation to pre-built binary distributions.](https://docs.astral.sh/uv/pip/compatibility/#-only-binary-enforcement)

Tho I don't know if it is required here. I have never used it too.

ntessore:
In the past, we had new/incompatible versions of dependencies try to build e.g. numpy in the CI, which took forever and eventually failed. So this just made it fail early.

Issue #67, opened 2023-03-03
Array API compatibility


ntessore:
As much as possible of the core library should be made to work with compatible array types (numpy, torch, jax, dask, cupy, ...).

There is a nascent standard to support this: [Array API](https://data-apis.org/array-api/)

For the time being, the [array-api-compat](https://github.com/data-apis/array-api-compat) library can be used to get a universal namespace that replaces `np`, `jnp`, etc.

ntessore:
This is work in progress under the [glass-compat](https://github.com/glass-dev/glass-compat) branch, to be merged back into the core library when ready.

PR #326, opened 2024-10-04
gh-198: enforce `python>3.8` & `numpy>1.21`

### LINKED ISSUE #198: Should we drop `3.8-3.9` support?

paddyroddy:
Closes #198. For Array API support (#67), we need minimum versions of:

1. `Python 3.9` as specified here https://data-apis.org/array-api-compat/changelog.html#id9
2. `NumPy 1.22` as specified here https://numpy.org/doc/stable/reference/array_api.html#array-api-standard-compatibility

In this PR I have enforced both of these. The code changes have come from `ruff` linting, utilising the benefits from [3.9](https://docs.python.org/3/whatsnew/3.9.html).

I also removed some of the overzealous comments on imports. They weren't adding anything and meant the imports took up more room.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.8"
           - "3.9"
           - "3.10"
           - "3.11"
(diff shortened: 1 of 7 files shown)


Saransh-cpp (approved):
Amazing, thank @paddyroddy!

Issue #328, opened 2024-10-04
`ellipticity_ryden04` has a shape mismatch bug


Saransh-cpp:
### Describe the Bug

The implementation right now keeps sampling values until it gets all positive values -

https://github.com/glass-dev/glass/blob/2eb3e8bd94adf5f25c67a54fba0484a49c38b212/glass/shapes.py#L142-L151

Let's say if `mu = [-1.85, -2.85]`, `gamma = 0.89`, `gamma = 0.222`, `sigma_gamma = 0.056` -

- `size` = `(2,)`
- `eps.shape` = `(2,)`
- `rng.normal(mu, sigma)` will have the shape `(2,)` because numpy broadcasts the parameters internally

If one value in `eps` is `bad` (`<0`) -

- `bad` = `[True, False]`
- `eps[bad].shape` = `(1,)`

then this will not work - 

```py
eps[bad] = rng.normal(mu, sigma, size=eps[bad].shape)
```

as a `(2,)` array cannot be resized to `(1,)`.

---

This bug was hidden by the `float value is not indexable` error when the broadcasting rule was not present.

I am not sure what should be done here. Should we consider using something like truncated normal here?

### To Reproduce

```python
from glass import ellipticity_ryden04


while True:
    ellipticity_ryden04([-1.85, -2.85], 0.89, 0.222, 0.056)  # rerun till it errors out
```

### Expected Behaviour

No shape mismatch error.

### Actual Behaviour

```py
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
      1 while True:
----> 2     ellipticity_ryden04([-1.85, -2.85], 0.89, 0.222, 0.056)

File ~/Code/UCL/glass/glass/shapes.py:146, in ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size, rng)
    144 bad = eps > 0
    145 while np.any(bad):
--> 146     eps[bad] = rng.normal(mu, sigma, size=eps[bad].shape)
    147     bad = eps > 0
    148 gam = rng.normal(gamma, sigma_gamma, size=size)

File numpy/random/_generator.pyx:1288, in numpy.random._generator.Generator.normal()

File _common.pyx:621, in numpy.random._common.cont()

File _common.pyx:539, in numpy.random._common.cont_broadcast_2()

File _common.pyx:250, in numpy.random._common.validate_output_shape()

ValueError: Output size (1,) is not compatible with broadcast dimensions of inputs (2,).
```

### Version In Use

main

[...]

PR #332, opened 2024-10-07
gh-328: fix shape mismatch bug in ellipticity_ryden04

Saransh-cpp:
Instead of trying to reshape an array to a smaller dimension, produce an array of the same `size` and mask it.

Fixes: #328 
Fixed: `ellipticity_ryden04` does not error our with shape mismatch sporadically

### DIFF
--- a/glass/shapes.py
+++ b/glass/shapes.py
@@ -140,19 +140,19 @@ def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None): 
 
     # default size if not given
     if size is None:
-        size = np.broadcast(gamma, sigma_gamma, mu, sigma).shape
+        size = np.broadcast(mu, sigma, gamma, sigma_gamma).shape
 
     # draw gamma and epsilon from truncated normal -- eq.s (10)-(11)
     # first sample unbounded normal, then rejection sample truncation
     eps = rng.normal(mu, sigma, size=size)
     bad = eps > 0
     while np.any(bad):
-        eps[bad] = rng.normal(mu, sigma, size=eps[bad].shape)
+        eps[bad] = rng.normal(mu, sigma, size=size)[bad]
         bad = eps > 0
     gam = rng.normal(gamma, sigma_gamma, size=size)
     bad = (gam < 0) | (gam > 1)
     while np.any(bad):
-        gam[bad] = rng.normal(gamma, sigma_gamma, size=gam[bad].shape)
+        gam[bad] = rng.normal(gamma, sigma_gamma, size=size)[bad]
         bad = (gam < 0) | (gam > 1)
 
     # compute triaxial axis ratios zeta = B/A, xi = C/A


glass/shapes.py
```diff
@@ -140,19 +140,19 @@ def ellipticity_ryden04(mu, sigma, gamma, sigma_gamma, size=None, *, rng=None):
 
     # default size if not given
     if size is None:
-        size = np.broadcast(gamma, sigma_gamma, mu, sigma).shape
+        size = np.broadcast(mu, sigma, gamma, sigma_gamma).shape
```
paddyroddy, line 143: Sorry what's going on here?

Saransh-cpp: I just did that to match the argument order in the function signature - 

https://github.com/glass-dev/glass/blob/59048a7d2fe6d981ab9232391b2792314241352a/glass/shapes.py#L100

nothing else 😬

paddyroddy: I see, okay. The order isn't important?

Saransh-cpp: Nope!

paddyroddy (approved):
Hopefully this solves it

ntessore:
This solves it, but at very high cost: the more shapes are being sampled, the higher the probability of a bad entry, and the higher the cost of resampling.

In light of this, I am tempted to say we only allow 1D arrays as inputs.

ntessore:
Can we not broadcast all inputs to `size` before sampling, then we can select `[bad]` on the input arrays?

```py
eps[bad] = rng.normal(mu[bad], sigma[bad])
```

Saransh-cpp:
Numpy will not be able to broadcast 1D arrays of different lengths to one shape -
```py
In [11]: mu, sigma, gamma, sigma_gamma = [-1.85, -2.85], [0.89, 0.89, 0.33], [0.056, 0.02, 0.3, 0.2], [0.222]

In [12]: size = np.broadcast(mu, sigma, gamma, sigma_gamma).shape
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[12], line 1
----> 1 size = np.broadcast(mu, sigma, gamma, sigma_gamma).shape

In [13]: np.broadcast_arrays(mu, sigma)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[13], line 1
----> 1 np.broadcast_arrays(mu, sigma)

File ~/Code/UCL/glass/.env/lib/python3.11/site-packages/numpy/lib/_stride_tricks_impl.py:558, in broadcast_arrays(subok, *args)
    551 # nditer is not used here to avoid the limit of 32 arrays.
    552 # Otherwise, something like the following one-liner would suffice:
    553 # return np.nditer(args, flags=['multi_index', 'zerosize_ok'],
    554 #                  order='C').itviews
    556 args = [np.array(_m, copy=None, subok=subok) for _m in args]
--> 558 shape = _broadcast_shape(*args)
    560 result = [array if array.shape == shape
    561           else _broadcast_to(array, shape, subok=subok, readonly=False)
    562                           for array in args]
    563 return tuple(result)

File ~/Code/UCL/glass/.env/lib/python3.11/site-packages/numpy/lib/_stride_tricks_impl.py:433, in _broadcast_shape(*args)
    428 """Returns the shape of the arrays that would result from broadcasting the
    429 supplied arrays against each other.
    430 """
    431 # use the old-iterator because np.nditer does not handle size 0 arrays
    432 # consistently
--> 433 b = np.broadcast(*args[:32])
    434 # unfortunately, it cannot handle 32 or more arguments directly
    435 for pos in range(32, len(args), 31):
    436     # ironically, np.broadcast does not properly handle np.broadcast
    437     # objects (it treats them as scalars)
    438     # use broadcasting to avoid allocating the full array

ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (2,) and arg 1 with shape (3,).

In [14]: np.broadcast_to(mu, (3,))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[14], line 1
----> 1 np.broadcast_to(mu, (3,))

File ~/Code/UCL/glass/.env/lib/python3.11/site-packages/numpy/lib/_stride_tricks_impl.py:424, in broadcast_to(array, shape, subok)
    377 @array_function_dispatch(_broadcast_to_dispatcher, module='numpy')
    378 def broadcast_to(array, shape, subok=False):
    379     """Broadcast an array to a new shape.
    380 
    381     Parameters
   (...)
    422            [1, 2, 3]])
    423     """
--> 424     return _broadcast_to(array, shape, subok=subok, readonly=True)

File ~/Code/UCL/glass/.env/lib/python3.11/site-packages/numpy/lib/_stride_tricks_impl.py:359, in _broadcast_to(array, shape, subok, readonly)
    356     raise ValueError('all elements of broadcast shape must be non-'
    357                      'negative')
    358 extras = []
--> 359 it = np.nditer(
    360     (array,), flags=['multi_index', 'refs_ok', 'zerosize_ok'] + extras,
    361     op_flags=['readonly'], itershape=shape, order='C')
    362 with it:
    363     # never really has writebackifcopy semantics
    364     broadcast = it.itviews[0]

ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (2,)  and requested shape (3,)
```

but, if we have arrays of same length -
```py
In [15]: mu, sigma, gamma, sigma_gamma = [-1.85, -2.85], [0.89, 0.89], [0.056, 0.02], [0.222, 0.3]

In [16]: eps = rng.normal(mu, sigma)

In [17]: eps
Out[17]: array([ 0.07663722, -2.76959735])

In [18]: eps[bad] = rng.normal(np.array(mu)[bad], np.array(sigma)[bad])

In [19]: eps
Out[19]: array([-1.0051209 , -2.76959735])
```

We can alternatively append `0`s to match the shapes, but that would be inefficient too.

ntessore:
Of course, the initial shapes must be broadcastable, but that is already true.

I am imagining this:

```py
    # default size if not given
    if size is None:
        size = np.broadcast(mu, sigma, gamma, sigma_gamma).shape

    # broadcast all inputs to output shape
    # this makes is possible to resample later
    mu = np.broadcast_to(mu, size, True)
    sigma = np.broadcast_to(sigma, size, True)
    gamma = np.broadcast_to(gamma, size, True)
    sigma_gamma = np.broadcast_to(sigma_gamma, size, True)

    # draw gamma and epsilon from truncated normal -- eq.s (10)-(11)
    # first sample unbounded normal, then rejection sample truncation
    eps = rng.normal(mu, sigma)
    while np.any(bad := eps > 0):
        eps[bad] = rng.normal(mu[bad], sigma[bad])
    gam = rng.normal(gamma, sigma_gamma, size=size)
    while np.any(bad := (gam < 0) | (gam > 1)):
        gam[bad] = rng.normal(gamma[bad], sigma_gamma[bad])
```

Saransh-cpp:
Maybe I am picking up the wrong cases (cases that scientists won't use). For instance, this would not be broadcastable -

```py
mu, sigma, gamma, sigma_gamma = [-1.85, -2.85], [0.89, 0.89, 0.33], [0.056, 0.02, 0.3, 0.2], [0.222]
```

But yes, using `np.broadcast_to` with the existing test cases should work. Let me create a PR for this and transfer the discussion there (I should have created a new issue for the discussion above 😬).

Issue #137, opened 2023-10-19
Denote that `redshifts_from_nz` is deprecated for direct use


mwiet:
Make note of [`galaxies.redshifts_from_nz`](https://github.com/glass-dev/glass/blob/main/glass/galaxies.py#L60) being deprecated for direct use by the user. The function may encourage the variation of the sample `nz` by the user, but it is fixed by [`shells.restrict`](https://github.com/glass-dev/glass/blob/main/glass/shells.py#L296) or [`shells.partition`](https://github.com/glass-dev/glass/blob/main/glass/shells.py#L334).

Instead, only [`galaxies.redshifts`](https://github.com/glass-dev/glass/blob/main/glass/galaxies.py#L33) should be employed directly by users.

PR #333, opened 2024-10-07
gh-137: deprecate `redshifts_from_nz` in favor of `redshifts`

Saransh-cpp:
I have started looking into tests for `galaxies.py` and this popped up during my research.

Closes: #137

### DIFF
--- a/glass/galaxies.py
+++ b/glass/galaxies.py
@@ -19,6 +19,7 @@ Functions
 
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -99,6 +100,10 @@ def redshifts_from_nz(
         samples from all populations.
 
     """
+    warnings.warn(
+        "redshifts_from_nz is deprecated; use redshifts instead", stacklevel=2
+    )
+
     # get default RNG if not given
     if rng is None:
         rng = np.random.default_rng()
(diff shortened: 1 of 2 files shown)


paddyroddy (approved):
Good spot

glass/galaxies.py
```diff
@@ -99,6 +103,13 @@ def redshifts_from_nz(
         samples from all populations.
 
     """
+    if warn:
+        warnings.warn(
+            "when sampling galaxies, redshifts_from_nz() is often not the function you"
+            " want. Try redshifts() instead. Use warn=False to suppress this warning.",
+            stacklevel=2,
+        )
```
Saransh-cpp, line 111: Maybe we don't need the `warn` keyword, and we should let the ignoring part be at the user end  (users can suppress warnings with something like `import warnings; warnings.filterwarnings("ignore")`)

ntessore: They can, with significantly more effort :)

ntessore: But, yeah, I agree that this is slightly muddying the separation of concerns. My reason for proposing the parameter is that I don't think many users would a priori know how to suppress a warning. So either we would have to let them figure it out by themselves, or explain it in the docs of the function, at which point the users would be blindly copying code. It seems easier to take the hit and add the flag ourselves.

paddyroddy (approved):
👌

Issue #336, opened 2024-10-08
Add `Python 3.13` support


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

[3.13 was released today](https://docs.python.org/3/whatsnew/3.13.html)

PR #337, opened 2024-10-08
gh-336: support Python 3.13

Saransh-cpp:
- Add Python 3.13 classifier
- Test on Python 3.13 in CI
- Support for Python 3.13 in `noxfile`

Closes: #336


Saransh-cpp:
>  Version 3.13 was not found in the local cache
>  Error: The version '3.13' with architecture 'x64' was not found for Ubuntu 22.04.
>  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

I am not sure where the `cache` is coming from. I have tried re-running this multiple times and have also deleted the existing [caches from GitHub Actions](https://github.com/glass-dev/glass/actions/caches), but it somehow does not work. I will try re-running this later today, and maybe create an issue in `setup-python` if it still does not work? (Other pure-python repositories that I work on are not encountering this issue).

cc: @paddyroddy

paddyroddy:
It will take some time for `3.13` to become available across GitHub

paddyroddy:
> Ah, `healpix` does not have `3.13` binaries on PyPI and we don't allow pip to build it from source -

We should definitely wait until all dependencies have `3.13` support. It is quite common for the ecosystem to take a while to catch up.

ntessore:
healpix for cp313 has been uploaded

paddyroddy:
Just marking as ready to see what happens in CI

ntessore:
Also built `flt` wheels for cp312 and cp313. Excluding healpy, are these all of our native extension dependencies done?

paddyroddy:
> Also built `flt` wheels for cp312 and cp313. Excluding healpy, are these all of our native extension dependencies done?

I think so, hard to tell really unless we comment out `healpy`

ntessore:
Still waiting on a new healpy release does make me wonder if our CI wouldn't be better served by using conda ... But this probably causes a lot of issues with tox/nox?

paddyroddy:
> Still waiting on a new healpy release does make me wonder if our CI wouldn't be better served by using conda ... But this probably causes a lot of issues with tox/nox?

Unless I've missed something, `conda` alone won't sort `3.13` support? It's still on the same version `v1.17.3` https://anaconda.org/conda-forge/healpy, so will still need the library fixing.

Some libraries take a long time to support a new release, for example check out https://github.com/numba/numba/issues/9197.

A different question is do we need `healpy` as well as `healpix`?

paddyroddy:
> Still waiting on a new healpy release does make me wonder if our CI wouldn't be better served by using conda ... But this probably causes a lot of issues with tox/nox?

I do think we want to use whatever you anticipate end users using. I don't use `conda` these days in favour of [uv](https://github.com/astral-sh/uv).

paddyroddy (approved):

ntessore (approved):

Saransh-cpp:
@paddyroddy, is this now blocked by something?

paddyroddy:
> @paddyroddy, is this now blocked by something?

Nope, sorry! I think I forgot this was only adding `3.13` support, not removing `3.9`.

Issue #338, opened 2024-10-08
Use `pre-commit: ` in pre-commit PR titles


Saransh-cpp:
Right, I was really just wondering where the 5 characters `MNT: ` are coming from. Since it's not used anywhere else, we could remove them, or change it to `maint: ` or `ci: ` or `pre-commit: ` or something else?

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/335#issuecomment-2399400177_

PR #339, opened 2024-10-08
gh-338: better autoupdate and autofix messages for pre-commit.ci

Saransh-cpp:
`MNT` and `STY` keywords are not used anywhere in glass now, so it makes sense to make these automated commits and PRs have a more meaningful message.

Closes: #338


paddyroddy (approved):
Looks good

Issue #342, opened 2024-10-09
CI is broken


Saransh-cpp:
### Add your issue here

Two different error messages in different jobs.

[The first one](https://github.com/glass-dev/glass/actions/runs/11256548887/job/31298841621) -

```
  pip install nox
  shell: /usr/bin/bash -e {0}
  env:
    FORCE_COLOR: 1
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
```

and [the second one](https://github.com/glass-dev/glass/actions/runs/11256548887/job/31298843631) -

```
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [187 lines of output]
      /home/runner/work/glass/glass/.nox/doctests-3-11/lib/python3.11/site-packages/setuptools/__init__.py:94: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
      !!
      
              ********************************************************************************
              Requirements should be satisfied by a PEP 517 installer.
              If you are using pip, you can try `pip install --use-pep517`.
              ********************************************************************************
      
      !!
        dist.fetch_build_eggs(dist.setup_requires)
      running bdist_wheel
      running build
      running build_py
      creating build/lib.linux-x86_64-cpython-311/fitsio
      copying fitsio/header.py -> build/lib.linux-x86_64-cpython-311/fitsio
      copying fitsio/fits_exceptions.py -> build/lib.linux-x86_64-cpython-311/fitsio
      copying fitsio/util.py -> build/lib.linux-x86_64-cpython-311/fitsio
      copying fitsio/fitslib.py -> build/lib.linux-x86_64-cpython-311/fitsio
[...]

PR #308, opened 2024-10-03
gh-358: add static types and fix (some) `mypy`

paddyroddy:
Closes #280. I anticipate this being a fair bit of work. I will use `# type: ignore[ ]` syntax throughout to minimise the initial work. I will raise a similar issue to #248 to address this.

Issue #174, opened 2024-09-10
Consider @renovatebot instead of @dependabot


paddyroddy:
https://paddyroddy.github.io/talks/renovate-automating-dependency-management

paddyroddy:
Closing again. I don't like @dependabot as it isn't as powerful, but seems to be okay for our current use case.

PR #343, opened 2024-10-09
gh-342: install libbz2-dev and use setup-python

Saransh-cpp:
All of a sudden `libbz2-dev` is not present by default in GH Actions now + we were using system Python for a couple of jobs.

Closes: #342

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install nox and coverage.py
         run: pip install coverage[toml] nox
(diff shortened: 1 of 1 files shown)


paddyroddy (approved):
Good spot, thanks for fixing. Have you found the relevant GitHub issue?

paddyroddy:
Well it still appears to be there https://github.com/actions/python-versions/blob/48bcef97643138997e5935f3ae147629b0f931c4/builders/ubuntu-python-builder.psm1#L73 so we may be able to revert this at some point

paddyroddy:
Remind me, I will apply #174 after I finish #308

Issue #270, opened 2024-09-25
Add/review tests for `glass.galaxies`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Part of #161

PR #344, opened 2024-10-09
gh-270: add missing tests for glass.galaxies

Saransh-cpp:
Add missing tests for `glass.galaxy_shear`

Closes: #270

### DIFF
--- a/tests/test_galaxies.py
+++ b/tests/test_galaxies.py
@@ -96,9 +96,9 @@ def test_galaxy_shear(rng):
     # check shape of the output
 
     kappa, gamma1, gamma2 = (
-        rng.normal(size=(256,)),
-        rng.normal(size=(256,)),
-        rng.normal(size=(256,)),
+        rng.normal(size=(12,)),
+        rng.normal(size=(12,)),
+        rng.normal(size=(12,)),
     )
 
     shear = galaxy_shear([], [], [], kappa, gamma1, gamma2)
--- a/tests/test_galaxies.py
+++ b/tests/test_galaxies.py
@@ -93,39 +93,73 @@ def test_redshifts_from_nz():
 
 
 def test_galaxy_shear(rng):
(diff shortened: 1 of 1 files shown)


Saransh-cpp:
@ntessore should the output size -

https://github.com/glass-dev/glass/blob/1631757e7436a89501497aff3f8592a4729294f0/glass/galaxies.py#L181-L185

of the variables also depend on the size of `kappa`,` gamma1`, and `gamma2`? Given that if these arguments are two dimensional then the shear should be two dimensional as well (two values for each galaxy)? Or maybe I misunderstood something.

ntessore:
> @ntessore should the output size of the variables also depend on the size of `kappa`,` gamma1`, and `gamma2`?

These variables — `kappa`,` gamma1`, and `gamma2` — are HEALPix maps. Their size is `12 * nside**2` and depends on the resolution parameter `nside`.

The variables `lon`, `lat`, and `eps` are the positions and shapes of a set of galaxies. Their sizes are arbitrary.

The function looks up the value of `kappa`,` gamma1`, and `gamma2` at the position of the galaxy, then carries out a computation. The output size therefore only depends on `lon`, `lat`, and `eps`.

A situation that could happen is that the user provides a stack of maps, in which case `kappa`,` gamma1`, and `gamma2` would have shape `(*nmaps, 12 * nside**2)`. In principle, the function could then return values for all of these maps. For example, if `lon`, `lat`, and `eps` have shape `ngals`, the overall output shape could be `(*nmaps, *ngals)`. But if that is not currently implemented, then it's not terribly important, either.

Saransh-cpp:
Oh I see, thanks for the detailed explanation! The case with output shape `(*nmaps, *ngals)` is not implemented yet so I'll remove the tests for that.

paddyroddy (approved):
Coverage almost at 60%!

Issue #131, opened 2023-09-27
Rename `gaussian_gls` function to `discretized_cls`


ntessore:
The `gaussian_gls()` function does not compute "Gaussian" Gls as such, but rather applies the angular discretisation (`lmax`, `ncorr`, `nside`) to the Cls it is given. The function is also used as such by other functions, e.g. `lognormal_gls()`, where the current name makes little sense. The function should therefore be renamed to `discretized_cls()` to better reflect its purpose.

Issue #334, opened 2024-10-07
Explain effective redshift distribution in docs


ntessore:
### Add your issue here

Add an explanation about the effective redshift distribution of sources when sampling from the discretised matter density shells.

This should help clear up the confusion of `redshifts_from_nz()` vs. `redshift()`.

Saransh-cpp:
> This link could potentially appear in the warning text in the future.
 
-- https://github.com/glass-dev/glass/pull/333#issuecomment-2397221853

ntessore:
Done in #567.

PR #345, opened 2024-10-09
gh-131: rename `gaussian_gls` to `discretized_cls`

Saransh-cpp:
This popped up while I was looking to add tests for `fields.py`.

- rename `gaussian_gls` to `discretized_cls`
- update docstring and internal references
- don't use `gaussian_gls` in `lognormal_gls`

Closes: #131
Changed: `gaussian_gls` has been renamed to `discretized_cls`
Changed: `lognormal_gls` no longer uses `gaussian_gls` internally and the arguments passed to `gaussian_gls` have been removed


examples/1-basic/cls.npy
paddyroddy, line 1: This should be in `.gitignore`

paddyroddy (changes requested):
Looks good to me, although @ntessore should definitely have the final say

ntessore:
One critical thing that is missing here is that, by taking `discretized_cls()` out of `lognormal_gls()`, the discretisation is no longer applied to the `cls` used in the examples.

I think the "discretisation step" would be best shown in each example. Something along the lines of:

```diff
-# compute Gaussian cls for lognormal fields for 3 correlated shells
-gls = glass.lognormal_gls(cls, nside=nside, lmax=lmax, ncorr=3)
+# apply discretisation to the full set of spectra:
+# - HEALPix pixel window function (`nside=nside`)
+# - maximum angular mode number (`lmax=lmax`)
+# - number of correlated shells (`ncorr=3`)
+cls = glass.discretized_cls(cls, nside=nside, lmax=lmax, ncorr=3)
+
+# compute Gaussian cls for lognormal fields for 3 correlated shells
+gls = glass.lognormal_gls(cls)
```

As a matter of fact, since notebooks are run locally and their outputs stored (i.e., `cls.npz`), it probably no longer makes sense to pre-compute the Cls for other examples; that was only done to save time when running on GitHub actions.

Saransh-cpp:
Thanks for the help! I have updated the notebooks to show the discretization step.

examples/1-basic/shells.ipynb
ntessore, line 1: I wouldn't remove this example entirely, as it's still useful as a first step towards understanding how the simulations are set up. In fact, I would incorporate #334 here in the future.

ntessore:
Thanks for the great work @Saransh-cpp, and sorry for the annoying change requests. Here is one more: The `## Setup` section still mentions that the examples use pre-computed spectra. (The text saying that the set up matches `shells` can stay and is probably useful to guide newcomers to a more basic example.)

Saransh-cpp:
No worries, @ntessore! I have updated the PR. Could you please take a look again?

ntessore:
Unrelated to this PR, but we should tweak the "Test examples" action to not just run the examples, but also produce and upload HTML artefacts of the results.

Saransh-cpp:
Thanks! I have created relevant issues for the comments above.

examples/1-basic/shells.ipynb
paddyroddy: Not sure if this is a GitHub rendering issue. I can't see the changes
image

ntessore: Only in "Files changed" for me too when it comes to notebooks!

ntessore: Same, but I can see the full file w/ changes under "View file"

ntessore: And the source diff!

ntessore (approved):

paddyroddy (approved):

Issue #188, opened 2024-09-10
Add docstrings to all functions


paddyroddy:
Currently, most have them but those that are lacking do stick out in the documentation - which automatically pulls from them

PR #21, opened 2022-07-27
Inclusion of the NLA model for intrinsic alignments

mwiet:
The ia_nla function updates the kappa maps by adding the convergence from intrinsic alignments calculated using the Non-linear Aligments Model (NLA) to the convergence maps purely from lensing. The convergence from intrinsic aligments, $\kappa_{\rm{IA}}$, is given by ([Catelan et al. 2000](https://doi.org/10.48550/arXiv.astro-ph/0005470), [Hirata & Seljak 2004](https://doi.org/10.48550/arXiv.astro-ph/0406275), [Bridle & King 2007](https://doi.org/10.48550/arXiv.0705.0166))

$\kappa_{\rm{IA}} = - A_{\rm{IA}} \frac{C_{1} \overline{\rho}(z)}{\overline{D}(z)} \bigg(\frac{1+z}{1+z_{0}}\bigg)^{\eta} \bigg(\frac{L}{L_{0}}\bigg)^{\beta} \delta(z)$

where $A_{\rm{IA}}$ is the intrinsic alignments amplitude, $C_{1}$ is a normalisation constant set to $C_{1} = 5 \times 10^{-14} (h^{2} M_{\odot}/Mpc^{-3})^{-2}$ ([Hirata & Seljak 2004](https://doi.org/10.48550/arXiv.astro-ph/0406275)), $\overline{\rho}$ is the mean matter density of the Universe at redshift $z$, $\overline{D}(z)$ is the growth factor, $\eta$ is the power of the power law which describes the redshift-dependence of the IA with respect to $z_{0}$, $\beta$ is the power of the power law which describes the dependence on luminosity, $L$, of the IA with respect to $L_{0}$, and $\delta(z)$ is the matter overdensity field at redshift $z$.

PR #347, opened 2024-10-09
gh-346: enforce `NumPy` docstrings + use static typing in docs

### LINKED ISSUE #346: Deduplicate type information in docstrings

The problem at the moment is that we have type in the docstrings, and through #308 we are getting static types. However, sometimes there is a mismatch, and the bigger problem is that it is two duplicate things to maintain. If one removes the type as it is currently (I believe `numpy` style) then we trigger [ruff D417](https://docs.astral.sh/ruff/rules/undocumented-param/). We can enforce the docstring convention through `ruff` https://docs.astral.sh/ruff/settings/#lintpydocstyle.

They look like https://google.github.io/styleguide/pyguide.html#244-decision
```python
"""Connects to the next available port.

Args:
  minimum: A port value greater or equal to 1024.

Returns:
  The new minimum port.

Raises:
  ConnectionError: If no available port is found.
"""
```
The documentation will still show the types.

_No response_

### Additional Context

_No response_

paddyroddy:
Closes #346. Working towards #188.

The main changes here have been:
- Remove all types in the docstrings in favour of proper static typing (being added in #308)
- Switch from `numpydoc` to `sphinx.ext.napolean` due to https://github.com/numpy/numpydoc/issues/196

Have had to change the references to `[1]` rather than `[1]_` due to this bug, https://github.com/sphinx-doc/sphinx/issues/9689. Hopefully this can be fixed in the future.

Example:
`glass.lensing.from_convergence`
image

Raised #350, #351.

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -5,7 +5,7 @@ Random fields
 .. currentmodule:: glass
 
 The following functions provide functionality for simulating random
-fields on the sphere.  This is done in the form of HEALPix maps.
+fields on the sphere. This is done in the form of HEALPix maps.
 
 Functions
 ---------
@@ -163,7 +163,7 @@ def gaussian_gls(
 
     Depending on the given arguments, this truncates the angular power spectra
     to ``lmax``, removes all but ``ncorr`` correlations between fields, and
-    applies the HEALPix pixel window function of the given ``nside``.  If no
+    applies the HEALPix pixel window function of the given ``nside``. If no
     arguments are given, no action is performed.
 
     """
@@ -221,7 +221,7 @@ def generate_gaussian(
     ``nside``.
 
     The optional argument ``ncorr`` can be used to artificially limit now many
-    realised fields are correlated.  This saves memory, as only `ncorr` previous
+    realised fields are correlated. This saves memory, as only `ncorr` previous
     fields need to be kept.
 
     The ``gls`` array must contain the auto-correlation of each new field
@@ -327,19 +327,14 @@ def getcl(cls, i, j, lmax=None):
     Return the angular power spectrum for indices *i* and *j* from an
     array in *GLASS* ordering.
 
-    Parameters
-    ----------
-    cls : list of array_like
-        List of angular power spectra in *GLASS* ordering.
-    i, j : int
-        Combination of indices to return.
-    lmax : int, optional
-        Truncate the returned spectrum at this mode number.
+    Args:
+        cls: List of angular power spectra in *GLASS* ordering.
+        i: Indices to return.
+        j: Indices to return.
+        lmax: Truncate the returned spectrum at this mode number.
 
     Returns:
-    -------
-    cl : array_like
-        The angular power spectrum for indices *i* and *j*.
+        cl: The angular power spectrum for indices *i* and *j*.
 
     """
     if j > i:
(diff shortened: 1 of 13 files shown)


paddyroddy:
Reminder I need to update https://github.com/glass-dev/glass/pull/21

glass/shells.py
Saransh-cpp, line 1: Given this file has a "non-conditional" import of `numpy.typing`, it renders (ugly imo) cross-references to `numpy` types -

image

Should we should stick to one thing? Either import `numpy.typing` without a condition or with. If we want to add the condition, we can move the types declared in this file to a utility file (`_typeutils.py` or something).

glass/fields.py
```diff
@@ -329,16 +329,18 @@ def getcl(cls, i, j, lmax=None):
...
+        Indices to return.
+    j:
+        Indices to return.
+    lmax:
         Truncate the returned spectrum at this mode number.
 
     Returns
     -------
-    cl : array_like
+    cl:
```
Saransh-cpp, line 343: The return values for napoleon + numpy requires specifying both the name and type of the variable - https://numpydoc.readthedocs.io/en/latest/format.html#returns. Kind of annoying, but right now sphinx does not render the return values properly - 

image

Saransh-cpp (changes requested):
Thanks, @paddyroddy! A quick glance -

glass/fields.py
ntessore: I think the issue here is that it treats "Returns" and "Return type" as independent information. In which case we probably want to remove "Returns" from the docstrings?

glass/shells.py
paddyroddy: I had seen these. The condition meaning the `TYPE_CHECKING` block? It does feel a bit unncessary so happy to remove

paddyroddy: That file it also cannot be in the `TYPE_CHECKING` due to the custom types...

glass/fields.py
paddyroddy: So yeah it seems that we are forced to specify types in the docstrings. I think I'm going to move the returns bits to the summary sections.

paddyroddy: > Kind of annoying, but right now sphinx does not render the return values properly -

`cl` isn't a type, it think it is

glass/shells.py
paddyroddy: Can you raise a separate issue for this? I think this is out of scope of the intention of this PR

glass/fields.py
paddyroddy: Have moved `Returns` and that has works alright, I think

paddyroddy:
So I have removed `Returns` and put those in the main summary. Here are screenshots from the 3 bits of documentation referred to above. This is not meant to be the final version of the docs (#188), it's a work in progress. This should make #308 easier.

image
image
image

glass/fields.py
```diff
@@ -329,18 +329,15 @@ def getcl(cls, i, j, lmax=None):
...
+        Indices to return.
+    j:
+        Indices to return.
+    lmax:
         Truncate the returned spectrum at this mode number.
 
-    Returns
-    -------
-    cl : array_like
-        The angular power spectrum for indices *i* and *j*.
```
paddyroddy, line 342: This was already specified above

glass/shells.py
Saransh-cpp: Yes, sorry, I meant it to be worked in a separate PR. I'll raise an issue.

Saransh-cpp (approved):
Thanks for the amazing work on this, @paddyroddy! :)

PR #356, opened 2024-10-14
gh-355: consistent use of `typing` and `collections.abc`

### LINKED ISSUE #355: Tidy up `typing`

Going to consistently apply `typing` imports throughout. Simplifying the #308 PR.

_No response_

_No response_

### Additional Context

_No response_

paddyroddy:
Fixes #355, simplifying #308. Change all import of `typing`/`collections.abc` to `import <x>` rather than `from <x> import` syntax. This is useful because it is straightforward to see at a glance where the `import` is coming from. Further, we have both `numpy.random.Generator` and `collections.abc.Generator` in use - so it separates them. I've also moved everything out of the `TYPE_CHECKING` block, except from where `ruff` says we should.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ Issues = "https://github.com/glass-dev/glass/issues"
 
 [tool.coverage]
 report = {exclude_also = [
-    "if TYPE_CHECKING:",
+    "if typing.TYPE_CHECKING:",
 ], omit = [
     "glass/_version.py",
 ], skip_covered = true, sort = "cover"}
--- a/glass/core/algorithm.py
+++ b/glass/core/algorithm.py
@@ -2,12 +2,8 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
 import numpy as np
-
-if TYPE_CHECKING:
-    import numpy.typing as npt
+import numpy.typing as npt
 
 
 def nnls(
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -69,7 +69,7 @@ Issues = "https://github.com/glass-dev/glass/issues"
 
 [tool.coverage]
 report = {exclude_also = [
-    "if TYPE_CHECKING:",
+    "if typing.TYPE_CHECKING:",
```
paddyroddy, line 72: This is important, we might potentially want to add both in case someone doesn't use the format this PR is trying to enforce

Saransh-cpp (approved):
Thanks, @paddyroddy! Looks good!

ntessore:
I really don't like this change, as it goes against established typing practice (e.g., https://typing.readthedocs.io).

paddyroddy:
It also clutters the namespace https://softwareengineering.stackexchange.com/a/187432/320391

ntessore:
> It also clutters the namespace https://softwareengineering.stackexchange.com/a/187432/320391

Only if the imports are not guarded by the future import and `if TYPE_CHECKING:`. I usually don't have much, if any, of the static type present at runtime.

paddyroddy:
Have raised https://github.com/glass-dev/glass/issues/361

Issue #352, opened 2024-10-10
Latex blocks are somehow broken in the `latest` docs


Saransh-cpp:
### Which Section of the Documentation Needs Improving?

See the docstring of https://glass.readthedocs.io/latest/reference/shells.html#glass.partition

![Image](https://github.com/user-attachments/assets/af50cd88-438e-4890-9f99-be179109f563)

### What Can be Improved About This Section

Fix latex :(

ntessore:
Working: https://glass--237.org.readthedocs.build/237/reference/shells.html#glass.shells.partition
Broken: https://glass--241.org.readthedocs.build/241/reference/shells.html#glass.partition

Issue #231, opened 2024-09-17
Add `ruff` linting


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Following #199, adding in a separate PR to separate linting and formatting

PR #357, opened 2024-10-14
gh-352: fix latex rendering

Saransh-cpp:
Single backslash is for latex commands, double backslash is for newline. I don't know when or how the behavior changed, [but their latest docs still say](https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/math.html#module-sphinx.ext.mathbase) -

> Keep in mind that when you put math markup in Python docstrings read by [autodoc](https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/autodoc.html#module-sphinx.ext.autodoc), you either have to double all backslashes, or use Python raw strings (r"raw").

Closes: #352

### DIFF
--- a/glass/lensing.py
+++ b/glass/lensing.py
@@ -438,9 +438,9 @@ def deflect(
     Notes
     -----
     Deflections on the sphere are :term:`defined <deflection>` as
-    follows:  The complex deflection :math:`\\alpha` transports a point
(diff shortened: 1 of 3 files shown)


paddyroddy (approved):
Thanks for fixing. I suspect the changes happened in #231, whereby it would have prompted us to use raw strings.

Issue #358, opened 2024-10-14
Add static types


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

I've split #280 up into two issues. This issue is for specifically adding static typing - and fixing some `mypy` errors where possible.

Issue #330, opened 2024-10-07
Change `npt.ArrayLike` to `npt.NDArray[<type>]` where appropriate


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Currently, we have a lot of `npt.ArrayLike`, but we could benefit more from type checking through `mypy` #280 and using the `npt.NDArray[type]`.

paddyroddy:
This is being done #308

PR #359, opened 2024-10-14
gh-280: add `mypy` to codebase

paddyroddy:
Adding `mypy` to a pre-existing codebase is never easy. I initially attempted this in #308, but have since split this up into separate issues:
- [x] #347
- [x] #356
- [ ] #358

In this PR, `mypy` is added but in order for CI to pass, the `# type: ignore[]` syntax is used throughout. I didn't want to tackle these here too (see #308) as it gets quite messy.

One thing I have done (following #356) is change every empty `npt.NDArray` to `npt.NDArray[typing.Any]` (see #330), as this actually results in fewer errors than leaving them all blank. Ideally, we'd like to fill in as many of the `typing.Any` as possible - they're a bit useless by themselves. However, that is not the priority for now. Plus, I expect typing to change when #67 is tackled.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,19 @@ build.targets.sdist.exclude = [
 ]
 version.source = "vcs"
 
+[tool.mypy]
+disallow_untyped_decorators = false
+enable_error_code = [
+    "ignore-without-code",
+    "redundant-expr",
+    "truthy-bool",
+]
+plugins = [
+    "numpy.typing.mypy_plugin",
+]
+strict = true
+warn_unreachable = true
+
 [tool.pytest.ini_options]
 addopts = [
     "--strict-config",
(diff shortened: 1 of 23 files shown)


pyproject.toml
```diff
@@ -92,6 +92,19 @@ build.targets.sdist.exclude = [
 ]
 version.source = "vcs"
 
+[tool.mypy]
```
paddyroddy, line 95: Some of these are related to #187

pyproject.toml
```diff
@@ -92,6 +92,19 @@ build.targets.sdist.exclude = [
...
+    "ignore-without-code",
+    "redundant-expr",
+    "truthy-bool",
+]
+plugins = [
+    "numpy.typing.mypy_plugin",
+]
+strict = true
+warn_unreachable = true
+
```
paddyroddy, line 107: Some of these are related to #187

Saransh-cpp (approved):
This looks like a good start, thanks @paddyroddy!

Saransh-cpp:
Ah 😬 I meant a good start for the other PR(s) that will add more type informations 😉

Issue #248, opened 2024-09-18
Consider fixing most/all `noqa:` statements scattered throughout the code


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

To see the list of rules, have a look here, https://docs.astral.sh/ruff/rules. Some of these are trivial, but many are not. The majority of these rules are pretty sensible. The ones with the `TODO:` listed in #232, I suggest not fixing.

PR #360, opened 2024-10-14
gh-248: a few fixes for ignored and recommended ruff rules

Saransh-cpp:
Fixed:
- missing-trailing-comma
- commented-out-code
- single-line-implicit-string-concatenation
- numpy-legacy-random
- pytest-raises-too-broad

Refs: #248


pyproject.toml
```diff
@@ -160,11 +156,10 @@ lint.per-file-ignores = {"__init__.py" = [
     "ANN201", # TODO: issing-return-type-undocumented-public-function
     "ANN202", # TODO: missing-return-type-private-function
     "D100", # undocumented-public-module
-    "D103", # TODO: undocumented-public-function
```
Saransh-cpp, line 163: I think it should be okay to not have docstrings in the test functions

pyproject.toml
```diff
@@ -121,12 +121,8 @@ lint.ignore = [
     "ANN002", # TODO: missing-type-args
     "ANN003", # TODO: missing-type-kwargs
     "ANN201", # TODO: missing-return-type-undocumented-public-function
```
Saransh-cpp, line 123: Hopefully these will be fixed in the mypy PR

paddyroddy: Although, more accurately, the static types PR #308

pyproject.toml
```diff
@@ -121,12 +121,8 @@ lint.ignore = [
...
-    "COM812", # missing-trailing-comma (ruff-format recommended)
```
paddyroddy, line 124: You have to leave this one in, that is why I didn't write TODO... In fact, there are more than I thought - just found the docs https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules

pyproject.toml
```diff
@@ -121,12 +121,8 @@ lint.ignore = [
...
     "D203", # one-blank-line-before-class
     "D212", # blank-line-before-class
-    "ERA001", # TODO: commented-out-code
-    "ISC001", # single-line-implicit-string-concatenation (ruff-format recommended)
```
paddyroddy, line 128: You have to leave this one in, that is why I didn't write TODO... In fact, there are more than I thought - just found the docs https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules

examples/1-basic/photoz.ipynb
ntessore, line 1: If we are introducing a new concept into the example (creating a RNG), then it should ideally happen explicitly and somewhere near the top.

pyproject.toml
paddyroddy: I do think it's something we should work towards

examples/1-basic/photoz.ipynb
paddyroddy, line 1: Can we add a random seed to `default_rng`?

paddyroddy (changes requested):
Thanks for these, and I prefer the look of the comma options. However, see my concerns about `ruff-format` and conflicting rules

examples/1-basic/lensing.ipynb
ntessore, line 1: There are comments like this one in all of the examples, and they no longer make sense because now the user does not have to import N different modules.

Maybe we could comment something like

```py
# almost all GLASS functionality is available from the `glass` namespace
```

Saransh-cpp:
Thanks for the reviews! This should be better now.

pyproject.toml
paddyroddy: Could you make an issue to look into these conflicting rules?

paddyroddy (approved):
Looks good, thanks for sorting

Issue #363, opened 2024-10-15
`test_ellipticity_gaussian` is flaky


Saransh-cpp:
### Add your issue here

See the failing workflows on main - 
- https://github.com/glass-dev/glass/actions/runs/11345132033/job/31551339600
- https://github.com/glass-dev/glass/actions/runs/11329143777/job/31503990664

I think we should just increase the tolerance.

ntessore:
In the past, I have used https://github.com/pytest-dev/pytest-rerunfailures for these kinds of random tests.

Saransh-cpp:
Might be an overkill for a simple fix, but can look into this. (Looks very useful in such cases tho!)

@paddyroddy do you have any opinions on this?

PR #364, opened 2024-10-15
gh-363: fix flakiness of test_ellipticity_gaussian

Saransh-cpp:
Increase the tolerance for the test to pass always.

Closes: #363


paddyroddy (approved):
👌

ntessore (commented):
I disagree with this fix. For a non-truncated Gaussian, the chance of this test failing is something like 3.3e-8 -- if we are seeing noticeably more failures than that, I should investigate whether the function isn't actually broken. But for a sample size $n = 1\,000\,000$, we should be able to narrow the expected value down to closer than $O(1)/\sqrt{n}$.

Saransh-cpp:
Oops, should have waited for a review. If the chance of it failing is that low (maybe that's why we noticed it 1.5 months into the project), I will just revert and use `pytest-rerunfailures`.

Issue #365, opened 2024-10-15
Produce and upload HTML artifacts of the notebook runs


Saransh-cpp:
### Add your issue here

> Unrelated to this PR, but we should tweak the "Test examples" action to not just run the examples, but also produce and upload HTML artefacts of the results.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/345#issuecomment-2413721226_

I think it would be better to get rid off the "Test examples" session and just run them on readthedocs, given that one has to pre-run the notebooks in order for them to show up with good outputs on readthedocs.

ntessore:
We spent quite a lot of effort originally to not have the notebooks render on RTD, since that was unbearably slow.

What I meant was simply to replace

    jupyter execute examples/**/*.ipynb

with

    jupyter nbconvert --to html --embed-images --execute examples/**/*.ipynb

and then upload the resulting HTML files with the GitHub action.

Saransh-cpp:
The uploading HTML part is not very clear to me. Do you mean automatically committing the generated HTML output and pushing it to the GLASS repository? But that will not show up on readthedocs. I don't think there is a way to push to readthedocs?

ntessore:
I just meant

```yaml
- uses: actions/upload-artifact@v4
  with:
    path: examples/*.html
```

so we can check the outputs from the test run if necessary (since we are recomputing the examples anyway).

ntessore:
In the future, we could a) recompute the example notebooks in place, and fail the test if the outputs change, and b) have an automated way to recompute the examples. But neither is relevant _right now_ for what we need to do.

Saransh-cpp:
Ahh, now I understand it better, thanks!

PR #367, opened 2024-10-15
gh-365: upload HTML rendering of notebooks as artifact

Saransh-cpp:
The workflows now produce an HTML rendering of the notebook and upload it as artifact. I have tweaked the nox session to help with this. All the notebooks have been rerun to update the outputs with the recent changes.

Closes: #365


ntessore:
Great work @Saransh-cpp! Can you revert the changes to the .ipynb files, which will be nothing but updated timestamps and redrawn plots?

.github/workflows/examples.yml
```diff
@@ -45,3 +45,9 @@ jobs:
         run: pip install nox
       - name: Run examples
         run: nox -s examples
+      - name: Generate HTML
+        run: nox -s examples -- html
+      - uses: actions/upload-artifact@v4
+        with:
+          name: Examples
+          path: examples/**/*.html
```
ntessore, line 53: I think this now executes all of the examples twice: once with `jupyter execute examples/**/*.ipynb` and once with `jupyter nbconvert --execute`. You could either remove the first step, or I believe you can remove `--execute` from the second step.

ntessore:
Could you add a dummy file somewhere that triggers the "Test examples" step, to be removed before merging?

paddyroddy:
> Could we add this to `pre-commit`? This keeps the plots, and removes all metadata which doesn't actually get displayed on [glass.readthedocs.io](https://glass.readthedocs.io) anyway
> 
> ```yaml
> repos:
>   - repo: https://github.com/kynan/nbstripout
>     rev: 0.7.1
>     hooks:
>       - id: nbstripout
>         args:
>           - --keep-output
> ```

Have added this and raised https://github.com/glass-dev/glass/pull/380

Saransh-cpp:
The artifact upload worked - https://github.com/glass-dev/glass/actions/runs/11701933760

glass/remove_me.py
```diff
@@ -0,0 +1 @@
+"""to test example artifacts."""
```
paddyroddy, line 1: Should this be removed?

.github/workflows/examples.yml
```diff
@@ -52,3 +52,11 @@ jobs:
...
         env:
           FORCE_COLOR: 1
+      - name: Generate HTML
+        run: nox -s examples -- html
+        env:
+          FORCE_COLOR: 1
+      - uses: actions/upload-artifact@v4
+        with:
+          name: Examples
+          path: examples/**/*.html
```
paddyroddy, line 62: Just to keep in line with #388
```suggestion

      - name: Generate HTML
        run: nox -s examples -- html
        env:
          FORCE_COLOR: 1

      - name: Upload artefact
        uses: actions/upload-artifact@v4
        with:
          name: Examples
          path: examples/**/*.html
```

paddyroddy (changes requested):

glass/remove_me.py
Saransh-cpp: Yes, was added to trigger the tests.

paddyroddy (approved):
Looks good!

Issue #362, opened 2024-10-15
Review "ruff-format recommended" ignore rules


Saransh-cpp:
### Add your issue here

At the moment, we have several "ruff-format recommended" ignored rules in `pyproject.toml` -

https://github.com/glass-dev/glass/blob/cfe3245785164e4bfff9bb1786230960529237d1/pyproject.toml#L132

It would be good to review these and remove the ones that are not required to be ignored.

paddyroddy:
The reason why I only had those two is that the warning only refers to those

> warning: The following rules may cause conflicts when used with the formatter: `COM812`, `ISC001`. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the `select` or `extend-select` configuration, or adding them to the `ignore` configuration.

Saransh-cpp:
Keeping all of them ignored shouldn't be bad, but I don't have any strong opinions here. I can revert it back to the old state or just close this issue as resolved 👀

paddyroddy:
I think leave it, and leave the issue open

PR #368, opened 2024-10-15
gh-358: add static types support

### LINKED ISSUE #358: Add static types

paddyroddy:
Closes #358. Replaces #308.

This PR adds static type support, hence satisfying `mypy`. With the adaptation of #67 I anticipate typing to change, but it's much easier to do once `mypy` is passing initially.

The only `# type: ignore[]` comments remaining are `attr-defined` which are all attributed to `np.trapz`, those can easily be fixed by #207.

I have turned off:
- `disallow_untyped_decorators` this is being triggered by `pytest` and `nox`. The equivalent `# type: ignore[]` error is `misc` which isn't the most informative
- `warn_return_any` which is due to external libraries like `cosmology` - this could be revisited in the future

I have decided to re-write some of the tests which were simply not following the typing - like passing is a `list` rather than a `np.ndarray`. The alternative approach would be having really messy typing.

I'm not expecting this to be fully perfect. We will likely have to iterate on this, and completely change once we follow #67. But this is a good first step.

Do make sure to look carefully. Hard to avoid such a big PR with this. There were over 200 `# type: ignore[]` comments to fix!


paddyroddy:
FYI I'm leaving this non-draft as I accidentally broke the tests

.git-blame-ignore-revs
```diff
@@ -4,3 +4,4 @@ b42067f6776dcd763827000d585a88e071b78af3
 f9bac62f497a7288aa71fd4dbd948945c37f854e
 # applied ruff-linting - https://github.com/glass-dev/glass/pull/232
 e58d8616c03b8447aba90996905a98b42f18ba0a
+# added static typing - https://github.com/glass-dev/glass/pull/368
```
paddyroddy, line 7: A reminder I should add the hash once merged

pyproject.toml
```diff
@@ -104,6 +105,7 @@ plugins = [
     "numpy.typing.mypy_plugin",
 ]
 strict = true
+warn_return_any = false
```
paddyroddy, line 108: There are 17 errors which are ignored by this. They are almost all related to untyped external libraries. Do we want this here or the individual `# type: ignore[no-any-return]` comments?

tests/test_shells.py
```diff
@@ -18,16 +18,19 @@ def test_tophat_windows():  # type: ignore[no-untyped-def]
...
     # Gaussian test function
     z = np.linspace(0.0, 5.0, 1000)
     f = np.exp(-(((z - 2.0) / 0.5) ** 2) / 2)
 
     # window for restriction
-    w = RadialWindow(za=[1.0, 2.0, 3.0, 4.0], wa=[0.0, 0.5, 0.5, 0.0], zeff=None)  # type: ignore[arg-type]
+    w = RadialWindow(
+        za=np.array([1.0, 2.0, 3.0, 4.0]),
+        wa=np.array([0.0, 0.5, 0.5, 0.0]),
+    )
```
paddyroddy, line 33: I made the default `zeff` to be `0` as `None` didn't make sense as a type

tests/test_fits.py
```diff
@@ -1,37 +1,41 @@
...
 @pytest.mark.skipif(not HAVE_FITSIO, reason="test requires fitsio")
-def test_basic_write(tmp_path):  # type: ignore[no-untyped-def]
+def test_basic_write(tmp_path: pathlib.Path) -> None:
     import fitsio
 
     filename_gfits = "gfits.fits"  # what GLASS creates
     filename_tfits = "tfits.fits"  # file created on the fly to test against
 
+    def _test_append(
+        fits: fitsio.FITS,
```
paddyroddy, line 27: Move this here as I wanted `fitsio` for typing

glass/user.py
```diff
@@ -17,12 +17,24 @@
...
 import numpy as np
+import numpy.typing as npt
+
+if typing.TYPE_CHECKING:
+    import collections.abc
+    import importlib.util
+    import pathlib
+
+    if importlib.util.find_spec("fitsio") is not None:
+        import fitsio
```
paddyroddy, line 34: This is a bit hacky, but seems to work for `mypy`

glass/user.py
```diff
@@ -55,23 +67,32 @@ class _FitsWriter:
...
 
-    def _append(self, data, names=None) -> None:  # type: ignore[no-untyped-def]
+    def _append(
+        self,
+        data: npt.NDArray[np.float64] | list[npt.NDArray[np.float64]],
+        names: list[str] | None = None,
+    ) -> None:
         """Write the FITS file."""
-        if self.ext is None or self.ext not in self.fits:
+        if self.ext not in self.fits:
```
paddyroddy, line 81: Now `self.ext` defaults to `""` then no harm in having a blank extension

glass/fields.py
```diff
@@ -35,36 +34,22 @@
...
-# types
-Size = typing.Optional[typing.Union[int, tuple[int, ...]]]
-Iternorm = tuple[typing.Optional[int], npt.NDArray[typing.Any], npt.NDArray[typing.Any]]
-ClTransform = typing.Union[
-    str, typing.Callable[[npt.NDArray[typing.Any]], npt.NDArray[typing.Any]]
-]
-Cls = collections.abc.Sequence[
-    typing.Union[npt.NDArray[typing.Any], collections.abc.Sequence[float]]
-]
-Alms = npt.NDArray[typing.Any]
```
paddyroddy, line 47: I got rid of these as I don't really like custom types. Can return if there are strong feelings.

paddyroddy: Although I suspect `mypy` won't pass if all have the same type, i.e. `Cls` do vary depending on context

glass/shells.py
```diff
@@ -625,14 +716,18 @@ def combine(
         Find weights for a given function.
 
     """
-    return sum(
+    return (
```
paddyroddy, line 719: `sum(<x>)` is equivalent to `np.sum(<x>, axis=0)` and is more accurately showing what is happening + fixes `mypy`

glass/user.py
ntessore: But those are different situations: a blank EXTNAME is not the same as not having an EXTNAME.

docs/conf.py
ntessore, line 1: Would also make sense to me to include only `glass/` in the type checking.

examples/2-advanced/cosmic_shear.ipynb
ntessore, line 1: Why are the notebooks being formatted in this PR?

.gitignore
ntessore, line 1: Not sure this is really a change that needs to happen in this PR.

glass/core/array.py
```diff
@@ -72,10 +101,19 @@ def trapz_product(f, *ff, axis=-1):  # type: ignore[no-untyped-def]
...
+        x,
+        axis=axis,
+    )
 
 
-def cumtrapz(f, x, dtype=None, out=None):  # type: ignore[no-untyped-def]
+def cumtrapz(
+    f: npt.NDArray[np.int_] | npt.NDArray[np.float64],
+    x: npt.NDArray[np.int_] | npt.NDArray[np.float64],
+    dtype: type | None = None,
```
ntessore, line 114: better to use [DTypeLike](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike)

glass/fields.py
```diff
@@ -35,36 +34,22 @@
...
+    import collections.abc
 
 
 def iternorm(
     k: int,
-    cov: collections.abc.Iterable[npt.NDArray[typing.Any]],
-    size: Size = None,
-) -> collections.abc.Generator[Iternorm, None, None]:
+    cov: collections.abc.Iterable[npt.NDArray[np.float64]],
+    size: tuple[int, ...] = (),
```
ntessore, line 44: this was correct before: `size` has type `int | tuple[int, ...] | None` and default should remain `None`, which is semantically different from `()`

glass/fields.py
```diff
@@ -110,28 +95,33 @@ def iternorm(
 
 
 def cls2cov(
-    cls: Cls, nl: int, nf: int, nc: int
-) -> collections.abc.Generator[npt.NDArray[typing.Any], None, None]:
+    cls: npt.NDArray[np.float64],
```
ntessore, line 98: this was correct before: `cls` are a sequence of arrays or sequence of float, not one array

glass/fields.py
```diff
@@ -110,28 +95,33 @@ def iternorm(
...
-                    raise ValueError(msg)
-                n = len(cl)
-                cov[:n, i] = cl
-                cov[n:, i] = 0
+            if i == 0 and np.any(np.less(cl, 0)):
+                msg = "negative values in cl"
+                raise ValueError(msg)
+            n = len(cl)
+            cov[:n, i] = cl
+            cov[n:, i] = 0
```
ntessore, line 114: `None` entries must be supported here for backward compatibility

glass/fields.py
```diff
@@ -140,16 +130,21 @@ def multalm(alm: Alms, bl: npt.NDArray[typing.Any], *, inplace: bool = False) ->
     return out
 
 
-def transform_cls(cls: Cls, tfm: ClTransform, pars: tuple[typing.Any, ...] = ()) -> Cls:
+def transform_cls(
+    cls: npt.NDArray[np.float64],
```
ntessore, line 134: same as above, these types are incorrect now

glass/fields.py
```diff
@@ -140,16 +130,21 @@ def multalm(alm: Alms, bl: npt.NDArray[typing.Any], *, inplace: bool = False) ->
...
+    tfm: str | typing.Callable[[npt.NDArray[np.float64]], npt.NDArray[np.float64]],
+    pars: tuple[typing.Any, ...] = (),
+) -> list[list[float] | npt.NDArray[np.float64]]:
     """Transform Cls to Gaussian Cls."""
     gls = []
     for cl in cls:
-        if cl is not None and len(cl) > 0:  # type: ignore[redundant-expr]
+        if len(cl) > 0:
```
ntessore, line 141: `None` should be accepted here for backward compatibility

glass/fields.py
```diff
@@ -158,12 +153,12 @@ def transform_cls(cls: Cls, tfm: ClTransform, pars: tuple[typing.Any, ...] = ())
 
 
 def discretized_cls(
-    cls: Cls,
+    cls: list[list[float]] | npt.NDArray[np.float64],
```
ntessore, line 156: This should allow arbitrary sequences, not just lists

glass/fields.py
```diff
@@ -200,20 +195,20 @@ def discretized_cls(
 
 
 def lognormal_gls(
-    cls: Cls,
+    cls: npt.NDArray[np.float64],
```
ntessore, line 198: see above

glass/user.py
paddyroddy: But the function (as I understand it) will add "", which if it already has one isn't going to do anything. I can change this to

```py
if not self.ext or self.ext not in self.fits:
```

it just felt redundant

glass/fields.py
```diff
@@ -200,20 +195,20 @@ def discretized_cls(
...
     shift: float = 1.0,
-) -> Cls:
+) -> list[list[float] | npt.NDArray[np.float64]]:
     """Compute Gaussian Cls for a lognormal random field."""
     return transform_cls(cls, "lognormal", (shift,))
 
 
 def generate_gaussian(
-    gls: Cls,
+    gls: npt.NDArray[np.float64],
```
ntessore, line 206: see above, I don't understand how these pass currently?

glass/fields.py
```diff
@@ -200,20 +195,20 @@ def discretized_cls(
...
     nside: int,
     *,
     ncorr: int | None = None,
     rng: np.random.Generator | None = None,
-) -> collections.abc.Generator[npt.NDArray[typing.Any], None, None]:
+) -> collections.abc.Generator[npt.NDArray[np.complex128]]:
```
ntessore, line 211: This yields the real-valued maps, not the complex-valued alms

glass/fields.py
```diff
@@ -250,7 +245,7 @@ def generate_gaussian(
         ncorr = ngrf - 1
 
     # number of modes
-    n = max((len(gl) for gl in gls if gl is not None), default=0)  # type: ignore[redundant-expr]
+    n = max((len(gl) for gl in gls), default=0)
```
ntessore, line 248: `None` entries should be supported for backward compatibility

glass/fields.py
```diff
@@ -293,13 +288,13 @@ def generate_gaussian(
 
 
 def generate_lognormal(
-    gls: Cls,
+    gls: npt.NDArray[np.float64],
```
ntessore, line 291: see above

glass/fields.py
```diff
@@ -293,13 +288,13 @@ def generate_gaussian(
...
     nside: int,
     shift: float = 1.0,
     *,
     ncorr: int | None = None,
     rng: np.random.Generator | None = None,
-) -> collections.abc.Generator[npt.NDArray[typing.Any], None, None]:
+) -> collections.abc.Generator[npt.NDArray[np.complex128]]:
```
ntessore, line 297: see above

glass/fields.py
```diff
@@ -321,7 +316,12 @@ def generate_lognormal(
         yield m
 
 
-def getcl(cls, i, j, lmax=None):  # type: ignore[no-untyped-def]
+def getcl(
+    cls: npt.NDArray[np.float64],
```
ntessore, line 320: see above

glass/fields.py
```diff
@@ -351,8 +351,12 @@ def getcl(cls, i, j, lmax=None):  # type: ignore[no-untyped-def]
     return cl
 
 
-def effective_cls(  # type: ignore[no-untyped-def]
-    cls, weights1, weights2=None, *, lmax=None
+def effective_cls(
+    cls: npt.NDArray[np.float64],
```
ntessore, line 355: see above

glass/fields.py
```diff
@@ -402,10 +406,11 @@ def effective_cls(  # type: ignore[no-untyped-def]
...
     # auto-spectra do not repeat identical computations
-    if weights2 is weights1:
-        pairs = combinations_with_replacement(np.ndindex(shape1[1:]), 2)
-    else:
-        pairs = product(np.ndindex(shape1[1:]), np.ndindex(shape2[1:]))  # type: ignore[assignment]
+    pairs = (
+        combinations_with_replacement(np.ndindex(shape1[1:]), 2)
+        if weights2 is weights1
+        else product(np.ndindex(shape1[1:]), np.ndindex(shape2[1:]))
+    )
```
ntessore, line 413: I find this way harder to read, but I understand you don't want to manually type `pairs` here

glass/galaxies.py
```diff
@@ -117,43 +117,44 @@ def redshifts_from_nz(
         rng = np.random.default_rng()
 
     # bring inputs' leading axes into common shape
-    dims, count, z, nz = broadcast_leading_axes((count, 0), (z, 1), (nz, 1))  # type: ignore[no-untyped-call]
+    dims, *rest = broadcast_leading_axes((count, 0), (z, 1), (nz, 1))
+    count_out, z_out, nz_out = rest
```
ntessore, line 121: This change should not be necessary if `broadcast_leading_axes()` is typed correctly

glass/lensing.py
```diff
@@ -364,9 +372,9 @@ def wlens(self) -> float:
 
 
 def multi_plane_matrix(
-    shells: collections.abc.Sequence[RadialWindow],
+    shells: list[RadialWindow],
```
ntessore, line 375: this works with any sequence though

glass/lensing.py
```diff
@@ -377,10 +385,10 @@ def multi_plane_matrix(
 
 
 def multi_plane_weights(
-    weights: npt.ArrayLike,
-    shells: collections.abc.Sequence[RadialWindow],
+    weights: npt.NDArray[np.float64],
+    shells: list[RadialWindow],
```
ntessore, line 389: see above, no need to restrict it to a list (and it's not always a list in practice)

glass/observations.py
```diff
@@ -78,16 +77,16 @@ def vmap_galactic_ecliptic(
...
+    return hp.Rotator(coord="CE").rotate_map_pixel(m)
 
 
 def gaussian_nz(
-    z: npt.NDArray[typing.Any],
-    mean: npt.ArrayLike,
-    sigma: npt.ArrayLike,
+    z: npt.NDArray[np.float64],
+    mean: npt.NDArray[np.float64],
+    sigma: npt.NDArray[np.float64],
```
ntessore, line 86: these should accept `float` since they are in practice often called with single float arguments

ntessore (commented):
First and foremost: monumental work @paddyroddy!

I only managed to make it down to `observations.py` before I have to jump on the next round of calls, but leaving a first set comments here in the meantime. Main issue I've seen is that a) some functions changed behaviour, and b) the type of `cls`/`gls` arguments is incorrect: these are ragged sequences of arrays, for example `[np.array(...), np.array(...), []]` for the `cls` of two shells with no cross-correlation.

Anyway, **excellent** work and titanic effort! :)

glass/user.py
paddyroddy: Have changed to `if not self.ext or self.ext not in self.fits:`

docs/conf.py
paddyroddy: As discussed, it is only possible to exclude modules. Excluding folders doesn't seem supported https://github.com/python/mypy/issues/17977

.gitignore
paddyroddy: Agreed, but can we leave it? I had to run them to see the outputs for typing

examples/2-advanced/cosmic_shear.ipynb
paddyroddy: It was related to #362. I briefly commented out those rules because with all the typing, it really made sense to force stuff onto their own line. I can revert the notebooks, but that was why.

glass/fields.py
paddyroddy: `mypy` actually told me this one before I got around to this comment!

glass/fields.py
paddyroddy: If it helps, they are faster too https://switowski.com/blog/for-loop-vs-list-comprehension/#why-is-list-comprehension-faster-than-a-for-loop

glass/galaxies.py
paddyroddy: Could we leave this now for now and fix properly in #386?

glass/fields.py
paddyroddy: So to clarify the expected type is

```python
collections.abc.Sequence[npt.NDArray[np.float64] | collections.abc.Sequence[float]]
```

not

```python
npt.NDArray[np.float64] | collections.abc.Sequence[collections.abc.Sequence[float]]
```

?

glass/fields.py
paddyroddy: As discussed, I'm deprecating the `None` option. But have restored the type as `int | tuple[int, ...]`

glass/fields.py
paddyroddy: As discussed, I'm deprecating the `None` option. But have restored the type as `int | tuple[int, ...]`

glass/fields.py
paddyroddy: As discussed, I'm deprecating the `None` option. But have restored the type as `int | tuple[int, ...]`

paddyroddy:
I've addressed your comments @ntessore

pyproject.toml
```diff
@@ -131,10 +133,6 @@ src = [
     "glass",
 ]
 lint.ignore = [
-    "ANN001", # TODO: missing-type-function-argument
-    "ANN002", # TODO: missing-type-args
-    "ANN003", # TODO: missing-type-kwargs
-    "ANN201", # TODO: missing-return-type-undocumented-public-function
```
Saransh-cpp, line 137: 🎉

pyproject.toml
Saransh-cpp: I will suggest adding individual # type: ignore[no-any-return]` comments to ensure that our functions don't violate this.

Saransh-cpp (commented):
Amazing work 🤯

.git-blame-ignore-revs
paddyroddy: Okay, have removed.

pyproject.toml
paddyroddy: Yeah makes sense, have done this.

examples/2-advanced/cosmic_shear.ipynb
paddyroddy: These have been addressed in #393 instead

glass/shells.py
```diff
@@ -54,24 +53,31 @@
 from glass.core.array import ndinterp
 
 if typing.TYPE_CHECKING:
-    from cosmology import Cosmology
+    import collections.abc
 
-# types
-ArrayLike1D = typing.Union[collections.abc.Sequence[float], npt.NDArray[typing.Any]]
-WeightFunc = typing.Callable[[ArrayLike1D], npt.NDArray[typing.Any]]
```
ntessore, line 61: I'm not sure eliminating this type alias makes things more readable -- to me, it makes it harder to understand the signatures below.

glass/shells.py
```diff
@@ -470,36 +500,53 @@ def partition_lstsq(
...
+            * np.reshape(
+                np.trapz(  # type: ignore[attr-defined]
+                    fz,
+                    z,
+                ),
+                (*dims, 1),
+            ),
+        ],
+        axis=-1,
+    )
```
ntessore, line 531: What happened here?

glass/shells.py
```diff
@@ -524,25 +571,53 @@ def partition_nnls(
...
+                right=0.0,
+            )
+            for za, wa, _ in shells
+        ],
+    )
+    a /= np.trapz(  # type: ignore[attr-defined]
+        a,
+        zp,
+        axis=-1,
+    )[..., None]
```
ntessore, line 590: Is that ruff's doing? I really dislike this, it wasn't exactly a long line before...

glass/shells.py
```diff
@@ -626,14 +718,21 @@ def combine(
...
+                / np.trapz(  # type: ignore[attr-defined]
+                    shell.wa,
+                    shell.za,
+                ),
+                left=0.0,
+                right=0.0,
+            )
+            for shell, weight in zip(shells, weights)
+        ],
+        axis=0,
```
ntessore, line 737: No need to use `np.sum()` and hold all of the intermediate arrays in memory at the same time.

glass/user.py
ntessore: But what's the point of the conditional import here?

glass/user.py
```diff
@@ -55,23 +74,32 @@ class _FitsWriter:
...
 
-    def _append(self, data, names=None) -> None:  # type: ignore[no-untyped-def]
+    def _append(
+        self,
+        data: npt.NDArray[np.float64] | list[npt.NDArray[np.float64]],
+        names: list[str] | None = None,
+    ) -> None:
         """Write the FITS file."""
-        if self.ext is None or self.ext not in self.fits:
+        if not self.ext or self.ext not in self.fits:
```
ntessore, line 88: These are not the same, `ext` can be ""

glass/user.py
```diff
@@ -55,23 +74,32 @@ class _FitsWriter:
...
             self.fits.write_table(data, names=names, extname=self.ext)
-            if self.ext is None:
+            if not self.ext:
```
ntessore, line 90: Same here

glass/user.py
```diff
@@ -89,7 +117,11 @@ def write(self, data=None, /, **columns) -> None:  # type: ignore[no-untyped-def
 
 
 @contextmanager
-def write_catalog(filename, *, ext=None):  # type: ignore[no-untyped-def]
+def write_catalog(
+    filename: pathlib.Path,
+    *,
+    ext: str = "",
```
ntessore, line 123: Not the same as `ext=None`.

docs/conf.py
ntessore: Would it not suffice to add `packages = "heracles"` to `[tool.mypy]`?

paddyroddy: Just tried that and it didn't work... 🤔

glass/shells.py
paddyroddy: I was trying to make it clear that the `# type: ignore[attr-defined]` is specific to `np.trapz`. This is "fixed" in #391

glass/shells.py
paddyroddy: https://github.com/glass-dev/glass/pull/368#discussion_r1831118703

glass/user.py
paddyroddy: I'm aware by what would `ext` ever be `None`? My understanding is "a.csv" + "" = "a.csv"

glass/user.py
paddyroddy: https://github.com/glass-dev/glass/pull/368#discussion_r1831121217

glass/user.py
paddyroddy: https://github.com/glass-dev/glass/pull/368#discussion_r1831121217

glass/user.py
ntessore: Oh, right. This `ext=` is the FITS extension name. Extensions don't need to have a name, so that's why `None` (no name, use extension number) and `""` (empty string) are different

paddyroddy: Okay, makes sense. I couldn't find much documentation on FITS, will change

glass/shapes.py
```diff
@@ -87,12 +91,20 @@ def triaxial_axis_ratio(zeta, xi, size=None, *, rng=None):  # type: ignore[no-un
     C = 1 + z2m1 * cos2_phi  # noqa: N806
 
     # eq. (12)
-    return np.sqrt(
+    return np.sqrt(  # type: ignore[no-any-return]
         (A + C - np.sqrt((A - C) ** 2 + B2)) / (A + C + np.sqrt((A - C) ** 2 + B2)),
     )
```
ntessore, line 96: 👀 Is that a stray comma at the end of the line that causes mypy to complain?

paddyroddy: No. The `no-any-return` I believe is because the result of `np.sqrt` is meant to be `npt.NDArray[np.float64]` but the actual result isn't guaranteed to be one. So we could use `.astype`. This is why I haven't fixed that one.

glass/shapes.py
```diff
@@ -24,13 +24,17 @@
...
-
 import numpy as np
 import numpy.typing as npt
 
 
-def triaxial_axis_ratio(zeta, xi, size=None, *, rng=None):  # type: ignore[no-untyped-def]
+def triaxial_axis_ratio(
+    zeta: float | npt.NDArray[np.float64],
+    xi: float | npt.NDArray[np.float64],
+    size: int | tuple[int, ...] = (),
```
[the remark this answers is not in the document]
paddyroddy: Okay fair enough

glass/shells.py
paddyroddy: Okay, your call. I'm not a fan of the custom types.

paddyroddy: Although things like `WeightFunc` currently don't all have the same signature. So will require some fiddling.

glass/user.py
paddyroddy: If I directly import `fitsio` then users _have_ to have `fistio` which you're trying to avoid?

glass/shells.py
paddyroddy: Not sure I follow? The problem is:

With `sum()` (not numpy one) the assumed output type is `int` so signature is wrong

The `[]` encapsulating the expression is because `mypy` errors with `Generator expression must be parenthesized`

paddyroddy:
@ntessore I've addressed through your commments

glass/shells.py
ntessore: Ah, got it, this is changed for the implicit type

ntessore (approved):
Fantastic! Well done!

Saransh-cpp (approved):
Awesome! Thanks for putting so much efforts into this, @paddyroddy!

paddyroddy:
Thanks both for the review. Will merge this and update the downstream PRs.

Issue #272, opened 2024-09-25
Add/review tests for `glass.points`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Add/review tests for `glass.shapes`

PR #372, opened 2024-10-16
gh-272: add tests for glass.points

Saransh-cpp:
Add tests and bump coverage to ~97 for `glass.points`.

Closes: #272

### DIFF
--- a/tests/test_points.py
+++ b/tests/test_points.py
@@ -167,7 +167,7 @@ def test_positions_from_delta(rng: np.random.Generator) -> None:  # noqa: PLR091
     # case: negative delta
 
     lon, lat, cnt = catpos(
-        positions_from_delta(ngal, np.linspace(-1, -1, 196608), None, vis)
+        positions_from_delta(ngal, np.linspace(-1, -1, npix), None, vis)
     )
 
     assert isinstance(cnt, int)
@@ -177,7 +177,7 @@ def test_positions_from_delta(rng: np.random.Generator) -> None:  # noqa: PLR091
     # case: large delta
 
     lon, lat, cnt = catpos(
-        positions_from_delta(ngal, rng.normal(100, 1, size=(196608,)), bias, vis)
+        positions_from_delta(ngal, rng.normal(100, 1, size=(npix,)), bias, vis)
     )
 
     assert isinstance(cnt, int)
--- a/tests/test_points.py
+++ b/tests/test_points.py
@@ -4,8 +4,21 @@ import typing
 
 import numpy as np
 import numpy.typing as npt
+import pytest
 
-from glass.points import position_weights, positions_from_delta, uniform_positions
+from glass.points import (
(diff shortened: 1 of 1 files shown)


Saransh-cpp:
I can't seem to hit the following lines -

https://github.com/glass-dev/glass/blob/c020f18422b9f2960a1fd04d3b395749d2135ee6/glass/points.py#L227-L230

@ntessore do you have a specific case in mind that'll hit these lines. Every combination of arguments I have tried results in `q[-1] == count`, `size` is always 0, and putting these values in, it does not matter what `batch` is, the condition is never met.

ntessore:
> @ntessore do you have a specific case in mind that'll hit these lines. Every combination of arguments I have tried results in `q[-1] == count`, `size` is always 0, and putting these values in, it does not matter what `batch` is, the condition is never met.

Ah, good catch. The reason is that the batching processes 1000 pixels at a time, but we are using HEALPix maps with only 12 pixels as inputs.

Solution: use a higher number of pixels (`npix`) and replace all uses of `12` with `npix`, and add another case where some of the visibility is zero (hence it needs many batches of empty pixels until it finds enough points).

```py
def test_positions_from_delta():  # type: ignore[no-untyped-def]

    # create maps that saturate the batching in the function
    nside = 128
    npix = 12 * nside**2

    # case: single-dimensional input

    ngal = 1e-3
    delta = np.zeros(npix)
    bias = 0.8
    vis = np.ones(npix)

    ...

    # case: only the southern hemisphere is visible

    vis[:vis.size//2] = 0.0

    lon, lat, cnt = catpos(positions_from_delta(ngal, delta, bias, vis))  # type: ignore[no-untyped-call]

    assert cnt.shape == (3, 2)
    assert lon.shape == (cnt.sum(),)
    assert lat.shape == (cnt.sum(),)
```

Saransh-cpp:
I see, thank for the detailed suggestion, @ntessore! I will apply this to the tests.

paddyroddy (changes requested):
Can we change the upstream to #368 and fix the typing here?

tests/test_points.py
```diff
@@ -34,19 +47,142 @@ def catpos(
...
+        positions_from_delta(ngal, delta, bias, vis, remove_monopole=True)
+    )
+
+    assert isinstance(cnt, int)
+    assert lon.shape == lat.shape == (cnt,)
+
+    # case: negative delta
+
+    lon, lat, cnt = catpos(
+        positions_from_delta(ngal, np.linspace(-1, -1, 196608), None, vis)
```
ntessore, line 170: why not `npix` here and below?

Saransh-cpp: Thanks, updated!

ntessore (approved):
Nice work @Saransh-cpp!

paddyroddy (approved):
LGTM

PR #373, opened 2024-10-16
Fix `Returns` in docstrings (I think entered in #357)

paddyroddy:
I think due to development velocity, I accidentally brought these back in when addressing merge conflicts

### DIFF
--- a/glass/observations.py
+++ b/glass/observations.py
@@ -50,6 +50,8 @@ def vmap_galactic_ecliptic(
     the galactic and ecliptic planes. The location of the stripes is set with
     optional parameters.
 
+    Returns a HEALPix :term:`visibility map`.
+
     Parameters
     ----------
     nside:
@@ -58,11 +60,6 @@ def vmap_galactic_ecliptic(
         The location of the galactic plane in the respective coordinate system.
     ecliptic:
         The location of the ecliptic plane in the respective coordinate system.
-
-    Returns
-    -------
-    vis:
-        A HEALPix :term:`visibility map`.
 
     Raises
     ------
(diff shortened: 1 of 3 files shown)


Saransh-cpp (approved):
Thanks!

paddyroddy:
@ntessore can you look at this?

ntessore:
Oh, I just had a thought! Can we hold this for a bit?

paddyroddy:
> Oh, I just had a thought! Can we hold this for a bit?

I think we should get this in because I _think_ that we have a mix of the new and old `Returns` bits. I think these changes sneaked back in due to things like #357 and many merge conflicts. Could we get this in and then raise an issue to improve these? Related to #188

ntessore:
Ok, there is an option `napoleon_use_rtype` which can be set to `False` to move the return type information into the `:return:` statement. But this happens independently of the `:rtype:` being added by `autodoc_typehints`, so it doesn't help us.

In addition, just now I experimented a bit with https://github.com/tox-dev/sphinx-autodoc-typehints and this could provide the kind of annotation we would like to see:

```py
def vmap_galactic_ecliptic(
    nside: int,
    galactic: tuple[float, float] = (30, 90),
    ecliptic: tuple[float, float] = (20, 80),
) -> npt.NDArray[typing.Any]:
    """
    ...

    Returns
    -------
    A HEALPix :term:`visibility map`.

    """
```

Which produces:

image

But that's indeed for another issue.

paddyroddy:
> Ok, there is an option `napoleon_use_rtype` which can be set to `False` to move the return type information into the `:return:` statement. But this happens independently of the `:rtype:` being added by `autodoc_typehints`, so it doesn't help us.

Yeah, I had played with `napoleon_use_rtype` and hadn't much luck

> In addition, just now I experimented a bit with [tox-dev/sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) and this could provide the kind of annotation we would like to see:

Ah, nice! I will raise an issue for this

Issue #269, opened 2024-09-25
Add/review tests for `glass.fields`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Part of #161

PR #374, opened 2024-10-16
gh-269: add tests for glass.fields

Saransh-cpp:
Add tests and bump coverage to ~100 for `glass.fields`.

Closes: #269

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,6 +1,198 @@
 import numpy as np
+import pytest
 
-from glass.fields import getcl
+from glass.fields import cls2cov, getcl, iternorm, multalm
+
+
+def test_iternorm() -> None:
+    # check output shapes and types
+
+    k = 2
+
+    generator = iternorm(k, np.array([1.0, 0.5, 0.5, 0.5, 0.2, 0.1, 0.5, 0.1, 0.2]))
+    result = next(generator)
+
+    j, a, s = result
+
+    assert isinstance(j, int)
+    assert a.shape == (k,)
+    assert isinstance(s, float)  # type: ignore[unreachable]
+    assert s.shape == ()  # type: ignore[unreachable]
+
+    # specify size
+
+    size = 3
+
+    generator = iternorm(
+        k,
+        np.array(
+            [
+                [1.0, 0.5, 0.5],
+                [0.5, 0.2, 0.1],
+                [0.5, 0.1, 0.2],
+            ]
+        ),
+        size,
+    )
+    result = next(generator)
+
+    j, a, s = result
+
+    assert isinstance(j, int)
+    assert a.shape == (size, k)
+    assert s.shape == (size,)
+
+    # test shape mismatch error
+
+    with pytest.raises(TypeError, match="covariance row 0: shape"):
+        list(iternorm(k, np.array([[1.0, 0.5], [0.5, 0.2]])))
+
+    # test positive definite error
+
+    with pytest.raises(ValueError, match="covariance matrix is not positive definite"):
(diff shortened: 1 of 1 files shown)


Saransh-cpp:
Should be ready now! I'll create a new issue for reviewing _scientific_ tests for `glass.fields`.

Saransh-cpp:
Almost at 80% with this PR

paddyroddy:
> Almost at 80% with this PR

Wow, that went up a lot!

tests/test_fields.py
```diff
@@ -1,6 +1,392 @@
...
+    assert gaussian_fields[0].shape == (hp.nside2npix(nside),)
+
+    # requires resetting the RNG for reproducibility
+    rng = np.random.default_rng(seed=42)
+    gaussian_fields = list(generate_gaussian(gls, nside, rng=rng))
+
+    assert gaussian_fields[0].shape == (hp.nside2npix(nside),)
+
+    # requires resetting the RNG for reproducibility
+    rng = np.random.default_rng(seed=42)
```
paddyroddy, line 355: Why have it twice within this function?

paddyroddy: Should we set it once at the top of the file maybe?

tests/test_fields.py
```diff
@@ -1,6 +1,392 @@
...
+        assert m.shape == (192,)
+        assert np.all(m >= -1)
+
+    # check shift
+
+    # requires resetting the RNG to obtain exact the same result multiplied by 2 (shift)
+    rng = np.random.default_rng(seed=42)
+    lognormal_fields = list(generate_lognormal(gls, nside, shift=1.0, rng=rng))
+
+    rng = np.random.default_rng(seed=42)
```
paddyroddy, line 385: Same here

tests/test_fields.py
Saransh-cpp: Need it twice to ensure that `rng.any_distribution()` is same both the times, and it needs to be redefined (instead of using `rng` from the arguments) so that it starts sampling from scratch.

Setting it at the top of the file gives me -
```py
        # requires resetting the RNG for reproducibility
>       gaussian_fields = list(generate_gaussian(gls, nside, rng=rng))
E       UnboundLocalError: cannot access local variable 'rng' where it is not associated with a value
```

paddyroddy: Okay! Might a little comment saying why it's being redefined?

ntessore: I think the right approach here is to use our central `rng` fixture and clone it:

```py
from copy import deepcopy

def test_xyz(rng):
    rng1 = deepcopy(rng)
    rng2 = deepcopy(rng)
    ...
```

ntessore: This probably needs a fairly recent numpy

ntessore: But what is this actually testing? If it's that the generation is reproducible, it's on the RNG, not on us. In which case I would rather test that the rng was actually used:

```py
initial_rng_state = rng.bit_generator.state

...

assert rng.bit_generator.state != initial_rng_state
```

Saransh-cpp: My initial thought behind adding this test was to check if `generate_gaussian` with and without `ncorr=1` outputs the same result (because that is a pattern that I observed while running the function manually). Is there a better test that can be added to check if passing the `ncorr` argument works, or should I leave it for #414?

Saransh-cpp:
pre-commit.ci autofix

paddyroddy (approved):
Looks good! @ntessore it would be good to get this in

paddyroddy:
Going to merge to get the testing in to assess coverage

Issue #379, opened 2024-10-24
Fix warnings which arise from running the notebooks


paddyroddy:
### Describe the Bug

Discovered in the unintentional run in #367

### To Reproduce

import glass
...

### Version In Use

N/A

### Additional Context

- Python version:
- Operating system:

paddyroddy:
Fixed in #380

PR #380, opened 2024-10-24
gh-379: fix warnings when running the notebooks

paddyroddy:
Spotted in #379. This fixes the warning issues that users may encounter and wouldn't look great on the documentation examples.


Saransh-cpp (approved):
Nice catch!

ntessore (approved):

Issue #375, opened 2024-10-18
Improve docstring return output with typing in `sphinx`


paddyroddy:
In addition, just now I experimented a bit with https://github.com/tox-dev/sphinx-autodoc-typehints and this could provide the kind of annotation we would like to see:

```py
def vmap_galactic_ecliptic(
    nside: int,
    galactic: tuple[float, float] = (30, 90),
    ecliptic: tuple[float, float] = (20, 80),
) -> npt.NDArray[typing.Any]:
    """
    ...

    Returns
    -------
    A HEALPix :term:`visibility map`.

    """
```

Which produces:

<img width="744" alt="image" src="https://github.com/user-attachments/assets/7af19061-758f-40d2-9ac4-8dd1d04bb28b">

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/373#issuecomment-2419856894_

ntessore:
I'm happy to experiment with this

paddyroddy:
@ntessore I'm looking at this as part of #188, have got something like:

![Image](https://github.com/user-attachments/assets/710f063e-e684-41e6-94de-2d1d9c9c5368)

```py
# -- autodoc -----------------------------------------------------------------

autodoc_typehints = "description"


# -- napoleon ----------------------------------------------------------------

napoleon_google_docstring = False
napoleon_use_rtype = False


# -- sphinx-autodoc-typehints ----------------------------------------------------------------

always_use_bars_union = True
typehints_use_rtype = False
typehints_defaults = "comma"
```

with these settings, but I can't work out how to get rid of the redundant `Returns`... Any idea?

ntessore:
To be honest, I am starting to think perhaps this is a feature, not a bug. For simple return types, this isn't the worst output:

    Returns:
      The value of the thing.

    Return type:
      int

Whereas for complicated situations, having both sections allows for much more nuance:

    Returns:
      - a (int) -- The value of a. Only returned if return_a is true.
      - b (float) -- The value of b. Only returned if return_b is true.
      - c (bool) -- The value of c. Only returned if return_c is true.

    Return type:
      <some complicated mess that is impossible to understand, ideally suppressed>

ntessore:
Perhaps a good question to answer first is whether the documentation of [`from_convergence()`](https://glass.readthedocs.io/latest/reference/lensing.html#glass.from_convergence) can be made human-readable using a number of `@typing.overload`?

paddyroddy:
[...]

Issue #369, opened 2024-10-15
Don't specify default values in docstrings


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Similar reason to the types. It's another thing to keep in sync, and the defaults can be seen in the documentation anyway

paddyroddy:
Closing as I will fix it in #188

Issue #350, opened 2024-10-10
Fix `intersphinx` for the `cosmology` package


paddyroddy:
### What is the Topic of Your Question

Documentation

### Add Your Question Below

Following #347 we now have static types rendering in the documentation. However, as can be seen in the screenshot, `npt.NDArray` isn't highlighted. Anyone know if it's possible?

![Image](https://github.com/user-attachments/assets/dc20a2ec-a51c-4267-b1dc-8fad14c527b9)

Saransh-cpp:
`intersphinx` does this, but we will need to `import numpy.typing` in every file without the `TYPE_CHECKING` condition (https://github.com/sphinx-doc/sphinx/issues/10746), or figure out a way to do that on the fly by writing a custom extension (see https://github.com/pybamm-team/PyBaMM/tree/develop/docs/sphinxext for instance).

Saransh-cpp:
Given the files have a "non-conditional" import of `numpy.typing`, some of the docs render (ugly imo) cross-references to `numpy` types -

<img width="754" alt="image" src="https://github.com/user-attachments/assets/387d77e9-4486-4c7f-8b77-f7ca539e0e88">

Should we should stick to one thing? Either import `numpy.typing` without a condition or with. If we want to add the condition, we can move the types declared in this file to a utility file (`_typeutils.py` or something).

_Originally posted by @Saransh-cpp in https://github.com/glass-dev/glass/pull/347#discussion_r1795879501_

paddyroddy:
So looking into this, it has nothing to do with `TYPE_CHECKING`. It's look alright as of #381. However, I'm trying to work out how to get it working for the [cosmology package](https://cosmology.readthedocs.io/en/stable), which doesn't seem to work out the box.

paddyroddy:
So looks like everything is working except https://cosmology.readthedocs.io/en/stable. I tried both

```python
intersphinx_mapping = {
    "cosmology": ("https://cosmology.readthedocs.io/en/stable", None),
    "numpy": ("https://numpy.org/doc/stable", None),
    "python": ("https://docs.python.org/3", None),
}
```

and

```python
intersphinx_mapping = {
    "cosmology": (
        "https://cosmology.readthedocs.io/en/stable",
        "https://cosmology.readthedocs.io/en/stable/objects.inv",
    ),
    "numpy": ("https://numpy.org/doc/stable", None),
    "python": ("https://docs.python.org/3", None),
}
```

any ideas?

[...]

PR #381, opened 2024-10-25
gh-188: add docstrings to all functions and tidy docs

### LINKED ISSUE #188: Add docstrings to all functions

paddyroddy:
Closes #188, #369, #375. Have set #358 as upstream as it eases the process.

Adds docstrings to all functions, namely a description for all parameters in the `glass` package. I don't know the code base well, so would be good for these to be checked that they make sense.

I'm also not really familiar with RST so worth checking if any of that can be improved.


glass/shells.py
```diff
@@ -413,6 +481,7 @@ def partition(
     approximate solution of
 
     .. math::
+
         \begin{pmatrix}
```
paddyroddy, line 485: This was previously breaking the maths

docs/conf.py
```diff
@@ -62,8 +64,8 @@
 # This config value contains the locations and names of other projects that
 # should be linked to in this documentation.
 intersphinx_mapping = {
+    "numpy": ("https://numpy.org/doc/stable", None),
     "python": ("https://docs.python.org/3", None),
-    "numpy": ("https://numpy.org/doc/stable/", None),
 }
```
paddyroddy, line 69: See #350

glass/lensing.py
```diff
@@ -61,27 +61,29 @@ def from_convergence(  # noqa: PLR0913
...
+    discretized
         Correct the pixel window function in output maps.
 
+    Returns
+    -------
+        The maps of:
+
+        * deflection potential if ``potential`` is true.
+        * potential (complex) if ``deflection`` is true.
+        * shear (complex) if ``shear`` is true.
```
Saransh-cpp, line 85: Maybe requires better indentation?

image

```suggestion
        The maps of:

            * deflection potential if ``potential`` is true.
            * potential (complex) if ``deflection`` is true.
            * shear (complex) if ``shear`` is true.
```

glass/lensing.py
```diff
@@ -61,27 +61,29 @@ def from_convergence(  # noqa: PLR0913
...
+
     Notes
```
Saransh-cpp, line 87: Nitpicking, but "Notes" and "References" in the rendered docs do not get a `:` by default -

image

Should we add it in the docstring, or just leave it as it is?

Saransh-cpp (commented):
Amazing, thanks @paddyroddy! Quite a lot of efforts in this 🙂

Saransh-cpp (approved):

glass/lensing.py
paddyroddy: Good idea

glass/lensing.py
paddyroddy: I had spotted this. I'd be inclined to follow the NumpyDoc standard https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

paddyroddy:
@ntessore can you have a look?

glass/shells.py
```diff
@@ -538,6 +631,21 @@ def partition_nnls(
     Uses the ``nnls()`` algorithm from ``scipy.optimize`` and thus
     requires SciPy.
```
ntessore, line 633: This isn't actually true, which took a lot of effort :)

ntessore (commented):
Excellent, this must have been hard work! A few trivial things noted below

glass/shells.py
paddyroddy: Nice, have removed that sentence

glass/points.py
```diff
@@ -148,37 +178,38 @@ def positions_from_delta(  # noqa: PLR0912, PLR0913, PLR0915
...
-    lon:
-        Columns of longitudes for the sampled points.
-    lat:
-        Columns of latitudes for the sampled points.
-    count:
-        The number of sampled points  If multiple populations are sampled, an
-        array of counts in the shape of the extra dimensions is returned.
+        The longitudes, latitudes and the number of sampled points. If
+        multiple populations are sampled, an array of counts in the shape
+        of the extra dimensions is returned.
```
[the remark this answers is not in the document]
paddyroddy: I feared you might say this. Will see what I can do. Seems to be some sphinx/autodoc oddity.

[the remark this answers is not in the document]
ntessore: Not sure if this works for the yields section, but worth a try maybe:

```py
# conf.py
napoleon_custom_sections = [("Yields", "params_style")]
```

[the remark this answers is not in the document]
paddyroddy: I found a way to do it without fiddling with `napolean`. It seems `Returns` and `Yields` operate slightly differently.

glass/points.py
```diff
@@ -300,20 +331,15 @@ def uniform_positions(
...
     ------
-    lon:
-        Columns of longitudes for the sampled points.
-    lat:
-        Columns of latitudes for the sampled points.
-    count:
-        The number of sampled points. For array inputs, an array of
-        counts with the same shape is returned.
+        The longitudes, latitudes and the number of sampled points. For array
+        inputs, an array of counts with the same shape is returned.
```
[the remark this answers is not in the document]
paddyroddy: Fixed https://github.com/glass-dev/glass/pull/381#discussion_r1840492686

paddyroddy:
@ntessore this is ready again

ntessore:
The yields sections are rendering the text as types now.  The original formatting actually produces the desired output:

```
    Yields
    ------
    lon: ndarray[Any, dtype[float64]]
        Columns of longitudes for the sampled points.
    lat: ndarray[Any, dtype[float64]]
        Columns of latitudes for the sampled points.
    count: int | ndarray[Any, dtype[int64]]
        The number of sampled points  If multiple populations are sampled, an
        array of counts in the shape of the extra dimensions is returned.
```

You could leave out the types here if you are concerned about duplicating the information, but I think there's little harm in these isolated cases.

paddyroddy:
What I wanted to do is what we've done for `from_convergence`, but I've tried every variant, and it doesn't work. This is what I mean when I say `Yields` act differently.

image

ntessore:
Or even, using

```py
# conf.py
napoleon_custom_sections = [
    ("Returns", "params_style"),
    ("Yields", "params_style"),
]
```

having it apply the correct bold font to the variable names:

image

This also seems to change the order to Parameters - Yields - Returns - Raises - Return Type which keeps the most relevant information together.

paddyroddy:
> But even if you don't want to repeat the types, what's wrong with this?

I might be imagining it, but I thought we'd discussed that syntax previously. Which is why I wasn't using it.

It looks like it does work. I guess the main drawback is that it implies a name for the output, which is in general up to the user. It would also make sense to do this throughout for the codebase for all `Returns`/`Yields`. Is that what you would like?

ntessore:
> I might be imagining it, but I thought we'd discussed that syntax previously. Which is why I wasn't using it.

As I remember it, we discussed this when we tried to get the return type into the returns section. Since that wasn't working well, I thought we decided to keep the return type section separate, I think there's nothing in the way of having multiple yields and returns?

paddyroddy:
> As I remember it, we discussed this when we tried to get the return type into the returns section. Since that wasn't working well, I thought we decided to keep the return type section separate, I think there's nothing in the way of having multiple yields and returns?

Yeah it must have been then. I will go through and change all `Returns`/`Yields` now.

paddyroddy:
> Yeah it must have been then. I will go through and change all `Returns`/`Yields` now.

Done. It looks like it was just that one `Returns` and those two `Yields`.

glass/lensing.py
```diff
@@ -78,11 +78,12 @@ def from_convergence(  # noqa: PLR0913
...
-
-            * potential (complex) if ``deflection`` is true.
-            * deflection potential if ``potential`` is true.
-            * shear (complex) if ``shear`` is true.
+    alpha
+        Map of the deflection (complex). Only returned if ``deflection`` if true.
+    psi
+        Map of the deflection potential. Only returned if ``potential`` is true.
+    gamma
+        Map of the shear (complex). Only returned if ``shear`` is true.
```
ntessore, line 86: My apologies, it seems like I was confusing myself before. The order is `psi, alpha, gamma` like the order of the derivatives.

```suggestion
    psi
        Map of the lensing (or deflection) potential. Only returned if ``potential`` is true.
    alpha
        Map of the deflection (complex). Only returned if ``deflection`` if true.
    gamma
        Map of the shear (complex). Only returned if ``shear`` is true.
```

ntessore (approved):
Fantastic effort, particularly with the many changes in specs! There is a remaining issue with single-line yields but it's so minuscule that I want to tackle it separately. Towards a new release!

Saransh-cpp (approved):
Thanks, @paddyroddy! Looks good and just needs fixing the minor pre-commit.ci failure.

paddyroddy:
pre-commit.ci autofix

paddyroddy:
> Fantastic effort, particularly with the many changes in specs! There is a remaining issue with single-line yields but it's so minuscule that I want to tackle it separately. Towards a new release!

Great, could you make an issue for this?

paddyroddy:
> Great, could you make an issue for this?

https://github.com/glass-dev/glass/issues/418

Issue #385, opened 2024-10-25
Set random seeds in the notebook outputs


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Makes sense to have consistent output on subsequent runs of the notebooks. It's not something we anticipate having to run often, but it will need to be ran time to time.

PR #389, opened 2024-10-30
gh-385: set random seeds in all glass functions used in notebooks

paddyroddy:
Leverage the `rng` input to all functions in `glass` so that the notebook figures should always be the same when run again.

Also add https://github.com/kynan/nbstripout with `--keep-output` which will clear the metadata of the notebooks but make sure figures are still shown.


Saransh-cpp (commented):
Nice, maintainability 📈

Thanks, @paddyroddy

Saransh-cpp (approved):

paddyroddy:
@ntessore can you have a look?

examples/1-basic/shells.ipynb
ntessore, line 1: This one doesn't do any random sampling

ntessore (commented):
Great effort to make the outputs more stable. There's one notebook that doesn't do any sampling, but still gets a `rng`.

In a separate change, we probably want to add a matplotlibrc or similar to control at least the resolution of the notebook figures – you can see here that the images still depend on who generated them.

Issue #390, opened 2024-10-30
Make a release


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

We need a release before tackling #67. Creating this issue largely so I can refer to it elsewhere.

paddyroddy:
Blocked by #381, #400.

PR #391, opened 2024-10-30
gh-163: drop `Python 3.9` & `NumPy < 2.1.1` support

### LINKED ISSUE #163: NumPy 2 compatibility (+ drop `3.9` support)

paddyroddy:
Closes #163. This will be required for #67. Blocked by #390.

As per the [NumPy 2.1.1 release notes](https://github.com/numpy/numpy/releases/tag/v2.1.1) Python prior to `3.9` is not supported. This release is also the first version with full Array API support. So in order to support it full we need these changes.

There are some code changes due to `ruff` and the minimum Python version being `3.10`.


glass/observations.py
```diff
@@ -258,7 +251,7 @@ def equal_dens_zbins(
     cuml_nz /= cuml_nz[[-1]]
     zbinedges = np.interp(np.linspace(0, 1, nbins + 1), cuml_nz, z)
 
-    return list(zip(zbinedges, zbinedges[1:]))
+    return list(itertools.pairwise(zbinedges))
```
Saransh-cpp, line 254: TIL

Saransh-cpp (commented):
Awesome! Will make the Array API transition much smoother

glass/observations.py
paddyroddy: You can thank `ruff` https://docs.astral.sh/ruff/rules/zip-instead-of-pairwise

Saransh-cpp (approved):

noxfile.py
```diff
@@ -10,7 +10,12 @@
 nox.options.reuse_existing_virtualenvs = True
 nox.options.sessions = ["lint", "tests"]
 
-ALL_PYTHON = ["3.9", "3.10", "3.11", "3.12", "3.13"]
+ALL_PYTHON = [
+  "3.10",
+  "3.11",
+  "3.12",
+  "3.13",
+]
```
paddyroddy, line 18: In future we'll just get one line change rather than many

paddyroddy:
pre-commit.ci autoupdate

PR #392, opened 2024-10-30
gh-384: rename all internal functions to `trapezoid` over `trapz`

### LINKED ISSUE #384: Change all functions with `trapz` in the name to `trapezoid` in line with NumPy v2

As part of #163 we will have to change all uses of `np.trapz` to `np.trapezoid`, it would make sense to change the functions like `cumtrapz` to `cumtrapezoid` etc.

_No response_

_No response_

### Additional Context

_No response_

paddyroddy:
Closes #384. Blocked by #391.

Keeping in line with `NumPy v2` renaming `np.trap` to `np.trapezoid`. It was agreed to change the internal name of functions to avoid potential confusion.


Saransh-cpp (approved):
Ah, I forgot to approve, just commented by mistake

ntessore (commented):
Maybe better to carry on agreeing with the scipy name, i.e., now `cumulative_trapezoid`?

paddyroddy:
> Maybe better to carry on agreeing with the scipy name, i.e., now `cumulative_trapezoid`?

Sure. `cumtrapezoid` without the `_` does seem a bit ugly.

Saransh-cpp (changes requested):
This, @paddyroddy! Could you fix the git history?

paddyroddy:
> This, @paddyroddy! Could you fix the git history?

We're squash merging... I don't see the benefit

paddyroddy:
> This, @paddyroddy! Could you fix the git history?

For reference, this happened because the head was #391. I do think tidying this is a waste of dev time with squash merging enabled.

Saransh-cpp (approved):
Fair enough, everything should be fine as long as the commit message of the squashed commit does not reflect this history.

ntessore (approved):
Nice, thanks!

I just realised that I thought we were testing the cumulative trapezoid by chopping the array down to 1, 2, 3, ... elements, not scipy. Maybe a change for the future to get rid of that dependency.

PR #393, opened 2024-10-30
gh-362: review the `ruff-format` recommended rules

### LINKED ISSUE #362: Review "ruff-format recommended" ignore rules

paddyroddy:
Closes #362. When you first run `ruff-format` it only warns about `COM812` and `ISC001`. This PR just reduces the others which technically also in conflict to just these two. I have used these extensively and have had no issues.


Saransh-cpp (approved):
Nice! Much cleaner now, thanks!

PR #397, opened 2024-11-07
gh-124: use `cosmology.api` over the `cosmology` package

### LINKED ISSUE #124: use cosmology API instead of deprecated cosmology package
Switch to use the [cosmology API](https://github.com/cosmology-api/cosmology.api) instead of the old, deprecated [cosmology package](https://github.com/glass-dev/cosmology).

There is a work-in-progress [CAMB wrapper](https://github.com/cosmology-api/cosmology.compat.camb), which needs to be checked for feature-completeness against all uses of `cosmo` in GLASS. I don't think we use anything that's not covered by the cosmology API currently (growth functions, power spectra).

Since the cosmology API functions have no dimensionless equivalent to the cosmology package, there is probably some amount of testing needed in e.g. the lensing module.

cc @nstarman

paddyroddy:
Closes #124. Use the improved `cosmology.api` package throughout the code.


glass/galaxies.py
```diff
@@ -15,10 +15,6 @@
 .. autofunction:: galaxy_shear
 .. autofunction:: gaussian_phz
 
-Intrinsic alignments
---------------------
-.. autofunction:: kappa_ia_nla
```
paddyroddy, line 20: I have reverted https://github.com/glass-dev/glass/pull/21 as I believe it is broken/needs testing. Don't want to waste time porting.

tests/test_lensing.py
```diff
@@ -31,20 +31,21 @@ def shells() -> list[RadialWindow]:
...
-        def omega_m(self) -> float:
+        def Omega_m0(self) -> float:  # noqa: N802
             return 0.3
 
-        def ef(self, z: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
-            return (self.omega_m * (1 + z) ** 3 + 1 - self.omega_m) ** 0.5
+        def H_over_H0(self, z: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:  # noqa: N802
+            return (self.Omega_m0 * (1 + z) ** 3 + 1 - self.Omega_m0) ** 0.5
 
         def xm(
```
paddyroddy, line 43: There is no equivalently named function in `cosmology.api`. In practice, this would be 

```python
cosmology.api.Standard.Cosmology.transverse_comoving_distance(z) / (cosmology.api.CosmologyConstantsNamespace.c / cosmology.api.Standard.Cosmology.H0)
```

glass/lensing.py
```diff
@@ -327,15 +330,19 @@ def add_plane(
         w2, self.w3 = self.w3, wlens
 
         # extrapolation law
-        x2, self.x3 = self.x3, self.cosmo.xm(self.z3)
+        hubble_length = CosmologyConstantsNamespace.c / StandardCosmology.H0
+        x2 = self.x3
+        self.x3 = self.cosmo.transverse_comoving_distance(self.z3) / hubble_length
```
paddyroddy, line 335: I've split these so the line isn't really long

ntessore:
Ok, the Cosmology API side is in a place where this should be able to progress!

There is a `cosmology.compat.camb` package now which contains a wrapper for CAMB, much like the old `cosmology` package. It requires passing a `results` object, which can be obtained from `results = camb.get_background(pars)`.

We can also be much more precise now in our use of cosmology functions by specifying a protocol of only the interfaces we need. I suggest we create a new internal module `glass/cosmology.py` in which we define our requirements on a cosmology object:

```py
class Cosmology(
    cosmology.api.HasHoverH0,
    cosmology.api.HasOmegaM0,
    cosmology.api.HasHubbleDistance,
    ...,
    Protocol,
):
    """
    Cosmology protocol for GLASS.

    GLASS requires a cosmology object with the following methods
    and attributes, as defined by the Cosmology API:
    * :meth:`cosmology.api.H_over_H0`
    * ...

    """
```

Our functions can then be typed using our `Cosmology` protocol.

(The `kappa_nla` model requires the growth factor, which is still TBA in the Cosmology API. But I am adding that asap.)

paddyroddy:
Nice! I'll take a look over the next few days

paddyroddy:
Closing as it seems like I may as well start again

PR #416, opened 2024-11-14
gh-286: replace `stable` in docs with tag for `sdist`

### LINKED ISSUE #286: Update README to point to correct examples
### Which Section of the Documentation Needs Improving?

README

### What Can be Improved About This Section

The README still points to the [external examples](https://github.com/glass-dev/glass-examples/) which have been sunset.

### How to Improve This Section

The README should point to the new [in-place examples](https://glass.readthedocs.io/latest/examples.html). There's a question whether this should point to `latest` (good for the repository) or to `stable` (good for the PyPI page, although even better would be pointing to the correct version, which seems a lot of work for very little gain).

paddyroddy:
Closes #286. Upon building the `sdist` we can replace `stable` with the relevant tag.

Have also used `stable` as the default throughout. This wasn't possible when issue #286 was created, as the `stable` tag didn't exist.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Replace stable version with the release version
         run: |-
           # store the release tag
-          tag="${{ github.event.release.tag_name }}"
+          tag="${{ github.ref_name }}"
 
           # files to search for
           md_rst_files=$(find . -type f \( -name "*.md" -o -name "*.rst" \))
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,17 +26,15 @@ jobs:
           # store the release tag
           tag="${{ github.ref_name }}"
 
-          # files to search for
-          md_rst_files=$(find . -type f \( -name "*.md" -o -name "*.rst" \))
-
           # regex patterns
           badge_pattern="s#(readthedocs\.org/projects/glass/badge/\?version=)stable#\1${tag}#g"
           url_pattern="s#(glass\.readthedocs\.io.*?)/stable#\1/${tag}#g"
 
           # perform the replacements
-          for file in $md_rst_files; do
-            sed -E -i "$badge_pattern; $url_pattern" "$file"
-          done
+          sed --in-place --regexp-extended \
+            --expression "$badge_pattern" \
+            --expression "$url_pattern" \
+            ${{ github.workspace }}/README.md
 
       - name: Build SDist and wheel
         run: pipx run build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,23 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v4
 
+      - name: Replace stable version with the release version
+        run: |-
+          # store the release tag
+          tag="${{ github.event.release.tag_name }}"
+
+          # files to search for
+          md_rst_files=$(find . -type f \( -name "*.md" -o -name "*.rst" \))
+
+          # regex patterns
+          badge_pattern="s#(readthedocs\.org/projects/glass/badge/\?version=)stable#\1${tag}#g"
+          url_pattern="s#(glass\.readthedocs\.io.*?)/stable#\1/${tag}#g"
+
+          # perform the replacements
+          for file in $md_rst_files; do
+            sed -E -i "$badge_pattern; $url_pattern" "$file"
+          done
+
       - name: Build SDist and wheel
         run: pipx run build
 
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -3,7 +3,7 @@ title: "GLASS: Generator for Large Scale Structure"
 message: >-
   If you use GLASS in your research, please cite the associated paper.
 repository-code: "https://github.com/glass-dev/glass"
-url: "https://glass.readthedocs.io"
+url: "https://glass.readthedocs.io/stable"
 license: MIT
 preferred-citation:
   authors:
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -3,7 +3,7 @@ title: "GLASS: Generator for Large Scale Structure"
 message: >-
   If you use GLASS in your research, please cite the associated paper.
 repository-code: "https://github.com/glass-dev/glass"
-url: "https://glass.readthedocs.io/stable"
+url: "https://glass.readthedocs.io"
 license: MIT
 preferred-citation:
   authors:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,7 +78,7 @@ rendering documentation on its website. The configuration file (`conf.py`) for
 `sphinx` can be found
 [under the `docs` folder](https://github.com/glass-dev/glass/blob/main/docs/conf.py).
 The documentation is deployed on <https://readthedocs.io>
-[here](https://glass.readthedocs.io/en/latest).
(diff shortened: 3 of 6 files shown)


paddyroddy:
I've tested the `sed` expression locally, which works. I tried to use the [workflow dispatch to TestPyPI](https://github.com/glass-dev/glass/blob/84f0b697c03b971e6d4e292c31dfcaba470c2bf4/.github/workflows/release.yml#L5-L9) which doesn't work as the credentials haven't been set up @ntessore. Probably worth getting this in for the new release.

image

Saransh-cpp (approved):
Looks nice, thanks!

CITATION.cff
```diff
@@ -3,7 +3,7 @@ title: "GLASS: Generator for Large Scale Structure"
 message: >-
   If you use GLASS in your research, please cite the associated paper.
 repository-code: "https://github.com/glass-dev/glass"
-url: "https://glass.readthedocs.io"
+url: "https://glass.readthedocs.io/stable"
```
ntessore, line 6: This can remain at the top-level URL since it forwards to `/stable`.

CONTRIBUTING.md
```diff
@@ -78,7 +78,7 @@ rendering documentation on its website. The configuration file (`conf.py`) for
 `sphinx` can be found
 [under the `docs` folder](https://github.com/glass-dev/glass/blob/main/docs/conf.py).
 The documentation is deployed on <https://readthedocs.io>
-[here](https://glass.readthedocs.io/en/latest/).
+[here](https://glass.readthedocs.io/en/latest).
```
ntessore, line 81: Any reason for touching this line?

docs/manual/first-steps.rst
```diff
@@ -3,4 +3,4 @@ First steps
 
 The best way to get started with *GLASS* is to follow the examples__.
```
ntessore, line 4: This is actually not working right now, the correct version is

```suggestion
The best way to get started with *GLASS* is to follow the :doc:`/examples`.
```

and then the external link removed.

pyproject.toml
```diff
@@ -67,7 +67,7 @@ test = [
 
 [project.urls]
 Changelog = "https://glass.readthedocs.io/stable/manual/releases.html"
-Documentation = "https://glass.readthedocs.io/"
+Documentation = "https://glass.readthedocs.io/stable"
```
ntessore, line 70: Again I think this just needs to point to the base URL.

.github/workflows/release.yml
```diff
@@ -21,6 +21,23 @@ jobs:
...
+
+          # regex patterns
+          badge_pattern="s#(readthedocs\.org/projects/glass/badge/\?version=)stable#\1${tag}#g"
+          url_pattern="s#(glass\.readthedocs\.io.*?)/stable#\1/${tag}#g"
+
+          # perform the replacements
+          for file in $md_rst_files; do
+            sed -E -i "$badge_pattern; $url_pattern" "$file"
+          done
+
```
ntessore, line 40: We only need to patch README.md really, since nothing else appears in the sdist or wheels.

README.md
```diff
@@ -3,7 +3,7 @@
 <!-- Essentials -->
 
 [![PyPI](https://img.shields.io/pypi/v/glass)](https://pypi.org/project/glass)
-[![Documentation](https://readthedocs.org/projects/glass/badge/?version=latest)](https://glass.readthedocs.io/latest)
+[![Documentation](https://readthedocs.org/projects/glass/badge/?version=stable)](https://glass.readthedocs.io/stable)
```
ntessore, line 6: In the README, which everyone will generally see on current main, this should point to `latest` not `stable` IMO.

ntessore (commented):
Nicely done, particularly catching both the image and link URLs! But I think the patch only needs to apply to README.md, as everything else won't matter for the package.

CITATION.cff
paddyroddy: I still think we should have it anyway, it's good practice. Is there a good reason to not?

CONTRIBUTING.md
paddyroddy: I dislike the trailing slash, and was semi-relevant as a RTD link

README.md
paddyroddy: Hmm... I guess we want the badge to refer to `stable`. But are you sure we want the docs link to be `latest`? People will in general be using the most recent stable version, so the docs could become out of sync.

I thought it would be off to have `latest` for the badge, and `stable` for the link - which is why I had them both as `stable`.

pyproject.toml
paddyroddy: https://github.com/glass-dev/glass/pull/416#discussion_r1843681121

CONTRIBUTING.md
ntessore: But the URL is `https://glass.readthedocs.io/en/latest/`, the version without slash is just a forward.

docs/manual/first-steps.rst
paddyroddy: Good spot, have tested it

README.md
ntessore: Yeah I guess the question here is who will be looking at the repo. Probably even people who are on stable, so maybe you are right that stable is better.

CITATION.cff
paddyroddy: As a safety net. Surely it follows a similar argument to https://github.com/glass-dev/glass/pull/416#discussion_r1843685913?

README.md
paddyroddy: I left `CONTRIBUTING.md` as `/latest` because it will be developers looking at that

.github/workflows/release.yml
paddyroddy: Okay, I'll change it to that

CITATION.cff
ntessore: Thinking along the same lines, I think "the URL" is the root one which is always safe to link to a valid place (e.g., when /en was removed). Below we need to point specifically to "the docs for stable/v2024.2" so I don't think the same argument applies.

paddyroddy: And for `pyproject.toml`

pyproject.toml
paddyroddy: What about this one (cc: https://github.com/glass-dev/glass/pull/416#discussion_r1843720045)? The `Changelog` has to have `stable` in it by RTD design

README.md
```diff
@@ -80,7 +80,7 @@ e.g. a design decision or API change, you can use our [Discussions] page.
 
 We also have a public [Slack workspace] for discussions about the project.
 
-[documentation]: https://glass.readthedocs.io/
-[examples]: https://glass.readthedocs.io/projects/examples/
+[documentation]: https://glass.readthedocs.io/stable
+[examples]: https://glass.readthedocs.io/projects/examples/stable
```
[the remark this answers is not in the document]
ntessore: Not yet because that's only on latest for now. But will work with the next release.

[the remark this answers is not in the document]
paddyroddy: Ah I see, okay will change this

pyproject.toml
paddyroddy: @ntessore think its just this comment left

ntessore: Indeed, makes sense to have it at stable!

ntessore (approved):
Thanks @paddyroddy looking good!

Issue #419, opened 2024-11-14
Some imports aren't at the top of the file


paddyroddy:
### Describe the Bug

Ignoring the `scipy`/`fitsio` conditional testing ones

### To Reproduce

```python
import glass
...
```

### Version In Use

2024.2.dev341+gf92c39c.d20241113

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Fixed in #420.

PR #420, opened 2024-11-14
gh-419: fix imports

paddyroddy:
Some imports weren't at the top. Also spotted a double `TYPE_CHECKING` block.


Saransh-cpp (approved):
Nice catch!

PR #422, opened 2024-11-14
gh-274: add missing tests for `glass.user`

### LINKED ISSUE #274: Add/review tests for `glass.user`

Part of #161

_No response_

_No response_

### Additional Context

_No response_

Saransh-cpp:
Closes #274. Adds tests for `glass.save_cls` and `glass.load_cls`.


tests/test_fits.py
Saransh-cpp, line 1: The file is named `tests/test_fits.py` but all the `fits` related functions are present in `glass.user`. Should this be renamed?

tests/test_fits.py
```diff
@@ -14,6 +14,25 @@
...
+    with np.load(tmp_path / cls_file) as npz:
+        values = npz["values"]
+        split = npz["split"]
+
+    np.testing.assert_allclose(values, np.concatenate(cls))
+    np.testing.assert_allclose(split, np.cumsum([len(cl) for cl in cls[:-1]]))
+    np.testing.assert_allclose(cls, np.split(values, split))
+
+    npz = load_cls(tmp_path / cls_file)
+    np.testing.assert_allclose(npz, cls)
```
Saransh-cpp, line 35: I could not find a better way of testing these functions in 2 independent tests. I could do that if I don't use the `tmp_path` fixture. Both cases have pros and cons, but I don't have any strong opinions.

ntessore: I think this is fine, we aren't testing that we adhere to an external file format here, so we don't need to test that `numpy.savez()` doesn't change.

tests/test_fits.py
ntessore: Makes sense!

tests/test_fits.py
```diff
@@ -14,6 +14,25 @@
...
+    cls = rng.normal(size=(10, 10))
+    save_cls(tmp_path / cls_file, cls)
+
+    assert pathlib.Path.exists(tmp_path / cls_file)
+
+    with np.load(tmp_path / cls_file) as npz:
+        values = npz["values"]
+        split = npz["split"]
+
+    np.testing.assert_allclose(values, np.concatenate(cls))
```
paddyroddy, line 30: Why `assert_allclose` here? Wouldn't https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html be more appropriate?

Saransh-cpp: Oh yes, thanks!

paddyroddy (approved):
Nice!

PR #425, opened 2024-11-15
gh-398: consistent plotting for notebooks

### LINKED ISSUE #398: Add a `matplotlibrc` to control the resolution of the notebooks figures
We probably want to add a matplotlibrc or similar to control at least the resolution of the notebook figures – you can see here that the images still depend on who generated them.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/pull/389#pullrequestreview-2422045974_

paddyroddy:
Closes #398. Adds a `.matplotlibrc` to stabilise the production of plots. For more detail on the config file, look at https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file. I haven't used a stylesheet (`plt.style.use`), as (except for `classic`) they don't define the resolution etc. https://github.com/matplotlib/matplotlib/tree/main/lib/matplotlib/mpl-data/stylelib.

I did run the notebooks, but they must be using these current values.


examples/2-advanced/stage_4_galaxies.ipynb
```diff
@@ -40,6 +43,9 @@
     "import glass\n",
     "import glass.ext.camb\n",
     "\n",
+    "# set up matplotlib to produce consistent figures\n",
+    "mpl.rc_file(pathlib.Path.cwd().parents[1] / \".matplotlibrc\")\n",
+    "\n",
```
Saransh-cpp, line 48: I though it automatically picks up the config from the config file in the current working directory? Is something like that not possible?

paddyroddy: From what I've read it has to be `maplotlibrc` (i.e. without the dot). Plus we have notebooks in two different subfolders, so I think it's safer to force load.

paddyroddy:
@ntessore I've added `matplotlibrc` symlink to both folders instead

paddyroddy:
> Thanks @paddyroddy, I think this is a nice and clean solution!
> 
> > I did run the notebooks, but they must be using these current values.
> 
> This surprised me, so I had a look. I guess you used our noxfile ... where we forgot to add the `--inplace` flag to the `jupyter execute` so that the execution actually changes the notebooks! At the moment, the notebooks run, but their outputs aren't stored anywhere (and our examples' HTML artefacts therefore never change). That's on me, I should have picked up on it in #367.

No, I ran them manually and verified it worked by changing the top level `.matplotlibrc`

PR #427, opened 2024-11-15
gh-421: fix TestPyPI deployment

### LINKED ISSUE #421: Can't deploy to TestPyPI
### Describe the Bug

The credentials aren't set correctly. Found in #416.

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

N/A

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Closes #421. Allows one to use `workflow_dispatch` to deploy the current setup to TestPyPI, this should allow us to test things work as expected.

However, it still ensures that the `publish` workflow is still protected by the GitHub environment. This is achieved by changing the name of the job to `test-publish` when called with called by `workflow_dispatch`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,10 +32,16 @@ jobs:
     needs: dist
     runs-on: ubuntu-latest
     environment:
-      name: ${{ github.event_name == 'release' && 'publish' || 'test-publish' }}
+      name: >-
+        ${{ (github.event_name == 'release' &&
+         github.event.action == 'published') &&
+         'publish' ||
+         'test-publish' }}
       url: >-
-        ${{ github.event_name == 'release' && 'https://pypi.org/project/glass'
-        || 'https://test.pypi.org/project/glass' }}
+        ${{ (github.event_name == 'release' &&
+         github.event.action == 'published') &&
+         'https://pypi.org/project/glass' ||
+         'https://test.pypi.org/project/glass' }}
     permissions:
       id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -51,13 +49,11 @@ jobs:
         run: ls -l dist/
 
       - name: Publish to PyPI
-        if: >-
-          github.event.inputs.target == 'pypi' || (github.event_name ==
-          'release' && github.event.action == 'published')
+        if: github.event_name == 'release'
```
ntessore, line 52: I think this is missing that the event.action needs to be "published", or it will also fire on other actions.

.github/workflows/release.yml
```diff
@@ -51,8 +47,16 @@ jobs:
...
+      name: >-
+        ${{ (github.event_name == 'release' &&
+         github.event.action == 'published') &&
+         'publish' ||
+         'test-publish' }}
+      url: >-
+        ${{ (github.event_name == 'release' &&
+         github.event.action == 'published') &&
+         'https://pypi.org/project/glass' ||
+         'https://test.pypi.org/project/glass' }}
```
Saransh-cpp, line 59: TIL: this works as inline if-else

Saransh-cpp (approved):
Thanks, @paddyroddy! Looks good!

.github/workflows/release.yml
paddyroddy: It's a weird syntax

.github/workflows/release.yml
ntessore: Isn't this still open?

paddyroddy: Fixed. I did it for `url` and `name` but forgot to here.

paddyroddy:
Merging as would be useful to test for future deployments

PR #428, opened 2024-11-15
gh-418: add variable names for `Yields` vars

### LINKED ISSUE #418: Single-line yields render as parameter names in docs
Fantastic effort, particularly with the many changes in specs! There is a remaining issue with single-line yields but it's so minuscule that I want to tackle it separately. Towards a new release!

_Originally posted by @ntessore in https://github.com/glass-dev/glass/pull/381#pullrequestreview-2433892737_

paddyroddy:
Closes #418. Without these names, they render as "special-cased".

### DIFF
--- a/glass/points.py
+++ b/glass/points.py
@@ -342,11 +342,11 @@ def uniform_positions(
 
     Yields
     ------
-    lon:
+    lon
         Columns of longitudes for the sampled points.
-    lat:
+    lat
         Columns of latitudes for the sampled points.
-    count:
+    count
         The number of sampled points. For array inputs, an array of
         counts with the same shape is returned.
 
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -62,7 +62,12 @@ def iternorm(
 
     Yields
     ------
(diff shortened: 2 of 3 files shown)


glass/points.py
```diff
@@ -342,11 +342,11 @@ def uniform_positions(
...
     Yields
     ------
-    lon:
+    lon
         Columns of longitudes for the sampled points.
-    lat:
+    lat
         Columns of latitudes for the sampled points.
-    count:
+    count
```
paddyroddy, line 349: They shouldn't be here. Unintentionally introduced in #381.

ntessore: Are these not necessary to make the names show up as names (bold) instead of types (italic)?

ntessore: Excellent, it renders fine on RTD! 🚀

ntessore (approved):
Thanks @paddyroddy!

glass/points.py
paddyroddy: Nope, I think it expects the type if you have them. We don't use colons anywhere else (like in the parameters).

PR #429, opened 2024-11-15
gh-430: fix for release action

ntessore:
This should fix the release action. The git tag is stored into the environment after the README.md is made dirty, where setuptools-scm picks it up as the current version.

Closes: #430

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Replace stable version with the release version
+        if: github.event_name == 'release' && github.event.action == 'published'
         run: |-
           # store the release tag
           tag="${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


Saransh-cpp (approved):
Thanks for looking into this, @ntessore!

paddyroddy:
Good spot @ntessore. I'd overlook this 🤔. Your approach is nice and simple.

We could actually fix the "dirty" checkout by doing this

```yaml
      - name: Update version references
        run: |
          tag="${{ github.ref_name }}"
          sed --regexp-extended \
            --expression "s#(readthedocs\.org/projects/glass/badge/\?version=)stable#\1${tag}#g" \
            --expression "s#(glass\.readthedocs\.io.*?)/stable#\1/${tag}#g" \
            README.md > README.modified
            
      - name: Upload modified README
        uses: actions/upload-artifact@v3
        with:
          name: modified-readme
          path: README.modified
```

And downloading the `README.modified` as `README.md` in a subsequent step. What do you think?

paddyroddy:
Actually, I don't think this would work @ntessore. We won't have access to `git tag --describe` in there.

PR #442, opened 2024-11-20
gh-441: import directly from `glass` in tests

### LINKED ISSUE #441: Tests should import directly from `glass` rather than the module
### Describe the Bug

This is how users are using the module, so probably best we do it in the same way

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

2024.3.dev195+g06b22a5.d20241120

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Closes #441. Closer replicates how users interact with the library. Means that users (/tests) don't need to know explicitly where the functions come from.


tests/conftest.py
```diff
@@ -4,6 +4,4 @@
 
 @pytest.fixture(scope="session")
 def rng() -> np.random.Generator:
-    import numpy as np
-
```
paddyroddy, line 8: Redundant import

ntessore:
Why does it need these type ignore? Should we fix the main namespace so that mypy understands what we are importing?

paddyroddy:
> Why does it need these type ignore? Should we fix the main namespace so that mypy understands what we are importing?

Fixed in #gh-445

ntessore (approved):

PR #444, opened 2024-11-20
gh-443: `if` should be `elif` in `fixed_zbins`

### LINKED ISSUE #443: There's an `elif` bug in `fixed_zbins`
### Describe the Bug

Discovered in #436. The function allows a user to use `nbins` or `dz`, but not both. The current implementation does not allow one to select `nbins`.

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

2024.3.dev195+g06b22a5.d20241120

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Closes #443. In theory one should be able to use `nbins` or `dz`, but the current implementation actually forbids `nbins`.

### DIFF
--- a/glass/observations.py
+++ b/glass/observations.py
@@ -225,7 +225,7 @@ def fixed_zbins(
     """
     if nbins is not None and dz is None:
         zbinedges = np.linspace(zmin, zmax, nbins + 1)
-    if nbins is None and dz is not None:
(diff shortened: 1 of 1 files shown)


ntessore (approved):
Well spotted!

Issue #411, opened 2024-11-11
Unhide `kappa_ia_nla`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Following #400, once the function is fixed we should unhide the `kappa_ia_nla` function

PR #446, opened 2024-11-21
gh-445: place all public function in `__all__`

### LINKED ISSUE #445: Should we fix the main namespace so that mypy understands what we are importing?
Why does it need these type ignore? Should we fix the main namespace so that mypy understands what we are importing?

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/442#issuecomment-2489458922_

paddyroddy:
Closes #445. Looks like this was missed in #221. All public functions need to be in [__all__](https://stackoverflow.com/a/35710527/7359333). This will address the comments from #442.


glass/__init__.py
```diff
@@ -19,7 +19,6 @@
     transform_cls,
 )
 from glass.galaxies import (
-    _kappa_ia_nla,
```
paddyroddy, line 22: Makes sense to remove from here until it's time to tackle #411.

paddyroddy:
> Excellent, sorry I missed that. Let's [follow PEP8](https://peps.python.org/pep-0008/#module-level-dunder-names) and keep `__all__` at the top of the module.

Always assumed it had to go after the imports, but clearly that isn't the case. Have changed.

ntessore (approved):

Issue #450, opened 2024-11-21
Move `pytest.fixture` to `conftest.py`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

I tend to put all `pytest.fixture` in `conftest.py`. This is also useful as the mock `cosmo` object is needed by multiple files.

Saransh-cpp:
Closed in #451

PR #449, opened 2024-11-21
gh-273: add tests for `glass.shells`

paddyroddy:
Closes #273. Adds tests for the `glass.shells` module. Coverage going up - 94.1%.

Worked on this before the pause. Can't remember where I got to, but makes sense to merge. We can always improve the tests later.

PR #451, opened 2024-11-21
gh-450: move all `pytest.fixture` into `conftest.py`

paddyroddy:
This makes them available to all tests. I need `cosmo` for use in #449.

Have also added `scope="session"` so they are only computed once when testing rather than multiple times.


Saransh-cpp (approved):
Thanks, much cleaner now!

Issue #460, opened 2024-11-28
`nox` is missing `--inplace` so HTML artefacts never change


paddyroddy:
This surprised me, so I had a look. I guess you used our noxfile ... where we forgot to add the `--inplace` flag to the `jupyter execute` so that the execution actually changes the notebooks! At the moment, the notebooks run, but their outputs aren't stored anywhere (and our examples' HTML artefacts therefore never change). That's on me, I should have picked up on it in #367.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/425#issuecomment-2504931387_

Saransh-cpp:
Ah, I see it now!

PR #461, opened 2024-11-28
gh-460: update notebooks inplace when running nox -s examples

Saransh-cpp:
`nox -s examples` now updates the notebooks inplace.

Closes: #460


ntessore:
Yeah, that's on me -- you had it right first time when you executed the notebooks to generate the HTML. If we don't want to execute twice, the first execution obviously has to have side effects :)

ntessore (approved):

paddyroddy (approved):
LGTM

Issue #424, opened 2024-11-14
Consider a pull request template for breaking changes


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Making releases (#390) are hard currently due to the need to inform users about possible breaking changes, as the API isn't guaranteed to be backwards compatible. Is there a way we could improve PRs so that making the necessary CHANGELOG would be easy in future?

Saransh-cpp:
One thing that we do at [PyBaMM](https://github.com/pybamm-team/PyBaMM/blob/develop/.github/PULL_REQUEST_TEMPLATE.md) is make developers add a CHANGELOG note in every PR that introduces a user facing change. Should we do that in GLASS as well, or should I just add a PR template that makes it easier for developers to write what user-facing change this PR is adding?

cc: @ntessore

paddyroddy:
I guess it would be good to document what we want in this. Some thoughts:
- [ ] Is your code passing linting?
- [ ] Is your code passing tests?
- [ ] Have you written tests if it's a new feature?
- [ ] Have you modified/extended the documentation?
- [ ] Are you adding a public function which should be exposed in `sphinx`?

PR #463, opened 2024-11-28
gh-424: add a simple PR template

Saransh-cpp:
Adds a PR template to make the CHANGELOG compiling process easier.

Closes: #424


paddyroddy (approved):
Hopefully this will prove useful and not slow down velocity 🤞

PR #396, opened 2024-11-06
gh-342: explicit libbz2-dev installation is not required now

Saransh-cpp:
None of the available Ubuntu images - [Ubuntu 20.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md#installed-apt-packages), [Ubuntu 22.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#installed-apt-packages), and [Ubuntu 24.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-apt-packages) - list it as an installed `apt` package so I am not sure how or when this was fixed.

Refs: #342

PR #471, opened 2025-01-13
gh-468: fix virtualenv in testing

### LINKED ISSUE #468: `python is not installed into the virtualenv` in testing
### Describe the Bug

Not sure when this regression started, but it's between #466 and #467. It is something to do with `nox`. My experience is largely with `tox` so I'm a bit stumped. The error (confirmed locally) is
```bash
❯ nox -s doctests-3.11
nox > Running session doctests-3.11
nox > Creating virtual environment (virtualenv) using python3.11 in .nox/doctests-3-11
nox > Command /Users/paddy/arc/euclid/glass/.venv/bin/python3 -m virtualenv /Users/paddy/arc/euclid/glass/.nox/doctests-3-11 -p python3.11 failed with exit code 1:
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.11'
nox > Session doctests-3.11 failed.
```

This looks like it could be a nox-specific error

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

2024.3.dev20+g371300b

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
Fixes #468, #479. The problem initially why the tests were failing was https://github.com/esheldon/fitsio/issues/413 https://github.com/esheldon/fitsio/issues/414. I have solved this by reinstating #396.

However, the tests were failing (just for `3.11`) for a very niche `nox` bug https://github.com/wntrblm/nox/issues/735. I'm still not particularly happy with `nox`, the ecosystem seems immature. I considered switching to `tox` https://github.com/glass-dev/glass/issues/475, as I haven't had those issues before. However, this proved a bit tricky as I didn't want to change too much how we currently our running things, and we're relying on a fair bit of Python logic. In the end, this comment https://github.com/wntrblm/nox/issues/735#issuecomment-1854599576 referred to about modifying the cache and that seems to have solved the issue.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,9 +53,7 @@ jobs:
         run: python -m pip install coverage[toml] nox
 
       - name: Install ubuntu dependencies for fitsio
-        run: |-
-          sudo apt-get update
-          sudo apt-get install -y libbz2-dev
+        run: sudo apt-get install -y libbz2-dev
 
       - name: Run doctests
         run: nox -s doctests-${{ matrix.python-version }} --verbose
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,11 @@ jobs:
       - name: Install nox and coverage.py
         run: python -m pip install coverage[toml] nox
 
+      - name: Install ubuntu dependencies for fitsio
+        run: |-
+          sudo apt-get update
+          sudo apt-get install -y libbz2-dev
+
       - name: Run doctests
         run: nox -s doctests-${{ matrix.python-version }} --verbose
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
     name: Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version:
           - "3.10"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Cache nox
         uses: actions/cache@v4
         with:
-          key: test-${{ hashFiles('pyproject.toml') }}
+          key: test-${{ hashFiles('pyproject.toml') }}-${{ env.pythonLocation }}
           path: .nox
 
       - name: Set up Python
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,6 @@ jobs:
     name: Tests
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         python-version:
           - "3.10"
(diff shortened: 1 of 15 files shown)


noxfile.py
```diff
@@ -7,8 +7,12 @@
 import nox
 
 # Options to modify nox behaviour
+nox.options.default_venv_backend = "uv|virtualenv"
```
paddyroddy, line 10: Fixed my issues locally using `uv` https://github.com/astral-sh/uv/issues/6579

.github/workflows/test.yml
```diff
@@ -51,6 +52,11 @@ jobs:
       - name: Install nox and coverage.py
         run: python -m pip install coverage[toml] nox
 
+      - name: Install ubuntu dependencies for fitsio
+        run: |-
+          sudo apt-get update
+          sudo apt-get install -y libbz2-dev
```
paddyroddy, line 58: See https://github.com/esheldon/fitsio/issues/414

.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Cache nox
         uses: actions/cache@v4
         with:
-          key: test-${{ hashFiles('pyproject.toml') }}
+          key: test-${{ hashFiles('pyproject.toml') }}-${{ env.pythonLocation }}
```
paddyroddy, line 41: See https://github.com/wntrblm/nox/issues/735#issuecomment-1854599576

Saransh-cpp: This is weird. I have never encountered this bug before.

glass/core/array.py
```diff
@@ -166,7 +166,7 @@ def trapezoid_product(
     y = np.interp(x, *f)
     for f_ in ff:
         y *= np.interp(x, *f_)
-    return np.atleast_1d(np.trapezoid(y, x, axis=axis))
+    return np.trapezoid(y, x, axis=axis)  # type: ignore[no-any-return]
```
Saransh-cpp, line 169: Should the mypy fixes be another PR?

noxfile.py
Saransh-cpp: I remember we could not use `uv` because of https://github.com/astral-sh/uv/issues/1794

I am surprised that it is working now. We still use the `--prefer-binary` flag in our tests -

https://github.com/glass-dev/glass/blob/f6cfde4487c0d1c1c155f765d2cc6da48fa9a911/.github/test-constraints.txt#L10

Saransh-cpp (approved):
Thanks for digging into this, @paddyroddy! See my comments below -

glass/core/array.py
paddyroddy: I hadn't asked for a review... The PR was open to get the tests running. I've merged in #484.

Saransh-cpp: Ah, my bad!

paddyroddy:
Merging this, linting still failing.

Issue #465, opened 2024-11-29
Enable installing GLASS from git archives


Saransh-cpp:
### Add your issue here

Given that we use hatch-vcs, users cannot install GLASS through git archives. This can be enabled by following [setuptools-scm's documentation](https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives).

PR #483, opened 2025-01-17
gh-465: enable installing glass through git archives

Saransh-cpp:
This is required if glass is installed through the files in GitHub releases.

Closes: #465


paddyroddy (approved):
Interesting

Issue #480, opened 2025-01-17
Fix mypy failures


Saransh-cpp:
### Add your issue here

Mypy is failing with the error -

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

glass/core/array.py:169: error: Unused "type: ignore" comment  [unused-ignore]
glass/core/array.py:169: error: Incompatible return value type (got "float64 | ndarray[tuple[int, ...], dtype[float64]]", expected "ndarray[tuple[int, ...], dtype[float64]]")  [return-value]
glass/core/array.py:169: note: Error code "return-value" not covered by "type: ignore" comment
glass/shells.py:239: error: Argument 1 has incompatible type "ndarray[tuple[int, ...], dtype[floating[Any]]]"; expected "Sequence[float] | ndarray[tuple[int, ...], dtype[float64]]"  [arg-type]
glass/shells.py:241: error: Argument 1 to "RadialWindow" has incompatible type "ndarray[tuple[int, ...], dtype[floating[Any]]]"; expected "ndarray[tuple[int, ...], dtype[float64]]"  [arg-type]
glass/shells.py:413: error: Incompatible return value type (got "tuple[ndarray[tuple[int, ...], dtype[Any]], ndarray[tuple[int, ...], dtype[floating[Any]]]]", expected "tuple[ndarray[tuple[int, ...], dtype[float64]], ndarray[tuple[int, ...], dtype[float64]]]")  [return-value]
glass/shells.py:568: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[floating[_64Bit]]]", variable has type "ndarray[tuple[int, ...], dtype[float64]]")  [assignment]
glass/shells.py:632: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[floating[_64Bit]]]", variable has type "ndarray[tuple[int, ...], dtype[float64]]")  [assignment]
[...]

PR #490, opened 2025-01-28
gh-480: fix mypy failures

Saransh-cpp:
# Description

Fixes most of the errors. See my comments below for a better explanation.

Refs: #480 

## Checks

### DIFF
--- a/glass/core/array.py
+++ b/glass/core/array.py
@@ -138,7 +138,7 @@ def trapezoid_product(
         npt.NDArray[np.float64],
     ],
     axis: int = -1,
-) -> npt.NDArray[np.float64]:
+) -> float | npt.NDArray[np.float64]:
     """
     Trapezoidal rule for a product of functions.
 
@@ -166,7 +166,7 @@ def trapezoid_product(
     y = np.interp(x, *f)
     for f_ in ff:
         y *= np.interp(x, *f_)
-    return np.trapezoid(y, x, axis=axis)  # type: ignore[return-value]
+    return np.trapezoid(y, x, axis=axis)
 
 
 def cumulative_trapezoid(
--- a/glass/lensing.py
+++ b/glass/lensing.py
@@ -410,8 +410,8 @@ class MultiPlaneConvergence:
         self.cosmo = cosmo
 
         # set up initial values of variables
-        self.z2: float = 0.0
-        self.z3: float = 0.0
+        self.z2: float | npt.NDArray[np.float64] = 0.0
+        self.z3: float | npt.NDArray[np.float64] = 0.0
         self.x3: float = 0.0
         self.w3: float = 0.0
         self.r23: float = 1.0
@@ -442,7 +442,7 @@ class MultiPlaneConvergence:
     def add_plane(
         self,
         delta: npt.NDArray[np.float64],
-        zsrc: float,
+        zsrc: float | npt.NDArray[np.float64],
         wlens: float = 1.0,
     ) -> None:
         """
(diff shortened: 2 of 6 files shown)


glass/core/array.py
```diff
@@ -166,7 +166,7 @@ def trapezoid_product(
     y = np.interp(x, *f)
     for f_ in ff:
         y *= np.interp(x, *f_)
-    return np.trapezoid(y, x, axis=axis)  # type: ignore[return-value]
+    return np.trapezoid(y, x, axis=axis)
```
Saransh-cpp, line 169: The failing `ignore[return-value]` checks were on numpy 2.2.0, and should be resolved now.

glass/points.py
```diff
@@ -52,7 +52,7 @@ def effective_bias(
     z: npt.NDArray[np.float64],
     bz: npt.NDArray[np.float64],
     w: RadialWindow,
-) -> npt.NDArray[np.float64]:
+) -> float | npt.NDArray[np.double]:
```
Saransh-cpp, line 55: - `np.trapezoid` can return a `float`
- Division between `np.float64` and `np.float64` can result in `np.double` ("float-any") for some reason.

glass/shells.py
```diff
@@ -742,7 +742,7 @@ def _uniform_grid(
     if step is not None and num is None:
         return np.arange(start, np.nextafter(stop + step, stop), step)
     if step is None and num is not None:
-        return np.linspace(start, stop, num + 1)  # type: ignore[return-value]
+        return np.linspace(start, stop, num + 1, dtype=np.float64)
```
Saransh-cpp, line 745: Returns "float-any" so be explicit about the type

glass/shells.py
```diff
@@ -235,10 +235,10 @@ def tophat_windows(
     ws = []
     for zmin, zmax in itertools.pairwise(zbins):
         n = max(round((zmax - zmin) / dz), 2)
-        z = np.linspace(zmin, zmax, n)
-        w = wht(z)  # type: ignore[arg-type]
+        z = np.linspace(zmin, zmax, n, dtype=np.float64)
+        w = wht(z)
         zeff = np.trapezoid(w * z, z) / np.trapezoid(w, z)
```
Saransh-cpp, line 240: `np.trapezoid` can return an array and I don't think they have enough overloads in place -- therefore `zeff` can be an array and the type signature of `RadialWindow` should reflect this

glass/lensing.py
```diff
@@ -442,7 +442,7 @@ def add_window(self, delta: npt.NDArray[np.float64], w: RadialWindow) -> None:
     def add_plane(
         self,
         delta: npt.NDArray[np.float64],
-        zsrc: float,
+        zsrc: float | npt.NDArray[np.float64],
```
Saransh-cpp, line 445: If `zeff` can be an array, `zsrc` in `add_window` can be one too, and `add_window` passes `zsrc` to `add_plane`

glass/lensing.py
```diff
@@ -410,8 +410,8 @@ def __init__(self, cosmo: Cosmology) -> None:
         self.cosmo = cosmo
 
         # set up initial values of variables
-        self.z2: float = 0.0
-        self.z3: float = 0.0
+        self.z2: float | npt.NDArray[np.float64] = 0.0
+        self.z3: float | npt.NDArray[np.float64] = 0.0
```
Saransh-cpp, line 414: Because of `zsrc`'s new type signature

ntessore (approved):

ntessore:
Merging this to unblock other PRs' tests.

PR #505, opened 2025-02-13
gh-506: cache dependencies correctly

Saransh-cpp:
# Description

- The nox environment should be cached after setup-python cache (as setup-python exposes env.pythonLocation which is used in nox cache)
- noxfile's hash should be used in the key while caching nox environments
- env.pythonLocation should be used in the cache key to not confuse the runners between system-wide and environment-specific python

Closes: #506 

## Checks

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -32,18 +32,20 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


paddyroddy (approved):
Let's hope this works

ntessore:
Nice one!

Issue #503, opened 2025-02-13
Switch over to importing the types from `typing`


paddyroddy:
I'd rather switch over to importing the types from `typing` like most of the ecosystem does. We can open an issue to make the rest of the code consistent.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/pull/494#discussion_r1953050481_

PR #507, opened 2025-02-13
gh-503: move all typing related imports into `from`

paddyroddy:
# Description

Following discussions, I have moved all `typing`, `collections.abc`, `numpy.typing` into `from` type imports. I have moved stuff into a `TYPE_CHECKING` section where appropriate.

Closes: #503

## Checks


paddyroddy:
Hmm... typing bug, which isn't happening with `import typing`

paddyroddy:
No idea why docs are failing? Can @ntessore manually re-run it?

glass/fields.py
```diff
@@ -3,22 +3,20 @@
...
 from gaussiancl import gaussiancl
 from transformcl import cltovar
 
 from glass import grf
 
 if TYPE_CHECKING:
-    import collections.abc
-    from collections.abc import Callable, Iterable, Iterator, Sequence
+    from collections.abc import Callable, Generator, Iterable, Iterator, Sequence
     from typing import Any
```
Saransh-cpp, line 20:
```suggestion
```

Saransh-cpp (changes requested):
Thanks, @paddyroddy! Minor repetition below -

glass/fields.py
paddyroddy: Good spot, although seems like I can remove the one outside the `TYPE_CHECKING` block

paddyroddy:
> No idea why docs are failing? Can @ntessore manually re-run it?

Docs passing now

Saransh-cpp (approved):

glass/fields.py
```diff
@@ -44,11 +42,9 @@ def inv_triangle_number(triangle_number: int) -> int:
 
 def iternorm(
     k: int,
-    cov: collections.abc.Iterable[npt.NDArray[np.float64]],
+    cov: Iterable[NDArray[np.float64]],
     size: int | tuple[int, ...] = (),
-) -> collections.abc.Generator[
-    tuple[int | None, npt.NDArray[np.float64], npt.NDArray[np.float64]]
-]:
+) -> Generator[tuple[int | None, NDArray[np.float64], NDArray[np.float64]]]:
```
ntessore, line 47: Not really a comment on the PR as such, but the unary `Generator` here and below could be changed to `Iterator` which is the same.

glass/shells.py
```diff
@@ -45,28 +45,28 @@
...
 from glass.core.array import ndinterp
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from cosmology import Cosmology
 
-ArrayLike1D = collections.abc.Sequence[float] | npt.NDArray[np.float64]
-WeightFunc = typing.Callable[[ArrayLike1D], npt.NDArray[np.float64]]
+ArrayLike1D = Sequence[float] | NDArray[np.float64]
+WeightFunc = Callable[[ArrayLike1D], NDArray[np.float64]]
```
ntessore, line 63: Can these go in the TYPE_CHECKING block?

glass/core/array.py
```diff
@@ -2,19 +2,21 @@
 
 from __future__ import annotations
 
-import typing
 from functools import partial
+from typing import TYPE_CHECKING
 
 import numpy as np
-import numpy.typing as npt
+from typing_extensions import Unpack
```
ntessore, line 9: Move into TYPE_CHECKING block and remove dependency

ntessore (changes requested):
Nice!! Removes so much visual clutter in the function signatures.

ntessore:
This shouldn't have been merged without review

paddyroddy:
> This shouldn't have been merged without review

I considered this trivial and had been ready for a while

glass/shells.py
paddyroddy: In my experience, only imports really go inside `TYPE_CHECKING`

glass/core/array.py
paddyroddy: Good idea, have raised https://github.com/glass-dev/glass/pull/517

glass/shells.py
ntessore: The `TYPE_CHECKING` block is for everything that doesn't matter at runtime. An example I saw earlier is e.g. [here](https://github.com/python/mypy/blob/8104d0198813a59dd26c8e9f50779533c60292b4/mypyc/build.py#L50-L59) in mypy itself. There are far fewer type aliases than type imports, so maybe that's the cause?

paddyroddy: If it causes no harm then happy to do, will raise a PR now

ntessore: @paddyroddy Causes no harm either way, so maybe we just do it if we ever touch that code again?

paddyroddy: https://github.com/glass-dev/glass/pull/520

Issue #491, opened 2025-01-28
glass namespace contains unwanted symbols


ntessore:
### Describe the Bug

These add noise to the `glass` namespace:

```py
import contextlib
from importlib.metadata import PackageNotFoundError

with contextlib.suppress(PackageNotFoundError):
    from ._version import __version__, __version_tuple__
```

and `try: ... catch ImportError: pass` should work just as well?

### To Reproduce

```python
import glass
print(glass.contextlib, glass.PackageNotFoundError)
```

### Expected Behaviour

Not have `contextlib` and `PackageNotFoundError` in the `glass` namespace.

### Actual Behaviour

Have `contextlib` and `PackageNotFoundError` in the `glass` namespace.

### Version In Use

2024.3

### Additional Context

```markdown

```

ntessore:
I think we need to suppress the import error since the `_version` isn't generated unless the code is installed, e.g., on `git clone`.

Saransh-cpp:
I see, usually it is not recommended to import a package without installing it, but we can switch to `try-except` if that is a use-case. I believe `contextlib` was used to follow suggestions from `ruff`. No strong opinions on polluting the namespace, but usually it is okay to have such imports in `__init__.py` (and not include them in `__all__`).

ntessore:
To be fair, it's been a while since I used `package.__version__` now that `importlib.metadata.version()` is available out of the box. Perhaps the import isn't even necessary?

Saransh-cpp:
`package.__version__` is still used at various places; hence, we should not remove it. Tagging @paddyroddy here for more opinions.

paddyroddy:
Interesting discussion here https://discuss.python.org/t/please-make-package-version-go-away/58501, too long so I've read a tiny fraction.

According to https://packaging.python.org/en/latest/discussions/single-source-version/

>When the intention is that a distribution package and its associated import package share the same version, it is recommended that the project include an automated test case that ensures import_name.__version__ and importlib.metadata.version("dist-name") report the same value (note: for many projects, import_name and dist-name will be the same name).

so it does feel like we should have the `__version__` attribute.

PR #511, opened 2025-02-13
gh-491: use try-except for __version__

Saransh-cpp:
# Description

Use a `try-except` block instead of `contextlib.suppress` so that the `glass` namespace is not polluted.

Also, the error we should be looking for is `ModuleNotFoundError` and not `PackageNotFoundError`.

Closes: #491

## Checks

### DIFF
--- a/glass/__init__.py
+++ b/glass/__init__.py
@@ -63,11 +63,10 @@ __all__ = [
 ]
 
 
-import contextlib
-from importlib.metadata import PackageNotFoundError
-
-with contextlib.suppress(PackageNotFoundError):
-    from ._version import __version__, __version_tuple__
+try:  # noqa: SIM105
+    from ._version import __version__
+except ModuleNotFoundError:
+    pass
 
 # modules
 from glass import grf


glass/__init__.py
```diff
@@ -63,11 +63,10 @@
 ]
 
 
-import contextlib
-from importlib.metadata import PackageNotFoundError
-
-with contextlib.suppress(PackageNotFoundError):
-    from ._version import __version__, __version_tuple__
+try:  # noqa: SIM105
```
paddyroddy, line 66: Still don't know I feel about this https://docs.astral.sh/ruff/rules/suppressible-exception/

ntessore: :disappointed: `contextlib.suppress()` expresses intent clearer than the idiomatic way to suppress an exception? flake8-simplify need to have a long, hard look at themselves

ntessore: Generally speaking, it seems like ruff has a catalogue of rules from every niche style checker ever written, and I'm not sure a lot of them are good ideas...

ntessore (approved):
:tada:

glass/__init__.py
paddyroddy: I think we should merge this. But ruff has a very active community and unpopular rules have been deprecated. For example, [ANN101](https://docs.astral.sh/ruff/rules/missing-type-self/#missing-type-self-ann101) was removed relatively quickly.

Issue #501, opened 2025-02-12
Switch everything over to `import glass`


Saransh-cpp:
I really don't like `from glass import ...`, also for users. Happy to have a separate PR where we switch everything over to `import glass`.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/pull/494#discussion_r1941448498_

PR #512, opened 2025-02-13
gh-501: change imports to be directly from `glass`

paddyroddy:
# Description

Changes the code throughout to use `import glass` rather than `from glass import` where possible. This is possible via the `__all__` statements in `__init__.py`. I have also done this for `glass.grf`.

Closes: #501

## Checks

### DIFF
--- a/glass/grf/_transformations.py
+++ b/glass/grf/_transformations.py
@@ -8,7 +8,7 @@ from typing import Any
 
 from numpy.typing import NDArray  # noqa: TC002
 
-from glass.grf._core import corr, dcorr, icorr
+from glass.grf import corr, dcorr, icorr
 
 
 @dataclass
--- a/docs/user/how-glass-works.rst
+++ b/docs/user/how-glass-works.rst
@@ -30,13 +30,13 @@ which are flat and non-overlapping.
 
 .. plot::
 
-    from glass.shells import redshift_grid, tophat_windows
+    import glass
 
     # create a redshift grid for shell edges
-    zs = redshift_grid(0., 0.5, dz=0.1)
+    zs = glass.redshift_grid(0., 0.5, dz=0.1)
 
     # create the top hat windows
-    ws = tophat_windows(zs)
+    ws = glass.tophat_windows(zs)
 
     # plot each window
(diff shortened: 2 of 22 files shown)


glass/grf/_transformations.py
```diff
@@ -8,7 +8,7 @@
 
 from numpy.typing import NDArray  # noqa: TC002
 
-from glass.grf._core import corr, dcorr, icorr
+from glass.grf import corr, dcorr, icorr
```
paddyroddy, line 11: This one has to remain `from glass.grf` syntax as otherwise we run into a circular import

Saransh-cpp (commented):
Thanks, looks good to me!

ntessore:
Thanks for doing all those changes!

I hadn't really thought about use of `glass` from inside the library itself, I was only considering tests and examples and "uses" of `glass`. Do you then prefer to use the `glass` namespace everywhere, even in library code itself? I have no opinion either way.

ntessore (approved):

Saransh-cpp (approved):

paddyroddy:
> Do you then prefer to use the `glass` namespace everywhere, even in library code itself? I have no opinion either way.

I don't really mind. I think there are benefits of both.

The way I've done it means that you are using the library as an external user would. So in a sense it's another check that `__all__` etc. is working.

On the other hand, having full path imports means that at a glance you know exactly where something is coming from. This is why I in general prefer `import <x>; <x>.<y>`, rather than `from <x> import <y>`, as at a glance I wouldn't know where `<y>` has come from.

@Saransh-cpp do you have any views of which style we should use here?

Saransh-cpp:
I like `import <x>; <x>.<y>` as well for the same reasons.

Also, all the popular conventions in the Python ecosystem are always of the form `import <x>; <x>.<y>` -
```python
numpy.array
plt.plot
pd.DataFrame
...
```

ntessore:
Ok, then let's always use `glass.` even in GLASS itself. I guess that doesn't cause any more circular import grief than the `from ... import ...` methods, since latter also imports `glass` first. LGTM!!

Issue #513, opened 2025-02-13
Turn algorithms into public module


ntessore:
### Add your issue here

Make `glass.core.algorithms` available as a public module `glass.algorithms`.

PR #514, opened 2025-02-13
gh-513: make algorithm a public module

ntessore:
Move `glass.core.algorithm` to `glass.algorithm` and document it for public use outside of GLASS itself.

Closes: #513


paddyroddy (changes requested):
Can you look at `docs/reference/fields.rst`?

Saransh-cpp:
Ah, missed that, thanks!

paddyroddy (approved):
👋 `core`

Saransh-cpp (approved):

Issue #515, opened 2025-02-14
Coverage is not reported correctly


ntessore:
### Describe the Bug

The coverage isn't reported correctly on coveralls. It looks like paths are missing the `glass/` prefix.

### To Reproduce

Go to https://coveralls.io/github/glass-dev/glass and try to check any file.

### Expected Behaviour

See the file.

### Actual Behaviour

File not found.

### Version In Use

n/a

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #516, opened 2025-02-14
gh-515: fix coveralls

Saransh-cpp:
# Description

Fix the root path for coveralls. Working -

image

image

Closes: #515

## Checks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,6 +75,7 @@ jobs:
         with:
           flag-name: run-${{ matrix.python-version }}
           parallel: true
+          base-path: glass/
 
   aggregate-tests:
     needs: tests


ntessore (approved):
Nice! Any idea why it didn't pick the path up? If I run e.g. `pytest --cov` locally I do get the correct path from coverage.

Saransh-cpp:
Not very sure, but this is definitely a coveralls thing (how coveralls interprets the coverage reports uploaded 😞) and not a `pytest-cov` thing.

paddyroddy:
> Not very sure, but this is definitely a coveralls thing (how coveralls interprets the coverage reports uploaded 😞) and not a `pytest-cov` thing.

Weird I've not had to do this before, i.e. https://coveralls.io/github/astro-informatics/sleplet?branch=main. But I have usually used `src` builds.

paddyroddy:
The docs say

> Path to the root folder of the project the coverage was collected in. Should be used in monorepos so that coveralls can process filenames from your coverage reports correctly (e.g. packages/my-subproject)

So I think it is because we're not using a `src` build

paddyroddy (approved):
Thanks for sorting

PR #517, opened 2025-02-14
gh-501: changes missed in #507

paddyroddy:
# Description

Missed from the merging in #507. Not sure why `ruff` wasn't smart enough to pick this one up. Glad to get rid of the extra dependency though.

## Checks


Saransh-cpp (approved):

ntessore (approved):
Thanks for changing this! 🚀

PR #520, opened 2025-02-14
gh-501: move custom types into `TYPE_CHECKING`

paddyroddy:
# Description

Moves custom types into `TYPE_CHECKING` block as suggested in https://github.com/glass-dev/glass/pull/507#discussion_r1956180229.

## Checks


Saransh-cpp (approved):

ntessore (approved):
That's great! Thanks!

Issue #525, opened 2025-02-19
Fix docs


ntessore:
### Add your issue here

Fix docs build by temporarily pinning to sphinx<8.2.

PR #526, opened 2025-02-19
gh-525: fix docs

ntessore:
Fixes temporary issue with Sphinx 8.2.0. Also should fix the version in latest to not be shown.

Closes: #525


ntessore:
Oh, it's not actually doing a shallow clone, it's fetching to a depth of 50, so that's why we aren't seeing the version.

paddyroddy (approved):

Saransh-cpp (approved):

Issue #487, opened 2025-01-24
Use ragged arrays for alms


ntessore:
### Add your issue here

Inspired by s2fft, I think it would be great to use ragged `alm` arrays. The first axis would be $\ell = 0, \ldots, \ell_{\max}$, the second axis would be ragged with $m = 0, \ldots, \ell$.

ntessore:
See also #382.

Saransh-cpp:
Is s2fft already using ragged arrays from some package right now? I could not find anything relevant. Their paper says -

> From a software engineering perspective, this raises some further complications. Ideally, the FFTs of Eq. [(31)](https://www.sciencedirect.com/science/article/pii/S0021999124003589#fm0310) and Eq. [(32)](https://www.sciencedirect.com/science/article/pii/S0021999124003589#fm0320) would be implemented as a single function call to jax.numpy.fft across a regular array. However, jax.numpy.fft does not support ragged arrays and therefore each ring within the [polar region](https://www.sciencedirect.com/topics/earth-and-planetary-sciences/polar-region) must computed by individual function calls within a scheduled loop.

but I am not sure where the ragged array (that they want to feed in `jax.numpy.fft`) is coming from.

ntessore:
You can see [here](https://astro-informatics.github.io/s2fft/api/sampling/spherical_samples.html#s2fft.sampling.s2_samples.flm_2d_to_hp) the format s2fft uses to store their alms:

```math
    \mathtt{alm}
    = \begin{bmatrix}
        0 & 0 & flm_{(0,0)} & 0 & 0 \\
        0 & flm_{(1,-1)} & flm_{(1,0)} & flm_{(1,1)} & 0 \\
        flm_{(2,-2)} & flm_{(2,-1)} & flm_{(2,0)} & flm_{(2,1)} & flm_{(2,2)}
    \end{bmatrix}
```

This is always for complex fields, so there are negative $m$ values present.

We would continue to use pairs of $E$ and $B$ mode alms instead, so no negative $m$ values. Our arrays could therefore look like this:

```math
    \mathtt{alm}
    = \begin{bmatrix}
        flm_{(0,0)} & 0 & 0 \\
        flm_{(1,0)} & flm_{(1,1)} & 0 \\
        flm_{(2,0)} & flm_{(2,1)} & flm_{(2,2)}
    \end{bmatrix}
```

[...]

PR #527, opened 2025-02-19
gh-509: `inv_triangle_number` -> `nfields_from_nspectra` (plus make it public)

Saransh-cpp:
# Description

Renamed `inv_triangle_number` -> `nfields_from_nspectra`
Made the function public and added it to `__all__`.

Closes: #509

## Checks

- [ ] Is your code passing linting?
- [ ] Is your code passing tests?
- [ ] Have you added additional tests (if required)?
- [ ] Have you modified/extended the documentation (if required)?
- [ ] Have you added a one-liner changelog entry above (if required)?

PR #533, opened 2025-02-21
gh-487: reorder alms in _generate_grf to glass ordering

Saransh-cpp:
# Description

The PR reorder alms in `_generate_grf` to glass ordering before operating on them and then reorders it back to healpix before passing into the yield function.

The code works but it is slowing down examples instead of making them faster. I think I am doing something wrong here. @ntessore could you look at this and point me in the right direction?

Fixes: #487

## Checks


glass/fields.py
```diff
@@ -388,6 +388,8 @@ def _generate_grf(
         # sample real and imaginary parts, then view as complex number
         rng.standard_normal(n * (n + 1), np.float64, z.view(np.float64))
 
+        z = np.asanyarray(healpix_to_glass_spectra(z))
```
Saransh-cpp, line 391: Should the functions return a numpy array instead?

ntessore:
Thanks for taking this on @Saransh-cpp!

I think we crossed wires with two similar but separate things going on at the same time:

* The ordering of the _spectra_. Here we are using GLASS ordering and we want to continue GLASS ordering. The two new functions are only useful for legacy mode interactions between GLASS and other HEALPix-based codes.
* The ordering of the _alms_. Here we are using HEALPix ordering at the moment, and the idea is to switch this around.

This switch is fully internal and there are no observable changes to anything from the outside.

The way our `alms` = $a_{lm}$ are sorted now is keeping $m$ fixes while $l$ increases, so first $m = 0$, $l = 0, 1, 2, \ldots$, then $m = 1$, $l = 1, 2, \ldots$, then $m = 2$, $l = 2, \ldots$, etc. This is encoded in how `multalm()` works.

After the change, our internal generation would keep $l$ fixed while $m$ increases, so first $l = 0$, $m = 0$, then $l = 1$, $m = 0, 1$, then $l = 2$, $m = 0, 1, 2$, etc. Note how this corresponds to switching from iterating over an upper triangular matrix to a lower triangular matrix.

The goal of this internal reordering is that we will be able to split everything up in blocks of $l$ values with different correlation lengths.

After this internal reordering, we still need to send our `alms` to healpy for the spherical harmonic transform. So there needs to be an internal reordering.

paddyroddy:
Can you ask me again for a review once the science is worked out? I'm not fully sure what is going on here

ntessore:
Oh, I just realised this is marked as a fix for #487. That issue is about an unrelated change. The reordering of the alms in the GRF sampling that we discussed is a purely internal change for the time being. Maybe #487 will never be done.

glass/fields.py
```diff
@@ -218,10 +218,18 @@ def multalm(
...
+    n = len(bl)
+    for d in range(n):
+        indices = []
+        for m in range(d, n):
+            # dynamically calculate indices
+            index = m * (m + 1) // 2 + d
+            indices.append(index)
+
+        if indices:
+            out[indices] *= bl[: len(indices)]
```
Saransh-cpp, line 231: The function now assumes that the incoming `alm`s are in the order -
```
[00, 10, 11, 20, 21, 22, 30, 31, 32, 33, ...]
```

Therefore, it now multiplies `alm` with `bl` in the following way -
```
bl := [1, 10, 100, 1000]

[0, 10, 20, 30] * bl → [0, 100, 2000, 30000]
[11, 21, 31] * bl[:3] → [11, 210, 3100]
[22, 32] * bl[:2] → [22, 320]
[33] * bl[:1] → [33]
```

glass/fields.py
```diff
@@ -402,6 +410,8 @@ def _generate_grf(
         if j is not None:
             y[:, j] = z
 
+        alm = np.asanyarray(glass.glass_to_healpix_spectra(alm))  # type: ignore[arg-type]
```
Saransh-cpp, line 413: IIRC, the `alm`s should be converted back to healpix ordering before being passed into `hp` functions. I have used the `glass.glass_to_healpix_spectra` because it does the ordering in the same way.

Saransh-cpp (commented):
@ntessore could you please take a look at this again and let me know if this is the right direction?

`multalm` is definitely slower now because of the dynamic indexing. It would really benefit from using ragged arrays for accessing alms in a NumPy-like way, but we can discuss that in the JAX project. Just using plain NumPy arrays would eat up extra memory because of the matrix structure.

ntessore:
We should really set a timeout on the examples test 😬

glass/fields.py
Saransh-cpp: Ah, I got the multiplication wrong! That's why I was confused about why we are switching to this order. Thanks for the correction!

glass/fields.py
Saransh-cpp: Thanks! Should I just revert #527 so that we don't use functions named for `spectra` inside functions meant for `alm`?

Saransh-cpp: Sounds good! I will wrap that up in another PR, which can be merged before this one to make everything cleaner.

tests/test_fields.py
```diff
@@ -545,6 +546,14 @@ def test_healpix_to_glass_spectra():
     np.testing.assert_array_equal(out, [11, 22, 21, 33, 32, 31, 44, 43, 42, 41])
 
 
+def test_glass_to_healpix_alm():
+    inp = np.array([11, 22, 21, 33, 32, 31, 44, 43, 42, 41])
+    out = glass.fields._glass_to_healpix_alm(inp)
+    np.testing.assert_array_equal(
+        out, np.array([11, 22, 33, 44, 21, 32, 43, 31, 42, 41])
+    )
```
ntessore, line 554: Let's make this more obvious

```suggestion
    inp = np.array([00, 10, 11, 20, 21, 22, 30, 31, 32, 33])
    out = glass.fields._glass_to_healpix_alm(inp)
    np.testing.assert_array_equal(
        out, np.array([00, 10, 20, 30, 11, 21, 31, 22, 32, 33])
    )
```

ntessore (approved):
Thanks for the excellent work @Saransh-cpp! I have reviewed the example outputs from the test run, and they all look reasonable (albeit not exactly the same as previously, given that the same random numbers are now used in different places).

Issue #136, opened 2023-10-19
Addition of new visibility object which allows anisotropy


mwiet:
Within [`glass.observations`](https://github.com/glass-dev/glass/blob/main/glass/observations.py), the addition of a new visibility object, which allows anisotropic and non-binary visibility HEALPix maps which vary with galaxy population/tomographic bin/shell, would make the implementation of variable observational depth possible.

The object should be flexible enough to allow independent visibility maps for each galaxy population/tomographic bin/shell, while also allowing the option of providing a single HEALPix map of an independent variable which traces variable depth plus arrays of functions which characterise the variation with galaxy population/tomographic bin/shell. The latter is particularly useful as it allows for memory efficient modelling of varying depth.

Issue #413, opened 2024-11-11
Split `positions_from_delta()` into individual functions


ntessore:
### Add your issue here

The current `positions_from_delta()` function carries out a large number of steps:

* Apply bias model to given `delta` map.
* Convert biased `delta` to mean number count in each map pixel.
* Apply a visibility map.
* Compute observed number counts from a distribution (e.g., Poisson).
* Accumulate the galaxies in the number count map into batches of a given size (e.g., 1_000_000)
* Sample random positions for each galaxy within its pixel.

This structure is very rigid, and makes it difficult to implement things such as #136.

We should split this up into individual functions, so that they can be mixed and matched with different implementations:

```py
for i, delta in enumerate(matter):
    # mean number count in each pixel
    nbar = glass.nbar_from_density(arcmin2=ngal[i], nside=nside)
    # galaxy density from linear bias model, could be a complicated model
    gal = nbar * (1 + beff[i] * delta)
    # apply visibility map for this shell, could be a complicated model
    gal *= vmap[i]
    # Poisson-sample galaxy indices from map in batches
    for ipix in glass.batched_poisson(gal, rng=rng):
        # randomly sample galaxy positions within each pixel
        lon, lat = glass.random_healpix_positions(nside, ipix)
```

This needs a bit of experimentation for multi-dimensional arrays, but in principle seems feasible.

*Edited*

Steps:

paddyroddy:
> We should split this up into individual functions, so that they can be mixed and matched with different implementations:
> 
> for i, delta in enumerate(matter):
>     # mean number count in each pixel
>     nbar = glass.nbar_from_density(arcmin2=ngal[i], nside=nside)
>     # galaxy density from linear bias model, could be a complicated model
>     gal = nbar * (1 + beff[i] * delta)
>     # apply visibility map for this shell, could be a complicated model
>     gal *= vmap[i]
>     # Poisson-sample galaxy indices from map in batches
>     for ipix in glass.batched_poisson(gal, rng=rng):
>         # randomly sample galaxy positions within each pixel
>         lon, lat = glass.random_healpix_positions(nside, ipix)
> 
> This needs a bit of experimentation for multi-dimensional arrays, but in principle seems feasible.

Just to clarify on this, are you saying this is how you envisage the function being split up? Been trying to map the current function to this in my head, and it's not entirely clear to me.

ntessore:
[...]

PR #535, opened 2025-02-24
gh-413: Implementation of variable depth

mwiet:
This merge request implements variable depth's effect on the galaxy density in the angular and line-of-sight directions into GLASS. This addresses issue #413.

This code allows for functionality similar to [SALMO](https://github.com/Linc-tw/salmo). The implementation adds two novel features:
- The interpolation between the variable depth tracer variable and the change in galaxy density can be interpolated with any function (previously, in SALMO only linear interpolation was possible).
- The effect of variable depth along the line-of-sight and in the angular direction can be modelled as fully independent (previously, in SALMO the selection of galaxy shapes and galaxy redshifts was always assumed to be the same). This allows for the application of a separate selection function for the galaxy shapes and the galaxy redshifts.

The `glass.observations.angular_variable_depth_mask` implements the basic structure and functionality for the object, but only outputs the variable depth masks in the angular direction as they have been input.

The `angular_los_variable_depth_mask` expends upon this functionality by implementing the aformentioned features. Hence, when sampling galaxies using `glass.points.positions_from_delta`, one can multiply the given `vis` visibility mask by the output of `angular_los_variable_depth_mask` for a given tomographic bin and redshift shell index.

# Tests with KiDS-1000-like mock

image

Testing this implementation with KiDS-1000 data (see [here](https://github.com/mwiet/kids_sbi/tree/main/data/KiDS/K1000)) gives results consistent with [previous mocks on the same data](https://arxiv.org/abs/2404.15402):

- Overall redshift distributions are consistent with inputs:
image

- Local redshift distributions are shifted due to variable depth (e.g. North vs. South field):
image

- Between 2% and 10% (on average) of additional power is added to the measured pseudo-Cls from the realised galaxy fields:
image

# Performance

The inclusion of the `angular_los_variable_depth_mask` when sampling the KiDS-1000-like galaxies added an additional 14 seconds to the total runtime on my local machine (from 7m 14s to 7m 28s, i.e. a 3.2% increase). In this time, `angular_los_variable_depth_mask` was evaluated 165 times (n_bins x n_shells = 5 x 33).

# Potential extensions

Note that this implementation does not include the variability in the intrinsic shapes of galaxies which may be caused by variable depth. Additional features will be necessary for this in `glass.shapes` to allow the input `sigma` to vary per pixel.

### DIFF
--- a/glass/observations.py
+++ b/glass/observations.py
@@ -22,6 +22,8 @@ Visibility
 ----------
 
 .. autofunction:: vmap_galactic_ecliptic
+.. class:: angular_variable_depth_mask
+.. class:: angular_los_variable_depth_mask
 
 """  # noqa: D400
 
--- a/glass/observations.py
+++ b/glass/observations.py
@@ -328,3 +328,163 @@ def tomo_nz_gausserr(
     binned_nz *= nz
 
     return binned_nz  # type: ignore[no-any-return]
+    
+class angular_variable_depth_mask:
+    r'''Variable depth mask for tomographic bins.
+    
+    This class allows to create a mask with a different variable depth mask in the angular direction for each tomographic bin.
+    
+    Parameters
+    ----------
+    vardepth_map : array_like
+        Map of variable which traces the depth per tomographic bin.
+    n_bins : int
+        Number of tomographic bins.
+    zbins : array_like
+        Shell redshift limits.
+    '''
+
+    def __init__(self, vardepth_map, n_bins, zbins):
+        self.vardepth_map = vardepth_map
+        self.n_bins = n_bins
+        self.zbins = zbins
+    
+    def test_index(self, index):
+        r'''Test the index for validity.
+        '''
+        if not isinstance(index, tuple):
+            raise ValueError('Index must be an tuple of two integers')
+
+        if index[0] >= self.n_bins:
+            raise ValueError('Leading index cannot exceed number of tomographic bins')
+
+        if index[1] >= len(self.zbins):
+            raise ValueError('Trailing index cannot exceed number of shells')
+        
+    def __getitem__(self, index):
+        r'''Get the mask for the given index.
+
+        Parameters
+        ----------
+        index : tuple
+            Indices of the tomographic bin and shell pair.
+
+        Returns
+        -------
+        mask : array_like
+            Mask for the given index.
+
+        Raises
+        ------
+        ValueError
+            If the index is invalid.
+
+        '''
+        self.test_index(index)
+
+        return self.vardepth_map[index[0]]
+    
+class angular_los_variable_depth_mask(angular_variable_depth_mask):
+    r'''Variable depth mask for tomographic bins.
+    
+    This class allows to create a mask with a different variable depth mask in both the angular and the line-of-sight directions for each tomographic bin.
+    
+    Parameters
+    ----------
+    vardepth_map : array_like
+        Map of variable which traces the depth per tomographic bin. If vardepth_tomo_functions is not provided, the values are treated like a map of galaxy count ratios.
+    n_bins : int
+        Number of tomographic bins.
+    zbins : array_like
+        Shell redshift limits.
+    ztomo : array_like, optional
+        Tomographic redshift bin limits.
+    dndz : array_like, optional
+        Redshift distributions per tomographic bin.
+    z : array_like, optional
+        Redshift domain of dndz.
+    dndz_vardepth : array_like
+        Redshift distribution affected by variable depth (n_bins x len(vardepth_values) x len(z)).
+    vardepth_values : array_like
+        Variable depth tracer domain/values of dndz_vardepth.
+    vardepth_los_tracer : array_like, optional
+        Map of the variable depth tracer for line-of-sight direction. If provided, it is assumed to cover the same domain as vardepth_values.
+    vardepth_tomo_functions : array_like, optional
+        List of functions which map the input vardepth_map to the ratio between the galaxy count due to the variable depth and the galaxy count without variable depth (for each tomographic bin). If provided, it is assumed that there is one vardepth_map which traces the variable depth for all tomographic bins.
+    
+    References
+    ----------
+    .. [1] Joachimi B., Lin, C.-A., et al., 2021, A&A, 646, A129.
+           doi:10.1051/0004-6361/202038831
+    .. [2] von Wietesheim-Kramsta M., Lin, K., et al., 2024, A&A, 695, A223.
+           doi:10.1051/0004-6361/202450487
+    '''
+    def __init__(self, vardepth_map, n_bins, zbins, ztomo, dndz, z, dndz_vardepth, vardepth_values, vardepth_los_tracer=None, vardepth_tomo_functions=None):
+        super().__init__(vardepth_map, n_bins, zbins)
+        self.ztomo = ztomo
+        self.dndz = dndz
+        self.z = z
+        self.dndz_vardepth = dndz_vardepth
+        self.vardepth_values = vardepth_values
+        self.vardepth_los_tracer = vardepth_los_tracer
+        self.vardepth_tomo_functions = vardepth_tomo_functions
+
+        if vardepth_tomo_functions is not None:
+            self.vardepth_map = np.atleast_2d(vardepth_map).reshape(1, -1)
+    
+    def get_los_fraction(self, index):
+        r'''Gets the fraction of galaxies affected by variable depth in the line-of-sight direction for the given tomographic bin and shell.
+        
+        Parameters
+        ----------
+        index : tuple
+            Indices of the tomographic bin and shell pair.
+
+        Returns
+        -------
+        fraction_vardepth : array_like
+            Fraction of galaxies affected by variable depth in the line-of-sight direction.
+        '''
+        is_in_shell = (self.zbins[index[1]][0] < self.z) & (self.z <= self.zbins[index[1]][1])
+
+        n_gal_in_tomo_vardepth   = np.trapz(self.dndz_vardepth[index[0]][:, is_in_shell], self.z[is_in_shell])
+        n_gal_in_tomo            = np.trapz(self.dndz[index[0]][is_in_shell], self.z[is_in_shell])
+        fraction_vardepth = np.divide(n_gal_in_tomo_vardepth, n_gal_in_tomo, out=np.ones_like(n_gal_in_tomo_vardepth), where=n_gal_in_tomo!=0)
+
+        return fraction_vardepth
+
+    def __getitem__(self, index):
+        r'''Get the mask for the given index.
+
+        Parameters
+        ----------
+        index : tuple
+            Indices of the tomographic bin and shell pair.
+
+        Returns
+        -------
+        mask : array_like
+            Mask for the given index.
+
+        Raises
+        ------
+        ValueError
+            If the index is invalid.
+
+        '''
+        self.test_index(index)
+
+        if self.vardepth_tomo_functions is None:
+            angular_vardepth_map = angular_tracer_map = self.vardepth_map[index[0]]
+        else:
+            angular_vardepth_map    = self.vardepth_tomo_functions[index[0]](self.vardepth_map[0])
+            angular_tracer_map      = self.vardepth_map[0]
+
+        los_fraction_vardepth = self.get_los_fraction(index)
+
+        if self.vardepth_los_tracer is None:
+            los_vardepth_map =  np.interp(angular_tracer_map, self.vardepth_values, los_fraction_vardepth)
+        else:
+            los_vardepth_map = np.interp(self.vardepth_los_tracer, self.vardepth_values, los_fraction_vardepth)
+        
+        return np.multiply(angular_vardepth_map, los_vardepth_map)


Saransh-cpp (changes requested):
Thanks for the amazing work on this, @mwiet! This PR will require some design changes and adhering to good practices (tests, PEP8 conventions, ...), but I think we can discuss those once the functionality is ready and approved.

paddyroddy:
pre-commit.ci autofix

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        ValueError
+            If the index is invalid.
+
+        """
+        self.test_index(index)
+
+        return self.vardepth_map[index[0]]
+
+
+class angular_los_variable_depth_mask(angular_variable_depth_mask):
```
paddyroddy, line 390: Again following `N801`

```suggestion
class AngularLosVariableDepthMask(AngularVariableDepthMask):
```

glass/observations.py
```diff
@@ -22,6 +22,8 @@
 ----------
 
 .. autofunction:: vmap_galactic_ecliptic
+.. class:: angular_variable_depth_mask
+.. class:: angular_los_variable_depth_mask
```
paddyroddy, line 26: Following `N801`

```suggestion
.. class:: AngularVariableDepthMask
.. class:: AngularLosVariableDepthMask
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
     binned_nz *= nz
 
     return binned_nz  # type: ignore[no-any-return]
+
+
+class angular_variable_depth_mask:
+    r"""Variable depth mask for tomographic bins.
+
+    This class allows to create a mask with a different variable depth mask in the angular direction for each tomographic bin.
```
paddyroddy, line 338: Word wrapping for consistency

```suggestion
    This class allows to create a mask with a different variable depth mask in the
    angular direction for each tomographic bin.
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+    n_bins : int
+        Number of tomographic bins.
+    zbins : array_like
+        Shell redshift limits.
+    """
+
+    def __init__(self, vardepth_map, n_bins, zbins):
+        self.vardepth_map = vardepth_map
+        self.n_bins = n_bins
+        self.zbins = zbins
```
paddyroddy, line 353: The actual types I'll leave to you, but can you change this to be properly typed rather than specifying in the docstring? Have a look around the code base and you'll see this is used consistently throughout.

```suggestion
    vardepth_map
        Map of variable which traces the depth per tomographic bin.
    n_bins
        Number of tomographic bins.
    zbins
        Shell redshift limits.
    """

    def __init__(
    	self,
    	vardepth_map: NDArray[np.float_],
    	n_bins: int,
    	zbins: NDArray[np.float_]
  	):
        self.vardepth_map = vardepth_map
        self.n_bins = n_bins
        self.zbins = zbins
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+
+    def test_index(self, index):
```
paddyroddy, line 355: Is this a test or validation? I don't think we should call it `test_` as that it implies it is an actual test (that you'll find in the `tests` folder)

paddyroddy, line 355: Adding typing would be great

```suggestion
    def test_index(self, index) -> None:
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        r"""Test the index for validity."""
+        if not isinstance(index, tuple):
+            raise ValueError("Index must be an tuple of two integers")
```
paddyroddy, line 358: This is really a type error

```suggestion
            raise TypeError("Index must be an tuple of two integers")
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+
+        Returns
+        -------
+        mask : array_like
+            Mask for the given index.
+
+        Raises
+        ------
+        ValueError
+            If the index is invalid.
```
paddyroddy, line 382: Docstring consistency

```suggestion
        index
            Indices of the tomographic bin and shell pair.

        Returns
        -------
            Mask for the given index.

        Raises
        ------
        ValueError
            If the index is invalid.
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        """
+        self.test_index(index)
+
+        return self.vardepth_map[index[0]]
+
+
+class angular_los_variable_depth_mask(angular_variable_depth_mask):
+    r"""Variable depth mask for tomographic bins.
+
+    This class allows to create a mask with a different variable depth mask in both the angular and the line-of-sight directions for each tomographic bin.
```
paddyroddy, line 393:
```suggestion
    This class allows to create a mask with a different variable depth mask in
    both the angular and the line-of-sight directions for each tomographic bin.
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        Map of the variable depth tracer for line-of-sight direction. If provided, it is assumed to cover the same domain as vardepth_values.
+    vardepth_tomo_functions : array_like, optional
+        List of functions which map the input vardepth_map to the ratio between the galaxy count due to the variable depth and the galaxy count without variable depth (for each tomographic bin). If provided, it is assumed that there is one vardepth_map which traces the variable depth for all tomographic bins.
+
+    References
+    ----------
+    .. [1] Joachimi B., Lin, C.-A., et al., 2021, A&A, 646, A129.
+           doi:10.1051/0004-6361/202038831
+    .. [2] von Wietesheim-Kramsta M., Lin, K., et al., 2024, A&A, 695, A223.
+           doi:10.1051/0004-6361/202450487
```
paddyroddy, line 423: Can you move these to https://github.com/glass-dev/glass/blob/main/docs/user/bibliography.rst? Also give them citation like `[Joachimi2021]`, `[Wietesheim-Kramsta2024]`

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        vardepth_map,
+        n_bins,
+        zbins,
+        ztomo,
+        dndz,
+        z,
+        dndz_vardepth,
+        vardepth_values,
+        vardepth_los_tracer=None,
+        vardepth_tomo_functions=None,
```
paddyroddy, line 437: You can change the actual types, but can you again move them out of the docstring to here?

```suggestion
        self,
        vardepth_map: NDArray[np.float_],
        n_bins: int,
        zbins: NDArray[np.float_],
        ztomo: NDArray[np.float_] | None,
        dndz: NDArray[np.float_] | None,
        z: NDArray[np.float_] | None,
        dndz_vardepth: NDArray[np.float_],
        vardepth_values: NDArray[np.float_],
        vardepth_los_tracer: NDArray[np.float_] | None=None,
        vardepth_tomo_functions: NDArray[np.float_] | None=None,
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+
+        Parameters
+        ----------
+        index : tuple
+            Indices of the tomographic bin and shell pair.
+
+        Returns
+        -------
+        fraction_vardepth : array_like
+            Fraction of galaxies affected by variable depth in the line-of-sight direction.
```
paddyroddy, line 462:
```suggestion
        index
            Indices of the tomographic bin and shell pair.

        Returns
        -------
            Fraction of galaxies affected by variable depth in the
            line-of-sight direction.
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        self.dndz_vardepth = dndz_vardepth
+        self.vardepth_values = vardepth_values
+        self.vardepth_los_tracer = vardepth_los_tracer
+        self.vardepth_tomo_functions = vardepth_tomo_functions
+
+        if vardepth_tomo_functions is not None:
+            self.vardepth_map = np.atleast_2d(vardepth_map).reshape(1, -1)
+
+    def get_los_fraction(self, index):
+        r"""Gets the fraction of galaxies affected by variable depth in the line-of-sight direction for the given tomographic bin and shell.
```
paddyroddy, line 452:
```suggestion
        r"""Gets the fraction of galaxies affected by variable depth in the
        line-of-sight direction for the given tomographic bin and shell.
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+        Returns
+        -------
+        fraction_vardepth : array_like
+            Fraction of galaxies affected by variable depth in the line-of-sight direction.
+        """
+        is_in_shell = (self.zbins[index[1]][0] < self.z) & (
+            self.z <= self.zbins[index[1]][1]
+        )
+
+        n_gal_in_tomo_vardepth = np.trapz(
```
paddyroddy, line 468: We are now using NumPy v2, so `np.trapz` is deprecated, please change to 

```suggestion
        n_gal_in_tomo_vardepth = np.trapezoid(
```

glass/observations.py
```diff
@@ -328,3 +330,192 @@ def tomo_nz_gausserr(
...
+
+        Parameters
+        ----------
+        index : tuple
+            Indices of the tomographic bin and shell pair.
+
+        Returns
+        -------
+        mask : array_like
+            Mask for the given index.
```
paddyroddy, line 492:
```suggestion
        index
            Indices of the tomographic bin and shell pair.

        Returns
        -------
            Mask for the given index.
```

paddyroddy (changes requested):
Sorry quite a lot of changes, but we can iterate.

As a minimum we need all CI to pass. Have a look at [pre-commit](https://pre-commit.com/#install). That is currently failing. You can run `pre-commit run -a` locally.

Also as @Saransh-cpp says we really need tests to go with this work.

glass/observations.py
```diff
@@ -328,3 +330,200 @@ def tomo_nz_gausserr(
...
+        vardepth_map: NDArray[np.float64],
+        n_bins: int,
+        zbins: list(tuple[float, float]),
+        ztomo: list(tuple[float, float]),
+        dndz: NDArray[np.float64],
+        z: NDArray[np.float64],
+        dndz_vardepth: NDArray[np.float64],
+        vardepth_values: NDArray[np.float64],
+        vardepth_los_tracer: NDArray[np.float64] | None = None,
+        vardepth_tomo_functions: list[callable] | None = None,
```
paddyroddy, line 443: You will need to add `from typing import Callable` at the top

```suggestion
        vardepth_tomo_functions: list[Callable] | None = None,
```

glass/observations.py
```diff
@@ -328,3 +330,200 @@ def tomo_nz_gausserr(
...
+        -------
+        fraction_vardepth
+            Fraction of galaxies affected by variable depth in the
+            line-of-sight direction.
+        """
+        is_in_shell = (self.zbins[index[1]][0] < self.z) & (
+            self.z <= self.zbins[index[1]][1]
+        )
+
+        n_gal_in_tomo_vardepth = np.trapzoid(
```
paddyroddy, line 476: Typo
```suggestion
        n_gal_in_tomo_vardepth = np.trapezoid(
```

glass/observations.py
```diff
@@ -328,3 +330,200 @@ def tomo_nz_gausserr(
...
+            self.dndz_vardepth[index[0]][:, is_in_shell], self.z[is_in_shell]
+        )
+        n_gal_in_tomo = np.trapzoid(
```
paddyroddy, line 479: Typo
```suggestion
        n_gal_in_tomo = np.trapezoid(
```

docs/user/bibliography.rst
```diff
@@ -43,3 +46,6 @@ Bibliography
 
 .. [Xavier16] Xavier H. S., et al., 2016, MNRAS, 459, 3693.
     doi:10.1093/mnras/stw874
+
+.. [Wietesheim-Kramsta2024] von Wietesheim-Kramsta M., Lin, K., et al.,
```
paddyroddy, line 50: Whitespace
```suggestion
.. [Wietesheim-Kramsta2024] von Wietesheim-Kramsta M., Lin, K., et al.,
```

glass/observations.py
```diff
@@ -328,3 +330,200 @@ def tomo_nz_gausserr(
...
+        provided, it is assumed that there is one vardepth_map which
+        traces the variable depth for all tomographic bins.
+    """
+
+    def __init__(
+        self,
+        vardepth_map: NDArray[np.float64],
+        n_bins: int,
+        zbins: list(tuple[float, float]),
+        ztomo: list(tuple[float, float]),
```
paddyroddy, line 437: Types need square brackets
```suggestion
        zbins: list[tuple[float, float]],
        ztomo: list[tuple[float, float]],
```

glass/observations.py
```diff
@@ -328,3 +332,203 @@ def tomo_nz_gausserr(
...
+        vardepth_map: NDArray[np.float64],
+        n_bins: int,
+        zbins: list[tuple[float, float]],
+        ztomo: list[tuple[float, float]],
+        dndz: NDArray[np.float64],
+        z: NDArray[np.float64],
+        dndz_vardepth: NDArray[np.float64],
+        vardepth_values: NDArray[np.float64],
+        vardepth_los_tracer: NDArray[np.float64] | None = None,
+        vardepth_tomo_functions: list[Callable] | None = None,
```
paddyroddy, line 449: Do you know the expected signature of `vardepth_tomo_functions`? This needs to be of the form

```python
list[Callable[[<input>], <output>]]
```

glass/observations.py
```diff
@@ -328,3 +332,203 @@ def tomo_nz_gausserr(
...
+    ) -> None:
+        super().__init__(vardepth_map, n_bins, zbins)
```
paddyroddy, line 451: So [mypy](https://github.com/python/mypy) is currently complaining about `zbins`.

In `AngularLosVariableDepthMask`, `zbins` is of type `list[tuple[float, float]]`. Whereas in `AngularVariableDepthMask` `zbins` is of type `NDArray[np.float64]`. Are these both correct? If so then we can change `AngularVariableDepthMask` to be something like `list[tuple[float, float]] | NDArray[np.float64]` or `Sequence[float]` etc.

paddyroddy (commented):
I made a few small commits, we're down to two `mypy` warnings. Thanks for your work on tackling these!

glass/observations.py
mwiet: It should be `list[tuple[float, float]]` in both. I will correct that now.

paddyroddy:
pre-commit.ci autofix

paddyroddy:
Great all passing CI is now

docs/user/bibliography.rst
```diff
@@ -43,3 +46,6 @@ Bibliography
...
```
paddyroddy, line 50: Your own name typo? 😆
```suggestion
.. [Wietesheim-Kramsta2024] von Wietersheim-Kramsta M., Lin, K., et al.,
```

paddyroddy:
Have received a notebook from @mwiet. As he is busy, it was agreed with @ntessore that we'll take this off his hands and get this PR in a suitable position to be merged.

> Hi @paddyroddy, Sorry about the delay! Here’s an example notebook for a KiDS-1000-like survey with mock n(z), mock variable depth and a stage-IV-like mask. The survey properties are not all physical, but they give an idea of how they should be formatted.
> Also, if you actually want to sample the galaxy catalogue. With the current galaxy density, it might take a really long time, so I would recommend just multiplying tomo_nz by 0.01.

For now, all I've done is add the notebook as it was running @pre-commit on it once removing the output (the initial figures made it 10MB). I will make it more inline with the other notebooks and reduce the figures displayed.

docs/examples.rst
```diff
@@ -46,3 +46,4 @@ More advanced examples doing multiple things at the same time.
    examples/2-advanced/cosmic_shear.ipynb
    examples/2-advanced/stage_4_galaxies.ipynb
    examples/2-advanced/legacy-mode.ipynb
+   examples/2-advanced/variable_depth_example.ipynb
```
paddyroddy, line 49: Not sure if we want this public, was more for making sure the docs CI passed

docs/user/bibliography.rst
```diff
@@ -29,6 +29,9 @@ Bibliography
 .. [Johnston19] Johnston, H., Georgiou, C., Joachimi, B., et al., 2019,
     A&A, 624, A30. doi:10.1051/0004-6361/201834714
 
+.. [Joachimi2021] Joachimi B., Lin, C.-A., et al., 2021, A&A, 646, A129.
+    doi:10.1051/0004-6361/202038831
+
```
ntessore, line 34: Year is 4 digits here and below, but not in the others.

examples/2-advanced/variable_depth_example.ipynb
ntessore, line 1: This file should simply be called variable-depth.ipynb

ntessore (changes requested):

ntessore:
Thanks for the fantastic work @mwiet! From your side, I think we are all set, you can tick the box on the to-do list :)

For us, I believe we should align this more closely with the overall roadmap for sampling galaxies in GLASS. I see a couple of tasks before merging:

We do not need the two classes. The base class checks for a valid index before accessing an array of maps; numpy does that for us, we can therefore simply use the stack of matrices. The work that the other class does in `__getitem__()` should be a standalone function that operates not on a stack of inputs, but on individual inputs (so the selection with `index` happens before the function is called).

And, importantly, a lot of the "variable depth functionality" is actually happening in the notebook, not in the library. Creating the stack of depth maps from a tracer, for example, is something that we need to provide. Think of it this way: how would a user go about creating the inputs for variable depth? Everything they would need to copy and paste verbatim from the example notebook should be provided by GLASS, I believe.

examples/2-advanced/variable-depth.ipynb
```diff
@@ -0,0 +1,521 @@
...
+    "# creating a numpy random number generator for sampling\n",
+    "rng = np.random.default_rng(seed=42)\n",
+    "\n",
+    "# cosmology for the simulation\n",
+    "h = 0.7\n",
+    "Oc = 0.25\n",
+    "Ob = 0.05\n",
+    "\n",
+    "# basic parameters of the simulation\n",
+    "nside = lmax = 64\n",
```
paddyroddy, line 40: The resolution was initially 1024. I have had to drop it to 64 as otherwise this notebook takes a really long time. 128 would be suitable, but the pre-allocation array step becomes too big in CI (based on GH runners), so went with 64. Still gives similar results (by eye).

examples/2-advanced/variable-depth.ipynb
```diff
@@ -0,0 +1,521 @@
...
...mg2gCCaQkwhlOMTRCkc5SrjBKzziRFEvuySJovjWjNkX++KrTY3z9S22m2JoSzrUW/TTUWsGOKMGno9iCEAmmDnKxbXyg65zBOqa5H89twakwRfRsg+drXJr7lK3kLtr+tYd9v4Yqey4TYf25/lPiMHzRyktUcpDRpC1h3Z8f7tsH9lXy8wuUAQTRGnXvlG9puR1NkhuQ0vhmLjhdhre/t7+kh/BP7MT5gYqmOrXbT19HOdL7eRzmv6cq93mJ12imyQNM5uJ0EMAdAUu16xFXs+v/mVTA9Slcf5StI0BgmajJArfso9iVJZaIfEk2JzwiikVE+zXskHN5bmS9jSeU+JnSqG8Vo9kf8+W/EgjKYHCKIp4NQr31gcSzaSo3OWEIlYkxO6big0Xoi9dh2zA58gSl0qFypefPOLEUO+v/bYzVZj7bTrADEEwPiwRRFnrW1ndlKYJ8QYJsRmhFyEZIFC5qI55qgTTxJGmuflK4Nzff6G2ICbwohdAxSQPapj7f1E1q/EgyiaDiCIJhhTCBH5xdDwvFyCVrmuuFGvK4bs8ynEkB3HW1rnmRM3L91aJXlO7n7++Y1DDMXGDY3nGseVgYIQAqB9cmFkZoc4QrJDMeVxdmbGlQ3yZY1cGRGNVbd2HA1zvQHNb/qcDT0OxZU98mWrfBkzTZbK10Zr8z06z6wZSrwxrBmvTxl9+rBPSy8R6DgQRBOILYRyxi2I6mabfHsNaeKldpOzY+rGZMaIzGSFOMmNQwxpYofGc41hZ6AgggDoDst+sihIEMU4yVXiKp3jpHghmSZ7HGlMH5os1BzzGsxn/eK8dCwRK5I4Ysr4pAxPSJbIFEaatU/Vc/WtvV1LEMyYEEaTBwTRBHHalacRUfWPN6Ym1ic2dI/1jnKhpXIpxJBvfKL6Yqjaxn09JHaIGJLOa/+6QwVWaHxtvJAxIIYA6B4rrtyyci42O+TLqkiiRTZ3qJojmHF8Bgdc5kJT0ubL5riup4YTUnWyTqFrn1y24lI833h5XCleCmtvV+boiU2buXLzgjCaHCCIJgRODEnWkkRCKZsgOHyZIM1ixhAB4xMWofbamjU4PsERuxYpdNNVbVx7fq5Y8hj8fELnJ/fTxdfGCxkDYgiA7mKKolgjhZB1POX+bjEUEismjhg/QtxoMj8pqZNpkgRhSDYpJNujuc7F09zPFOcCbb3NO
```
paddyroddy, line 343: This section was initially

```python
for i in range(nbins):
    for j in range(len(zbins)):
```

and displayed every single `hp.mollview`. This took a long time and made the notebook huge, so I have just made it one.

paddyroddy, line 384: This section was very slow in the initial implementation. By pre-allocating the catalogue, it is significantly quicker.

tests/test_observations.py
```diff
@@ -136,3 +138,86 @@ def test_tomo_nz_gausserr() -> None:
     # check the shape of the output
 
     np.testing.assert_array_equal(binned_nz.shape, (len(zbins), len(z)))
+
+
+@pytest.mark.parametrize(
```
paddyroddy, line 143: I started to add tests yesterday before @ntessore's review. It feels pointless continuing these given the proposed changes.

ntessore:
This actually closes #136, not 413.

paddyroddy:
> This actually closes #136, not 413.

That makes sense, I was a bit confused

PR #537, opened 2025-02-26
gh-124: use `cosmology.api` over the `cosmology` package

paddyroddy:
# Description

This PR aims to remove the deprecated `cosmology` package in favour of the new `cosmology.api`. The API is still in its infancy, so this might not work yet. I have raised https://github.com/cosmology-api/cosmology.api/pull/115 which will be needed for the inverse comoving distance. **This has since been released**.

Do we still want `_kappa_ia_nla` (added in https://github.com/glass-dev/glass/pull/21)? It's not tested, and have had to take some educated guesses what they mean.

Functions required:
- `dc`: `HasComovingDistance.comoving_distance`
- `dc_inv`: `HasInverseComovingDistance.inv_comoving_distance` (cosmology-api/cosmology.api#115)
- `ef`: `HasHoverH0.H_over_H0`
- `gf`: `HasGrowthFactor.growth_factor`
- `h`: `HasH0.H0` ÷ `100`
- `Om`: `HasOmegaM0.Omega_m0`
- `omega_m`: `HasOmegaM0.Omega_m0`
- `rho_c0`: `HasCriticalDensity0.critical_density0`
- `rho_m_z`: `HasCriticalDensity0.critical_density0` × `HasOmegaM.Omega_m`
- `xm`: `HasTransverseComovingDistance.transverse_comoving_distance` ÷ `HasHubbleDistance.hubble_distance`

I have attempted to adapt the `MockCosmology` object from the tests as well.

Closes: #124. Having another go following https://github.com/glass-dev/glass/pull/397.

## Changelog entry

Changed: Adds dependency on @cosmology-api rather than https://pypi.org/project/cosmolopy as before

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [ ] Have you added additional tests (if required)?
- [x] Have you modified/extended the documentation (if required)?
- [x] Have you added a one-liner changelog entry above (if required)?

PR #539, opened 2025-02-27
Suppress notebook rules rather than blanket turning off `ruff`

paddyroddy:
# Description

Was fiddling with this as part of #537 and felt it made sense to separate. Rather than turning off `ruff` altogether, we can suppress things like typing/docstrings etc. The formatted notebooks are much nicer to read too.

Closes: #537

## Checks


Saransh-cpp (approved):
Thanks, @paddyroddy!

Saransh-cpp (approved):

ntessore (approved):
LGTM

Issue #558, opened 2025-03-02
Contributors not rendering correctly in README


Saransh-cpp:
### Which Section of the Documentation Needs Improving?

The all-contributors bot keeps adding contributor profiles in the same row, which makes them not render correctly in the README - https://github.com/glass-dev/glass/blob/main/README.md#getting-in-touch.

### What Can be Improved About This Section

The contributors should automatically be distributed in multiple rows.

### How to Improve This Section

I am not sure what is causing this. I have never seen this behaviour.

paddyroddy:
> I am not sure what is causing this. I have never seen this behaviour.

I suspect this was just a merge error. Easily done.

Issue #552, opened 2025-02-28
Add docs on @all-contributors


paddyroddy:
### Which Section of the Documentation Needs Improving?

Would be good to note what to do in the contributing guidelines. Would also be good to mention we should do them one at a time otherwise you get into horrible merge conflicts as I'm currently having.

### What Can be Improved About This Section

-

PR #559, opened 2025-03-03
gh-558: regenerate contributors in README

Saransh-cpp:
# Description

Regenerated the table using `yarn all-contributors generate`.

I am guessing this happened because we triggered several bot requests, and the bot added every new person in the first row and the last column of the table. Ideally, this should not happen again, give that running the CLI locally fixed it.

Closes: #558

## Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,7 @@ We also have a public [Slack workspace] for discussions about the project.
 <!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
 <!-- prettier-ignore-start -->
 <!-- markdownlint-disable -->
-<table>
-  <tbody>
+<p>
     <tr>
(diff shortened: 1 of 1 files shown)


ntessore:
Up to you. Since we anyway have to make the decision about categories when adding people, we might as well display them, as long as it doesn't become a chore or point of discussion. I would like to see the names in the list.

paddyroddy:
Okay, in that case I think we should go back to the default, and just minimise the categories we use. I will raise a PR this afternoon.

Saransh-cpp:
I think it would be better to have the list without emojis. It almost always becomes a point of discussion if emojis are included (unless we create guidelines on what emojis are allowed, and what one has to do to have a particular emoji in their table cell). We can still have the categories in the JSON file with an internal agreement, because it is anyways not displayed to the users.

Saransh-cpp:
If we are reverting to the original format, I can add categories in the docs while working on #552

Issue #546, opened 2025-02-28
Can we run `pre-commit.ci autofix` automatically on @all-contributors PRs?


paddyroddy:
### Describe the Bug

See e.g. https://github.com/glass-dev/glass/pull/545. @all-contributors raises PR, I ran `pre-commit.ci autofix` and everything worked.

Two options:
* We exclude `.all-contributorsrc` from @pre-commit
* @Saransh-cpp has some wizardry to get the CI to run automatically

### To Reproduce

```python
import glass
...
```

### Version In Use

2024.3.dev66+g4110ef9.d20250220

### Additional Context

```markdown
- Python version:
- Operating system:
```

ntessore:
let's exclude it, we don't need bots correcting bots

PR #560, opened 2025-03-03
gh-546: ignore .all-contributorsrc from prettier

Saransh-cpp:
# Description

A bit annoying that there is no other way to ignore a JSON file. This should fix the problems we've been having with pre-commit failing on all-contributor PRs.

Closes: #546

## Checks

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+.all-contributorsrc


ntessore:
AFAIK you could also exclude the file at the pre-commit level

paddyroddy:
> AFAIK you could also exclude the file at the pre-commit level

True, but someone could format it with their IDE

Issue #554, opened 2025-02-28
Can we have a list of contributors in the user-facing documentation?


paddyroddy:
> Whatever means we choose, it would be great it we could have a flat list of people in the user-facing docs, probably on the front page! 

 _Originally posted by @ntessore in [#540](https://github.com/glass-dev/glass/issues/540#issuecomment-2690055022)_

paddyroddy:
The problem is that RST isn't supported, https://github.com/all-contributors/all-contributors/issues/391. I guess we could do either:
* Convert to markdown documentation
* Have some markdown to RST integration, like an import or something. We can specify extra files https://allcontributors.org/docs/en/bot/configuration#moving-the-all-contributors-table-into-a-different-file. So we could write to an additional contributor-only markdown file that we import into RST somehow.

ntessore:
I can quickly add a sphinx extension to parse https://github.com/glass-dev/glass/blob/main/.all-contributorsrc into a list, if all else fails.

paddyroddy:
I think what would be nicer is we add to https://github.com/glass-dev/glass/blob/d30b766e0102bbebf83bb1b1b11cefd0de8ceecc/.all-contributorsrc#L6

```json
files: [
	"README.md",
	".contributors.md"
]
```

Then we import `.contributors.md` into RST somehow?

ntessore:
Yeah that's also easy enough with MyST/include/parser I think

paddyroddy:
I will look into this

PR #563, opened 2025-03-03
gh-554: move contributors to a file and show in docs

paddyroddy:
# Description

Rather than writing directly to the README I am instead writing to a file that the documentation can then display using https://myst-parser.readthedocs.io/en/latest/index.html. For now I am using a link from the README to the same markdown file to avoid duplication, but we could have both if we wanted.

Regenerated using `npx all-contributors generate`.  

Closes: #554

## Changelog entry

Added: The list of contributors to the homepage of the documentation.

## Checks


Saransh-cpp (approved):
Looks really nice in the docs, thanks!

ntessore (approved):
That's great, thanks @paddyroddy!

Issue #569, opened 2025-03-06
The large-files-added pre-commit check should exclude the examples


paddyroddy:
### Add your issue here

From Slack
> The large-files-added pre-commit check should exclude the examples, or increase the file size. 500kb is easily hit with a couple of figures

PR #570, opened 2025-03-06
gh-569: enforce large file check with exception

paddyroddy:
# Description

Currently, the large file @pre-commit check only runs when that file is first added. This changes it to always run, which could be useful if something got snuck in.

As per suggestion, I have excluded `.ipynb` in the `examples` folder. I have also excluded `.png` from `docs/_static` as we already have some that are over (narrowly).

Closes: #569

## Checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,13 @@ repos:
     rev: v5.0.0
     hooks:
       - id: check-added-large-files
+        args:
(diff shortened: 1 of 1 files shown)


Saransh-cpp (approved):
Nice, thanks @paddyroddy!

ntessore (approved):
Excellent, thanks for fixing this

Issue #572, opened 2025-03-07
Should `multalm` be private?


Saransh-cpp:
### Add your issue here

`glass.fields.multalm` is available under the `glass` namespace, but it is not displayed in documentation. I was wondering, given that it is only used internally by other functions, should it be private (and not under the `glass` namespace)?

PR #576, opened 2025-03-07
gh-572: make `multalm` private

Saransh-cpp:
# Description

`glass.fields.multalm` is not used anywhere outside of the sampling function, so it should be private. This should not require deprecation as the function was never available in the documentation.

Closes: #572 

## Checks


paddyroddy (changes requested):
Examples test is failing

paddyroddy (approved):
Examples fixed!

ntessore (approved):

paddyroddy:
Merging

Issue #575, opened 2025-03-07
Add `inv_triangle_number` back along with a helper function


Saransh-cpp:
> Maybe we should add the private function back in, then make two helper functions with clear purpose (and error msg) instead of using it directly?

_Originally posted by @ntessore in https://github.com/glass-dev/glass/pull/533#discussion_r1985333055_

PR #577, opened 2025-03-07
gh-575: bring back inv_triangle_number

Saransh-cpp:
# Description

Add `_is_inv_triangle_number` to check for triangle numbers and use it internally in `nfields_from_nspectra`.

Closes: #575

## Checks

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -51,6 +51,16 @@ except ImportError:
         return decorator
 
 
+def _is_inv_triangle_number(triangle_number: int) -> tuple[int, bool]:
+    r"""
+    The :math:`n`-th triangle number is :math:`T_n = n \, (n+1)/2`.  If
+    the argument is :math:`T_n`, then :math:`n` is returned. Otherwise,
+    a :class:`ValueError` is raised.
+    """
+    n = math.floor(math.sqrt(2 * triangle_number))
+    return n, n * (n + 1) // 2 == triangle_number
+
+
 def nfields_from_nspectra(nspectra: int) -> int:
     r"""
     Returns the number of fields for a number of spectra.
@@ -58,8 +68,8 @@ def nfields_from_nspectra(nspectra: int) -> int:
     Given the number of spectra *nspectra*, returns the number of
     fields *n* such that ``n * (n + 1) // 2 == nspectra``.
     """
-    n = math.floor(math.sqrt(2 * nspectra))
-    if n * (n + 1) // 2 != nspectra:
+    n, is_inv_triangle = _is_inv_triangle_number(nspectra)
+    if not is_inv_triangle:
         msg = f"invalid number of spectra: {nspectra}"
         raise ValueError(msg)
     return n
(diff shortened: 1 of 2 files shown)


glass/fields.py
```diff
@@ -51,15 +51,25 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
         return decorator
 
 
+def _is_inv_triangle_number(triangle_number: int) -> tuple[int, bool]:
```
Saransh-cpp, line 54: I'm not a big fan of this return signature. The signature minimises repetitions but I am open to any better implementations.

glass/fields.py
```diff
@@ -51,15 +51,25 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
 def nfields_from_nspectra(nspectra: int) -> int:
     r"""
     Returns the number of fields for a number of spectra.
 
     Given the number of spectra *nspectra*, returns the number of
     fields *n* such that ``n * (n + 1) // 2 == nspectra``.
     """
-    n = math.floor(math.sqrt(2 * nspectra))
-    if n * (n + 1) // 2 != nspectra:
+    n, is_inv_triangle = _is_inv_triangle_number(nspectra)
```
Saransh-cpp, line 71: `_is_inv_triangle_number` will similarly be used in a new function (to be added in #533) which will be specific to `alm`.

tests/test_fields.py
```diff
@@ -404,6 +404,18 @@ def test_getcl() -> None:
             np.testing.assert_array_equal(result[2:], expected)
 
 
+def test_is_inv_triangle_number():
+    for n in range(10_000):
+        new_n, _ = glass.fields._is_inv_triangle_number(n * (n + 1) // 2)
+        assert new_n == n
+
+    not_triangle_numbers = [2, 4, 5, 7, 8, 9, 11, 12, 13, 14, 16, 17, 18, 19, 20]
```
paddyroddy, line 412: Should we make this a fixture? It's featuring twice now

Saransh-cpp: Refactored it out as a fixture!

Saransh-cpp:
The docstrings are fixed now.

tests/test_fields.py
```diff
@@ -5,6 +5,11 @@
 import glass
 
 
+@pytest.fixture
```
paddyroddy, line 8: Only run once
```suggestion
@pytest.fixture(scope="session")
```

Saransh-cpp: Thanks!

paddyroddy (approved):
Looks good!

glass/fields.py
ntessore: Just have `_inv_triangle_number()` throw an exception if it's not -- both downstream functions can catch and rethrow it

Saransh-cpp: Thanks! I have updated the implementation.

paddyroddy (changes requested):
We're getting test failures now.

I have recently got numeric errors here https://github.com/astro-informatics/sleplet/actions/runs/13656527580 so I wonder if there has been a recent NumPy change or something

ntessore (approved):
👍

Issue #578, opened 2025-03-10
Fix mutable argument bug in legacy mode notebook


paddyroddy:
### Describe the Bug

Introduced in https://github.com/glass-dev/glass/pull/539

### To Reproduce

```python
import glass
...
```

### Version In Use

2025.2.dev62+gd61e5f2

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #579, opened 2025-03-10
gh-578: mutable argument should be empty list rather than `None`

paddyroddy:
# Description

The `labels` argument was initially set to `[]` which is bad practice https://docs.astral.sh/ruff/rules/mutable-argument-default. This should be set to `None` and then within the function should be set to the initial empty list.

Fixes: #578

## Changelog entry

Fixed: A mutable argument bug in the legacy mode notebook

## Checks


Saransh-cpp (approved):
Thanks, @paddyroddy! Weird that it went unnoticed. I have created #580 to look into it.

paddyroddy:
Merging this now as it is blocking PRs

Issue #580, opened 2025-03-10
Examples should run when they are modified


Saransh-cpp:
### Describe the Bug

At the moment examples don't run when a file in `examples/` folder is modified, which should not be the case.

### To Reproduce

```python

```

### Version In Use

main

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #581, opened 2025-03-10
gh-580: examples should run in the CI if they are modified

Saransh-cpp:
# Description

At the moment, the example notebooks do not run in the CI if files in the `examples/` directory are modified.

Closes: #580

## Checks

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -5,11 +5,13 @@ on:
   push:
     paths:
       - glass/**
+      - examples/**
     branches:
       - main
   pull_request:
     paths:
       - glass/**
+      - examples/**
     types:
       - opened
       - ready_for_review


ntessore:
> Good spot. I wonder if we'd always want them to run? Or is the concern is that they are slow?

We want them to run if the code changes, or if the examples themselves change, but not for maintenance changes. They are very slow indeed.

ntessore (approved):

paddyroddy:
> We want them to run if the code changes, or if the examples themselves change, but not for maintenance changes. They are very slow indeed.

I guess the thing is at the moment if a dependency changed (i.e. NumPy v1 -> v2) and nothing else, then they wouldn't run

ntessore:
> > We want them to run if the code changes, or if the examples themselves change, but not for maintenance changes. They are very slow indeed.
> 
> I guess the thing is at the moment if a dependency changed (i.e. NumPy v1 -> v2) and nothing else, then they wouldn't run

Hm, maybe we could implement a "skip examples" label for PRs? And if that's present, examples aren't run, e.g., on small infrastructure PRs?

paddyroddy:
> Or maybe a "run examples" label makes more sense? And if added, the examples are tested? But I don't know if you can trigger additional tests on a label without requiring a new push.

Yeah you can trigger events on labels without push https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#label

paddyroddy:
For now, I will merge this, but have raised #587 as a potential enhancement

Issue #583, opened 2025-03-11
`test_generate` is flaky


Saransh-cpp:
### Describe the Bug

`test_generate` has failed twice for me in https://github.com/glass-dev/glass/actions/runs/13785594514/job/38553058683?pr=577.

```py
=========================== short test summary info ============================
FAILED tests/test_fields.py::test_generate - AssertionError: 
Not equal to tolerance rtol=0.01, atol=0

Mismatched elements: 1 / 3072 (0.0326%)
Max absolute difference among violations: 5.58741532e-14
Max relative difference among violations: 0.02186255
 ACTUAL: array([0.435218, 0.728692, 3.003346, ..., 0.065085, 0.299785, 2.281008],
      shape=(3072,))
 DESIRED: array([0.435218, 0.728692, 3.003346, ..., 0.065085, 0.299785, 2.281008],
      shape=(3072,))
```

We should increase the tolerance of this test.

### To Reproduce

```python
import glass
...
```

### Version In Use

main

### Additional Context

```markdown
- Python version: 3.10
- Operating system: ubuntu-latest
```

ntessore:
Probably enough to check absolute error of 1e-5 or so — the chance of sampling very small numbers can make relative errors blow up

paddyroddy:
Sorry saw this afterwards. I've had these numerical comparison tests like https://github.com/astro-informatics/sleplet/blob/41957a9945eb4e28822943003f3fb47c0d7f3302/tests/test_translation.py#L22-L41 for over 5 years which only started failing in the last month. I think there's been a NumPy change or something perhaps.

How old is `test_generate`?

Saransh-cpp:
Quite new actually, was added in https://github.com/glass-dev/glass/commit/97477f9a3b7559f789176a5e5bcffd62517945f2#diff-a828467ae280cb3ae472f4c3fa7fafe22f9e672829114f1ecccc14435fa205f8.

PR #584, opened 2025-03-11
gh-583: fix flakiness of `test_generate`

Saransh-cpp:
# Description

Updates the tests to use absolute tolerance (`1e-5`).

Closes: #583

## Checks


Saransh-cpp:
Haven't investigated the reason, but I don't see a new NumPy release, so not sure where this is coming from.

paddyroddy (approved):

Issue #587, opened 2025-03-13
Could we have labels selectively run the longer example workflow?


paddyroddy:
Hm, maybe we could implement a "skip examples" label for PRs? And if that's present, examples aren't run, e.g., on small infrastructure PRs?

Or maybe a "run examples" label makes more sense? And if added, the examples are tested? But I don't know if you can trigger additional tests on a label without requiring a new push.

_Originally posted by @ntessore in https://github.com/glass-dev/glass/issues/581#issuecomment-2720415816_

PR #588, opened 2025-03-14
gh-587: run examples if a PR is labeled 'run-examples'

Saransh-cpp:
# Description

Run the examples when a PR is labeled `run-examples`. I think this is better than adding a label to _not_ run the examples, as it would save a lot of CI time.

Closes: #587

## Checks


Saransh-cpp (commented):
I think we might have to merge the PR to turn on the workflow.

paddyroddy:
> I think we might have to merge the PR to turn on the workflow.

Yes, we will

.github/workflows/examples.yml
```diff
@@ -17,6 +17,7 @@ on:
       - ready_for_review
       - reopened
       - synchronize
+      - labeled
```
paddyroddy, line 20: I'm pretty sure this won't work. The `paths` filter for `pull_request` would still hold.

Saransh-cpp: Oh I see. I will take a look at it.

.github/workflows/examples-on-label.yml
```diff
@@ -0,0 +1,27 @@
...
+      - labeled
+
+concurrency:
+  cancel-in-progress: true
+  group: >-
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+
+jobs:
+  test-examples:
+    if: contains(github.event.pull_request.labels.*.name, 'run-examples')
```
Saransh-cpp, line 16: The workflow runs whenever _any_ label is added on a PR (cannot control this), but the
1. job is skipped if the labels do not contain the `run-examples` label
2. job runs if the labels contain the `run-examples` label

One can add multiple labels in any order and the job will run if `run-examples` is included.

.github/actions/examples/action.yml
Saransh-cpp, line 1: Refactored the workflow out as a composite action.

Saransh-cpp (commented):
This should now be ready for a review.

.github/actions/examples/action.yml
paddyroddy: Just out of interest, why a composite action over reusable workflow?

Typically, I've stored composite actions in an external place, and within a repo just used a reusable workflow.

Genuinely don't have a preference, just curious.

.github/actions/examples/action.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: Run examples
+description: Composite workflow for running examples
+runs:
+  using: composite
+  steps:
```
paddyroddy, line 6: This is the weird thing of calling composite actions _within_ the same repo that they are constructed. You require the `checkout` step, which you don't if you call them external to the repo.

Not a big problem, just annoying, and unintuitive. I think in a reusable workflow, we wouldn't have that issue.

.github/workflows/examples-on-commit.yml
```diff
@@ -0,0 +1,39 @@
...
+
+concurrency:
+  cancel-in-progress: true
+  group: >-
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+
+jobs:
+  test-examples:
+    if: github.event.pull_request.draft == false
+    name: Test examples
```
paddyroddy, line 29: Do we want one of these files or `example-on-label` to have a different name? Feel like it could get confusing.

.github/workflows/examples-on-label.yml
paddyroddy: Considering how long the examples workflows take, this seems like a very reasonable approach!

paddyroddy (approved):
Really nice piece of work @Saransh-cpp. I think the solution we discussed works really well, and the composite actions removes some duplication.

.github/workflows/examples-on-label.yml
ntessore: Does this run when a new commit is pushed to a PR already labelled as `run-examples`?

paddyroddy: That's a good point ^. We could possible have this action run on all PRs, but it only runs the action job in the presence of the `run-examples` label

.github/workflows/test-examples-on-label.yml
```diff
@@ -0,0 +1,16 @@
...
+concurrency:
+  cancel-in-progress: true
+  group: >-
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+
+jobs:
+  test-examples-on-label:
+    if: contains(github.event.pull_request.labels.*.name, 'run-examples')
+    name: Test examples on label
+    uses: glass-dev/glass/.github/workflows/run-examples.yml@saransh/run-examples-on-label
```
Saransh-cpp, line 16: Will switch this to `@main` once the PR is approved.

.github/workflows/test-examples-on-push.yml
```diff
@@ -0,0 +1,30 @@
...
+concurrency:
+  cancel-in-progress: true
+  group: >-
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+
+jobs:
+  test-examples-on-push:
+    if: github.event.pull_request.draft == false
+    name: Test examples on push
+    uses: glass-dev/glass/.github/workflows/run-examples.yml@saransh/run-examples-on-label
```
Saransh-cpp, line 30: Ditto

.github/actions/examples/action.yml
Saransh-cpp: TIL: reusable workflows

I had never tried reusable workflows before, but I agree, it is much more intuitive, so I have switched things up.

.github/workflows/examples-on-label.yml
Saransh-cpp: Now it does!

.github/actions/examples/action.yml
paddyroddy: I don't know why the two have such different names.

I have all my composite actions defined here https://github.com/paddyroddy/.github/tree/main/actions, but within a repo I do think reusable probably makes more sense

.github/workflows/test-examples-on-label.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: Test examples on label
+
+on:
+  pull_request:
+
+concurrency:
+  cancel-in-progress: true
+  group: >-
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
paddyroddy, line 10: I might be wrong, but with reusable workflows you can specify many more things in the file itself. Could this block be moved into `.github/workflows/run-examples.yml`?

Saransh-cpp: That works! Thanks!

.github/workflows/test-examples-on-label.yml
```diff
@@ -0,0 +1,11 @@
+---
+name: Test examples
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
paddyroddy: Yeah, we want it to run on all PRs (regardless of whether it is into `main`). Given that you have to manually put the label, this makes sense.

paddyroddy (approved):
These look nice. Pleasing how simple the workflow files are now too.

.github/workflows/test-examples-on-label.yml
```diff
@@ -0,0 +1,11 @@
...
+
+jobs:
+  test-examples-on-label:
+    if: contains(github.event.pull_request.labels.*.name, 'run-examples')
+    name: On label
+    uses: glass-dev/glass/.github/workflows/run-examples.yml@saransh/run-examples-on-label
```
Saransh-cpp, line 11:
```suggestion
    uses: glass-dev/glass/.github/workflows/run-examples.yml@main
```

.github/workflows/test-examples-on-push.yml
```diff
@@ -0,0 +1,25 @@
...
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
+
+jobs:
+  test-examples-on-push:
+    if: github.event.pull_request.draft == false
+    name: On push
+    uses: glass-dev/glass/.github/workflows/run-examples.yml@saransh/run-examples-on-label
```
Saransh-cpp, line 25:
```suggestion
    uses: glass-dev/glass/.github/workflows/run-examples.yml@main
```

Issue #495, opened 2025-01-31
Let RadialWindow compute its effective redshift if not given


ntessore:
### Is Your Feature Request Related to a Problem? Please Describe

I'm always frustrated when I have to manually compute `RadialWindow.zeff` even though I am passing the redshift distribution.

### Describe the Solution You'd Like

I would like the default `zeff` to be `None`, in which case `RadialWindow` computes the mean redshift as

    zeff = np.trapezoid(za * wa, za) / np.trapezoid(wa, za) if za.size > 0 else 0.0

PR #590, opened 2025-03-21
gh-495: allow `zeff` to be computed automatically

paddyroddy:
# Description

Closes: #495

## Checks

### DIFF
--- a/glass/shells.py
+++ b/glass/shells.py
@@ -175,7 +175,26 @@ class RadialWindow:
 
     za: NDArray[np.float64]
     wa: NDArray[np.float64]
-    zeff: float = 0
+    zeff: float | None = None
+
+    def __post_init__(self) -> None:
+        """Magic method to calculate the effective redshift if not given."""
+        if self.zeff is None:
+            object.__setattr__(self, "zeff", self._calculate_zeff())
+
+    def _calculate_zeff(self) -> float:
+        """Calculate ``zeff`` if not given.
+
+        Returns
+        -------
+            The effective redshift depending on the size of ``za``.
+
+        """
+        if self.za.size > 0:
+            return np.trapezoid(self.za * self.wa, self.za) / np.trapezoid(
+                self.wa, self.za
+            )
+        return 0.0
 
 
 def tophat_windows(
--- a/glass/shells.py
+++ b/glass/shells.py
@@ -45,9 +45,10 @@ Weight functions
 
 from __future__ import annotations
 
+import dataclasses
 import itertools
 import warnings
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING
 
 import numpy as np
 
@@ -131,7 +132,8 @@ def density_weight(
     return cosmo.rho_m_z(z) * cosmo.xm(z) ** 2 / cosmo.ef(z)  # type: ignore[no-any-return]
 
 
-class RadialWindow(NamedTuple):
+@dataclasses.dataclass(frozen=True)
+class RadialWindow:
     """
     A radial window, defined by a window function.
 
@@ -151,12 +153,12 @@ class RadialWindow(NamedTuple):
         >>> w1.zeff = 0.15
         Traceback (most recent call last):
           File "<stdin>", line 1, in <module>
-        AttributeError: can't set attribute
+        FrozenInstanceError: cannot assign to field 'zeff'
 
     To create a new instance with a changed attribute value, use the
-    ``._replace`` method::
+    ``dataclasses.replace`` method::
 
-        >>> w1 = w1._replace(zeff=0.15)
+        >>> w1 = dataclasses.replace(w1, zeff=0.15)
         >>> w1
         RadialWindow(za=..., wa=..., zeff=0.15)
 
@@ -168,11 +170,6 @@ class RadialWindow(NamedTuple):
         Weight array; the values (ordinates) of the window function.
     zeff
         Effective redshift of the window.
-
-    Methods
-    -------
-    _replace
-        Create a new instance with changed attribute values.
 
     """
 
@@ -578,7 +575,9 @@ def partition_lstsq(
     dz = np.gradient(zp)
 
     # create the window function matrix
-    a = np.array([np.interp(zp, za, wa, left=0.0, right=0.0) for za, wa, _ in shells])
+    a = np.array(
+        [np.interp(zp, shell.za, shell.wa, left=0.0, right=0.0) for shell in shells]
+    )
     a /= np.trapezoid(a, zp, axis=-1)[..., None]
     a = a * dz
 
@@ -646,12 +645,12 @@ def partition_nnls(
         [
             np.interp(
                 zp,
-                za,
-                wa,
+                shell.za,
+                shell.wa,
                 left=0.0,
                 right=0.0,
             )
-            for za, wa, _ in shells
+            for shell in shells
         ],
     )
     a /= np.trapezoid(a, zp, axis=-1)[..., None]


glass/shells.py
```diff
@@ -169,16 +171,31 @@ class RadialWindow(NamedTuple):
...
 
     za: NDArray[np.float64]
     wa: NDArray[np.float64]
-    zeff: float = 0
+    zeff: float | None = dataclasses.field(default=None)
+
+    def __post_init__(self) -> None:
+        """Magic method to calculate the effective redshift if not given."""
+        if self.zeff is None:
+            object.__setattr__(self, "zeff", self._calculate_zeff())
```
paddyroddy, line 183: We have to use `object.__setattr__` here as `zeff` is immutable

paddyroddy:
Sorry, I had to then fix the `doctests`. Passing now.

paddyroddy:
Examples failing...

examples/1-basic/density.ipynb
```diff
@@ -66,7 +66,11 @@
     "shells = glass.linear_windows(zb)\n",
     "\n",
     "# compute the angular matter power spectra of the shells with CAMB\n",
-    "cls = glass.ext.camb.matter_cls(pars, lmax, shells)"
+    "cls = glass.ext.camb.matter_cls(\n",
+    "    pars,\n",
+    "    lmax,\n",
+    "    [(s.za, s.wa, s.zeff) for s in shells],\n",
```
paddyroddy, line 72: This feels like a silly hack. Is it possible to change `glass.ext.camb.matter_cls` instead?

tests/test_shells.py
```diff
@@ -306,3 +308,47 @@ def test_distance_grid(cosmo: Cosmology) -> None:
...
+        w.za = za
+
+    assert str(excinfo.value) == "cannot assign to field 'za'"
+
+    with pytest.raises(dataclasses.FrozenInstanceError) as excinfo:
+        w.wa = wa
+
+    assert str(excinfo.value) == "cannot assign to field 'wa'"
+
+    with pytest.raises(dataclasses.FrozenInstanceError) as excinfo:
```
Saransh-cpp, line 331: Does the `match` kwarg in `pytest.raises` not work for checking the error message here?

glass/shells.py
```diff
@@ -195,16 +196,31 @@ class RadialWindow(NamedTuple):
...
-    -------
-    _replace
-        Create a new instance with changed attribute values.
-
     """
 
     za: NDArray[np.float64]
     wa: NDArray[np.float64]
-    zeff: float = 0
+    zeff: float | None = dataclasses.field(default=None)
```
Saransh-cpp, line 203: Hmm, why have it `None` as default when that `None` is never used. Having the `_calculate_zeff` value as default will also get rid of a few `type: ignore` comments. This is maybe a question for @ntessore.

Saransh-cpp: Unless the `__post_init__` function is lazy and evaluates `_calculate_zeff` only when required.

glass/shells.py
```diff
@@ -195,16 +196,31 @@ class RadialWindow(NamedTuple):
...
+    def _calculate_zeff(self) -> float:
+        """Calculate ``zeff`` if not given.
+
+        Returns
+        -------
+            The effective redshift depending on the size of ``za``.
+
+        """
+        if self.za.size > 0:
+            return np.trapezoid(  # type: ignore[return-value]
```
Saransh-cpp, line 219: Are we certain that the expression will never return `ndarray`?

Saransh-cpp (commented):
Thanks, @paddyroddy! Some discussion points below -

tests/test_shells.py
paddyroddy: I was following the docs https://docs.pytest.org/en/stable/how-to/assert.html#assertions-about-expected-exceptions

glass/shells.py
paddyroddy: I guess it depends on what `za` and `wa` are expected to be? My feeling is 1-D arrays, but might be wrong

tests/test_shells.py
Saransh-cpp: Ah, their documentation is indeed confusing, but there is a subsection within that section - https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

I don't have a strong preference, but to be consistent in our tests, this -
```py
with pytest.raises(dataclasses.FrozenInstanceError) as excinfo:
  w.wa = wa

assert str(excinfo.value) == "cannot assign to field 'wa'"
```
should change into -
```py
with pytest.raises(dataclasses.FrozenInstanceError, match="cannot assign to field 'wa'"):
  w.wa = wa
```

paddyroddy: Happy to do this. I vaguely remembered this was a thing, but the docs didn't have it first, so went with the `assert`.

glass/shells.py
ntessore: I would set the default to `math.nan` so that we indeed never move out of the `float` space. In `__post_init__`, we can recompute `zeff` if it's `nan`. What do you think?

glass/shells.py
ntessore: Yes, I think these should be 1D arrays always.

ntessore:
Hm, I understand the change to dataclasses is due to the need to run the post-init. But these were NamedTuple because it's very useful to be able to do somethig like `for za, wa, zeff in shells`. I hope there is not too much code breaking because of this

glass/shells.py
paddyroddy: Yep I think this is much better

paddyroddy:
> Hm, I understand the change to dataclasses is due to the need to run the post-init. But these were NamedTuple because it's very useful to be able to do somethig like `for za, wa, zeff in shells`. I hope there is not too much code breaking because of this

Happy to try a different approach if you have one?

We could do on the following:

1. So it would be possible to do something like this

	```python
	a = np.array(
	    [
	        np.interp(
	            zp,
	            za,
	            wa,
	            left=0.0,
	            right=0.0,
	        )
	        for za, wa, _ in (dataclasses.astuple(shell) for shell in shells)
	    ]
	)
	```

2. Or change the `dataclass` to have the `__iter__` method, but this might have unintended consequences

	```python
	@dataclasses.dataclass(frozen=True)
	class RadialWindow:
	    za: NDArray[np.float64]
	    wa: NDArray[np.float64]
	    zeff: float = dataclasses.field(default=math.nan)
	
	    def __iter__(self) -> Iterator[NDArray[np.float64] | float]:
	        """Magic method to iterate over the window function."""
	        yield self.za
	        yield self.wa
	        yield self.zeff
	
	    def __post_init__(self) -> None:
	        """Magic method to calculate the effective redshift if not given."""
	        if math.isnan(self.zeff):
	            object.__setattr__(self, "zeff", self._calculate_zeff())
	
	    def _calculate_zeff(self) -> float:
	        """Calculate ``zeff`` if not given.
	
	        Returns
	        -------
	            The effective redshift depending on the size of ``za``.
	
	        """
	        if self.za.size > 0:
	            return np.trapezoid(  # type: ignore[return-value]
	                self.za * self.wa,
	                self.za,
	            ) / np.trapezoid(self.wa, self.za)
	        return 0.0
	```

What do you think @ntessore?

glass/shells.py
```diff
@@ -195,16 +197,31 @@ class RadialWindow(NamedTuple):
...
+                self.za * self.wa,
+                self.za,
+            ) / np.trapezoid(self.wa, self.za)
+        return 0.0
```
ntessore, line 224: Maybe we should return `math.nan` here instead of 0?

ntessore:
I like the `__iter__()` solution as a workaround: then we don't need to change the example. Once `glass.ext.camb` is no longer necessary (hopefully very soon!) we can remove the `__iter__()` method again.

glass/shells.py
paddyroddy: Changed

paddyroddy:
> I like the `__iter__()` solution as a workaround: then we don't need to change the example. Once `glass.ext.camb` is no longer necessary (hopefully very soon!) we can remove the `__iter__()` method again.

Okay, I like this solution. I'm not fully sure what `glass.ext.camb` is for anyway.

paddyroddy:
https://coveralls.io/api/v1/jobs down for maintenance

glass/shells.py
```diff
@@ -195,16 +197,39 @@ class RadialWindow(NamedTuple):
...
-    -------
-    _replace
-        Create a new instance with changed attribute values.
-
     """
 
     za: NDArray[np.float64]
     wa: NDArray[np.float64]
-    zeff: float = 0
+    zeff: float = dataclasses.field(default=math.nan)
```
ntessore, line 204: no need for the overhead

```suggestion
    zeff: float = math.nan
```

ntessore (approved):
LGTM, thanks for fixing this annoyance :) I'd remove the `dataclasses.field()` call which serves no purpose, but no strong opinions either way

glass/shells.py
paddyroddy: Sure, I like the explicitness but agree it is superfluous

PR #591, opened 2025-03-24
gh-486: run individual `iternorms` on ell blocks

### LINKED ISSUE #486: Refactor `glass.fields.iternorm`
### Add your issue here

The function mutates arrays in-place heavily, and it should be refactored to work with immutable arrays and data structures. Moreover, the correlations do not need to be regular and could be irregular/ragged to save memory; hence, we should investigate #382 in parallel with this issue.

Saransh-cpp:
# Description

Closes: #486

## Checks


Saransh-cpp (commented):
Still needs `ncorr` to be a list, and -

> every block will need it's own stack of ncorr[i] previous iterations
>
> But I don't think it's that bad — if you figure out the block boundaries in alm beforehand, you need to iterate over them  (say k1, k2) zipped with each individual j, and you can update the stack of y with j and z[k1 : k2]
>
> That's why we rearrange the alm — now the block between l1  and l2 is contiguous, so you can cut a slice out of z

glass/fields.py
```diff
@@ -390,43 +390,62 @@ def _generate_grf(
...
 
+        # concatenate individual updates into one update
+        s = np.concatenate([block_s for _, _, block_s in results])
+
         # scale by standard deviation of the conditional distribution
         # variance is distributed over real and imaginary part
         alm = _multalm(z, s)
 
         # add the mean of the conditional distribution
+        y = np.zeros((n * (n + 1) // 2, ncorr), dtype=np.complex128)
```
Saransh-cpp, line 437: Given that `y` is has a dimension `ncorr`, should be initialised inside the loop or should it be given the dimension `ncorrs`.

glass/fields.py
```diff
@@ -390,43 +390,62 @@ def _generate_grf(
...
+        a = np.concatenate([block_a for _, block_a, _ in results])
         for i in range(ncorr):
             alm += _multalm(y[:, i], a[:, i])
 
+        for i in range(ncorr):
+            # calculate ks
+            pass
```
Saransh-cpp, line 444: I am thinking that this should be possible.

Saransh-cpp (commented):
@ntessore could you look at this and see if this is the right direction? Thanks!

paddyroddy:
Closing this as it has become stale. Following discussions on Tuesday, we are planning (for now) to have a separate JAX and non-JAX implementation of `iternorm`.

Issue #592, opened 2025-03-25
Should we add `--doctest-plus-generate-diff` to `doctest`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

The `doctests` failed in #590 and I wasn't sure why. @dstansby pointed me towards `--doctest-plus-generate-diff` which helped fix it.

### Describe the Solution You'd Like

The command I ran locally was `pytest --strict-config --strict-markers -ra -v --doctest-plus glass  --doctest-plus-generate-diff=overwrite`. We could add

```python
session.posargs.append("--doctest-plus-generate-diff=overwrite")
```

to the noxfile

Saransh-cpp:
Looks nice to me.

PR #593, opened 2025-03-25
gh-592: use `--doctest-plus-generate-diff=overwrite`

Saransh-cpp:
# Description

Add `--doctest-plus-generate-diff=overwrite` to the `doctests` Nox session so that the outputs are fixed on the fly.

Closes: #592

## Checks


paddyroddy (approved):
Note one would still have to commit the changes

ntessore (approved):

Issue #519, opened 2025-02-14
Array API: test functions for legacy mode


Saransh-cpp:
### Add your issue here

A few functions added in #497 comply with array APi, but they are not tested. These should be tested in the CI for robustness.

PR #595, opened 2025-04-03
gh-519: test array API compliance of legacy mode functions

Saransh-cpp:
# Description

Add array API tests for legacy mode functions added in `glass.algorithm`. The functions added in other modules do not fully support the array API standard; therefore, those functions can be dealt with once we add full support.

Closes: #519

## Checks

### DIFF
--- a/glass/algorithm.py
+++ b/glass/algorithm.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 
 if TYPE_CHECKING:
+    from jaxtyping import Array
     from numpy.typing import NDArray
 
 
@@ -93,9 +94,9 @@ def nnls(
 
 
 def cov_clip(
-    cov: NDArray[np.float64],
+    cov: NDArray[np.float64] | Array,
     rtol: float | None = None,
-) -> NDArray[np.float64]:
+) -> NDArray[np.float64] | Array:
     """
     Covariance matrix from clipping non-positive eigenvalues.
 
@@ -130,16 +131,15 @@ def cov_clip(
     # put matrix back together
     # enforce symmetry
     v = xp.sqrt(w[..., None, :]) * v
-    cov_clipped: NDArray[np.float64] = xp.matmul(v, xp.matrix_transpose(v))
-    return cov_clipped
+    return xp.matmul(v, xp.matrix_transpose(v))
 
 
 def nearcorr(
-    a: NDArray[np.float64],
+    a: NDArray[np.float64] | Array,
     *,
     tol: float | None = None,
     niter: int = 100,
-) -> NDArray[np.float64]:
+) -> NDArray[np.float64] | Array:
     """
     Compute the nearest correlation matrix.
 
@@ -177,7 +177,7 @@ def nearcorr(
         tol = n * xp.finfo(a.dtype).eps
 
     # current result, flatten leading dimensions
-    y = a.reshape(-1, n, n)
+    y = xp.reshape(a, (-1, n, n))
 
     # initial correction is zero
     ds = xp.zeros_like(a)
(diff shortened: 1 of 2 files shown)


ntessore:
Thanks @Saransh-cpp! What is the issue with `array.reshape()` here?

Saransh-cpp:
Thanks, @ntessore! `array.reshape()` exists in JAX and NumPy but it is not a part of the array API (`array_api_strict` tests fail).

glass/algorithm.py
```diff
@@ -245,9 +245,9 @@ def cov_nearest(
         raise ValueError(msg)
 
     # store the normalisation of the matrix
-    norm: NDArray[np.float64] = xp.sqrt(diag)
+    norm = xp.sqrt(diag)
     norm = norm[..., None, :] * norm[..., :, None]
 
     # find nearest correlation matrix
-    corr = cov / xp.where(norm > 0, norm, 1.0)
+    corr = cov / xp.where(norm > 0, norm, xp.asarray([1.0]))
```
ntessore, line 252: Why not the scalar argument for `xp.where` here?

tests/test_algorithm.py
```diff
@@ -33,9 +44,9 @@ def test_nnls(rng: np.random.Generator) -> None:
         glass.algorithm.nnls(a.T, b)
 
 
-def test_cov_clip(rng):
+def test_cov_clip(xp: types.ModuleType, urng: UnifiedGenerator) -> None:
     # prepare a random matrix
-    m = rng.random((4, 4))
+    m = xp.asarray(urng.random((4, 4)))
```
ntessore, line 49: Is this necessary? I would have hoped that `urng` returns the correct array type

Saransh-cpp: Was about to post a comment for this. See https://github.com/glass-dev/glass/pull/610#discussion_r2075129679,

glass/algorithm.py
Saransh-cpp: All the arguments passed to `xp.where` must be of the type `array` - https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.where.html

```py
where(condition: array, x1: array, x2: array, /) → array
```

ntessore: It's fine in the latest API, and circumvents the usual issues with creating new arrays, so I say we should use the scalar.

Saransh-cpp: Ah, I didn't see `v2024.12` as JAX and NumPy were still on `v2023.12`. JAX is now on `v2024.12` - https://github.com/jax-ml/jax/commit/a2d62e2d3a332b1d67e0f4ef7a23375182f1646e and NumPy will soon be (next release) - `v2.3.0`.

ntessore: Looks like NumPy is already 2024.22 compatible in the one place `where` we need it to be 🙂

tests/test_algorithm.py
```diff
@@ -101,19 +114,19 @@ def test_cov_nearest(rng, mocker):
...
-    norm = np.outer(sq_d, sq_d)
+    sq_d = xp.sqrt(xp.linalg.diagonal(a))
+    norm = xp.linalg.outer(sq_d, sq_d)
 
     # make sure nearcorr was called with correct input
     nearcorr.assert_called_once()
     np.testing.assert_array_almost_equal_nulp(
         nearcorr.call_args_list[0].args[0],
-        np.divide(a, norm, where=(norm > 0), out=np.zeros_like(a)),
+        xp.divide(a, norm),
```
paddyroddy, line 127: Just checking why dropping `where`/`out`?

tests/test_algorithm.py
Saransh-cpp, line 49:
```suggestion
    m = urng.random((4, 4))
```

tests/test_algorithm.py
```diff
@@ -101,19 +114,19 @@ def test_cov_nearest(rng, mocker):
...
     )
 
     # cannot deal with negative variances
     with pytest.raises(ValueError, match="negative values"):
-        glass.algorithm.cov_nearest(np.diag([1, 1, -1]))
+        glass.algorithm.cov_nearest(xp.asarray([[1, 0, 0], [0, 1, 0], [0, 0, -1]]))
```
Saransh-cpp, line 132: There is no `diag` function or alternative in array API standard.

tests/test_algorithm.py
Saransh-cpp: `array_api_strict.divide` does not support those arguments. `out` is not testing anything here and the `where` argument is probably just reducing the number of computations (which is not very large in this case).

Saransh-cpp:
Merging this so that it does not go stale again!

Issue #406, opened 2024-11-11
Array API: port straightforward functions in `core.array`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

- [ ] `core.array.broadcast_first`
- [ ] `core.array.broadcast_leading_axes`
- [ ] `core.array.ndinterp`
- [ ] `core.array.trapz_product`
- [ ] `core.array.cumtrapz`

paddyroddy:
Should be ready following #268

paddyroddy:
I'm going to look into this

PR #597, opened 2025-04-04
gh-406: array API support for `glass.arraytools`

paddyroddy:
# Description

WIP

Closes: #406

## Changelog entry

Changed: Made `glass.arraytools` Array API compliant

## Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
     "cosmology>=2022.10.9",
     "healpix>=2022.11.1",
     "healpy>=1.15.0",
+    "jax>=0.5.3",
     "numpy>=2.1.1",
     "transformcl>=2022.8.9",
 ]


tests/test_arraytools.py
```diff
@@ -54,44 +59,45 @@ def test_broadcast_leading_axes() -> None:
...
 
-def test_ndinterp() -> None:
+@tests.conftest.array_api_compatible
+def test_ndinterp(xp: types.ModuleType) -> None:
     # test 1d interpolation
 
-    xp = np.array([0, 1, 2, 3, 4])
-    yp = np.array([1.1, 1.2, 1.3, 1.4, 1.5])
+    xq = xp.asarray([0, 1, 2, 3, 4])
+    yq = xp.asarray([1.1, 1.2, 1.3, 1.4, 1.5])
```
paddyroddy, line 67: Have renamed these to `xq`, `yq`

glass/arraytools.py
```diff
@@ -11,12 +11,13 @@
     from collections.abc import Sequence
     from typing import Unpack
 
+    from jax.typing import ArrayLike
```
Saransh-cpp, line 14: Update: this should be `from jaxtyping import Array` to avoid having a dependency on jax (large dependency, does not work for pre-commit.ci servers).

paddyroddy: Okay thanks!

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "cosmology.api>=0.3.2",
     "healpix>=2022.11.1",
     "healpy>=1.15.0",
+    "jaxtyping>=0.3.1",
```
paddyroddy, line 26: Consider removing this dependence

Issue #599, opened 2025-04-08
Replace all vars currently called `xp` with `xq`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

We're using `xp` for the Array API stuff as per convention (i.e. `np` -> `xp`). Rather than having the changes gradually in PRs like #597 it would make sense to do this one go before any of the Array API stuff is merged.

PR #600, opened 2025-04-08
gh-599: swap all `xp` vars to `xq` (and similar)

paddyroddy:
# Description

We will be using `xp` etc. throughout the Array API conversion. Currently, changes in this PR have been put in PRs such as #597, but it would make sense that we make these changes separately so downstream PRs only reflect actual Array API changes.

Closes: #599

## Checks


paddyroddy:
I'm going to merge this in as it will make Array API related PRs simpler

Issue #602, opened 2025-04-09
mypy in pre-commit.ci giving false positive


Saransh-cpp:
### Add your issue here

Running -
```
pre-commit run mypy -a
```
or even just -
```py
mypy -p glass
```
locally has been giving us an error for the past few weeks -
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

glass/fields.py:30: error: Unused "type: ignore" comment  [unused-ignore]
glass/fields.py:38: error: Name "deprecated" already defined (possibly by an import)  [no-redef]
Found 2 errors in 1 file (checked 16 source files)
```

But the mypy check on pre-commit.ci has been passing continuously. The most recent run - https://results.pre-commit.ci/run/github/357271805/1744101231.jAgJnQxuTrisJdHmoVyFdQ.

paddyroddy:
This has been driving me up the wall! Had it for almost 2 months I reckon.

PR #603, opened 2025-04-09
gh-602: pin python version and numpy for pre-commit

Saransh-cpp:
# Description

I finally got it. The mypy error Python 3.13, but pre-commit.ci is for some reason not using Python 3.13.

I finally get the error on CI here - https://results.pre-commit.ci/run/github/357271805/1744297435.iNZbr3uAQbKHw8dma-TZww

This PR pins the Python version and adds a soft pin to the NumPy version (recommended).

Closes: #602

## Checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -79,7 +79,7 @@ repos:
       - id: mypy
         files: ^glass/
         additional_dependencies:
-          - numpy>=2.2.3
+          - numpy~=2.2.4
   - repo: https://github.com/kynan/nbstripout
     rev: 0.8.1
     hooks:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -79,7 +82,7 @@ repos:
       - id: mypy
         files: ^glass/
         additional_dependencies:
-          - numpy>=2.2.3
+          - numpy==2.2.4
```
Saransh-cpp, line 85:
```suggestion
          - numpy~=2.2.4
```

paddyroddy (approved):
Amazing! Thanks for solving this, was driving me up the wall

Issue #402, opened 2024-11-11
Array API: port the RNG functions


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

As identified in #383. These are non-trivial as the Array API doesn't appear to plan to have a unified random number generation scheme https://github.com/data-apis/array-api/issues/431. The only effort of an underlying API has been [SPEC 7](https://github.com/data-apis/array-api/issues/431).

* Ideally, we want something that supports `numpy.random` ↔ `jax.random`
* Seems easier to port JAX to NumPy
* Best to have a NumPy interface as most users will be using CPU

ntessore:
Working on a proof-of-concept random API wrapper for JAX in https://github.com/glass-dev/jrng.

ntessore:
While not a general solution (see discussion at data-apis/array-api#874) the implementation at https://github.com/glass-dev/rng-jax seems to work well in the GLASS situation where a `rng` object is passed into Python code. So we could start working on that.

Saransh-cpp:
Looking at this now 👀

PR #610, opened 2025-04-25
gh-402: add NumPy-like interface for JAX RNGs

Saransh-cpp:
# Description

WIP

Refs: #402

## Checks

### DIFF
--- a/tests/_jax_rng.py
+++ b/tests/_jax_rng.py
@@ -1,0 +1,274 @@
+"""JAX random number generation as a NumPy generator."""
+
+import math
+from threading import Lock
+from typing import Literal, Self, TypeAlias
+
+from jax import Array
+from jax.dtypes import issubdtype, prng_key
+from jax.numpy import array, broadcast_shapes, shape, uint8
+from jax.random import (
+    beta,
+    binomial,
+    bits,
+    chisquare,
+    choice,
+    dirichlet,
+    exponential,
+    f,
+    gamma,
+    key,
+    multivariate_normal,
+    normal,
+    permutation,
+    poisson,
+    randint,
+    split,
+    uniform,
+)
+from jax.typing import ArrayLike, DTypeLike
+
+RealArray: TypeAlias = ArrayLike
+Size: TypeAlias = int | tuple[int, ...] | None
+
+
+def _s(size: Size, *bcast: ArrayLike) -> tuple[int, ...]:
+    """
+    Return a size, which can be a single int or None, as a shape, which
+    is a tuple of int.
+    """
+    if size is None:
+        if bcast:
+            return broadcast_shapes(*map(shape, bcast))
+        return ()
+    if isinstance(size, int):
+        return (size,)
+    return size
+
+
+class Generator:
+    """Wrapper class for JAX random number generation."""
+
+    __slots__ = ("key", "lock")
+    key: Array
+    lock: Lock
+
+    @classmethod
+    def from_key(cls, key: Array) -> Self:
+        """Wrap a JAX random key."""
+        if not isinstance(key, Array) or not issubdtype(key.dtype, prng_key):
+            msg = "not a random key"
+            raise ValueError(msg)
+        rng = object.__new__(cls)
+        rng.key = key
+        rng.lock = Lock()
+        return rng
+
+    def __init__(self, seed: int | ArrayLike, *, impl: str | None = None) -> None:
+        """Create a wrapper instance with a new key."""
+        self.key = key(seed, impl=impl)
+        self.lock = Lock()
+
+    @property
+    def __key(self) -> Array:
+        """Return next key for sampling while updating internal state."""
+        with self.lock:
+            self.key, key = split(self.key)
+        return key
+
+    def split(self, size: Size = None) -> Array:
+        """Split random key."""
+        shape = _s(size)
+        with self.lock:
+            keys = split(self.key, 1 + math.prod(shape))
+            self.key = keys[0]
+        return keys[1:].reshape(shape)
+
+    def spawn(self, n_children: int) -> list[Self]:
+        """Create new independent child generators."""
+        with self.lock:
+            self.key, *keys = split(self.key, num=n_children + 1)
+        return list(map(self.from_key, keys))
+
+    def integers(
+        self,
+        low: int | ArrayLike,
+        high: int | ArrayLike | None = None,
+        size: Size = None,
+        dtype: DTypeLike = int,
+        endpoint: bool = False,
+    ) -> Array:
+        """
+        Return random integers from the "discrete uniform" distribution
+        of the specified dtype.  If *high* is None (the default), then
+        results are from 0 to *low*.
+        """
+        if high is None:
+            low, high = 0, low
+        if endpoint:
+            high = high + 1
+        return randint(self.__key, _s(size), low, high, dtype)
+
+    def random(self, size: Size = None, dtype: DTypeLike = float) -> Array:
+        """Return random floats in the half-open interval [0.0, 1.0)."""
+        return uniform(self.__key, _s(size), dtype)
+
+    def choice(
+        self,
+        a: Array,
+        size: Size = None,
+        replace: bool = True,
+        p: Array | None = None,
+        axis: int = 0,
+    ) -> Array:
+        """Generate a random sample from a given array."""
+        return choice(self.__key, a, _s(size), replace, p, axis)
+
+    def bytes(self, length: int) -> bytes:
+        """Return random bytes."""
+        shape = (length // uint8.dtype.itemsize,)
+        return bits(self.__key, shape, uint8).tobytes()
+
+    def permutation(self, x: int | Array, axis: int = 0) -> Array:
+        """Randomly permute a sequence, or return a permuted range."""
+        return permutation(self.__key, x, axis, False)
+
+    def beta(self, a: RealArray, b: RealArray, size: Size = None) -> Array:
+        """Draw samples from a Beta distribution."""
+        return beta(self.__key, a, b, _s(size))
+
+    def binomial(self, n: RealArray, p: RealArray, size: Size = None) -> Array:
+        """Draw samples from a binomial distribution."""
+        return binomial(self.__key, n, p, _s(size))
+
+    def chisquare(self, df: RealArray, size: Size = None) -> Array:
+        """Draw samples from a chi-square distribution."""
+        return chisquare(self.__key, df, _s(size))
+
+    def dirichlet(self, alpha: RealArray, size: Size = None) -> Array:
+        """Draw samples from the Dirichlet distribution."""
+        return dirichlet(self.__key, alpha, _s(size))
+
+    def exponential(self, scale: RealArray = 1.0, size: Size = None) -> Array:
+        """Draw samples from an exponential distribution."""
+        return array(scale) * exponential(self.__key, _s(size, scale))
+
+    def f(self, dfnum: RealArray, dfden: RealArray, size: Size = None) -> Array:
+        """Draw samples from an F distribution."""
+        return f(self.__key, dfnum, dfden, _s(size))
+
+    def gamma(
+        self,
+        a: RealArray,
+        scale: RealArray = 1.0,
+        size: Size = None,
+    ) -> Array:
+        """Draw samples from a Gamma distribution."""
+        return array(scale) * gamma(self.__key, a, _s(size, a, scale))
+
+    # geometric(p[, size])
+    # Draw samples from the geometric distribution.
+
+    # gumbel([loc, scale, size])
+    # Draw samples from a Gumbel distribution.
+
+    # hypergeometric(ngood, nbad, nsample[, size])
+    # Draw samples from a Hypergeometric distribution.
+
+    # laplace([loc, scale, size])
+    # Draw samples from the Laplace or double exponential distribution with specified location (or mean) and scale (decay).
+
+    # logistic([loc, scale, size])
+    # Draw samples from a logistic distribution.
+
+    # lognormal([mean, sigma, size])
+    # Draw samples from a log-normal distribution.
+
+    # logseries(p[, size])
+    # Draw samples from a logarithmic series distribution.
+
+    # multinomial(n, pvals[, size])
+    # Draw samples from a multinomial distribution.
+
+    # multivariate_hypergeometric(colors, nsample)
+    # Generate variates from a multivariate hypergeometric distribution.
+
+    def multivariate_normal(
+        self,
+        mean: RealArray,
+        cov: RealArray,
+        size: Size = None,
+        *,
+        method: Literal["svd", "eigh", "cholesky"] = "svd",
+    ) -> Array:
+        """Draw random samples from a multivariate normal distribution."""
+        return multivariate_normal(
+            self.__key,
+            mean,
+            cov,
+            shape=_s(size),
+            method=method,
+        )
+
+    # negative_binomial(n, p[, size])
+    # Draw samples from a negative binomial distribution.
+
+    # noncentral_chisquare(df, nonc[, size])
+    # Draw samples from a noncentral chi-square distribution.
+
+    # noncentral_f(dfnum, dfden, nonc[, size])
+    # Draw samples from the noncentral F distribution.
+
+    def normal(
+        self, loc: float, scale: float, size: Size = None, dtype: DTypeLike = float
+    ) -> Array:
+        return loc + scale * normal(self.__key, _s(size), dtype)
+
+    # pareto(a[, size])
+    # Draw samples from a Pareto II (AKA Lomax) distribution with specified shape.
+
+    def poisson(self, lam: float, size: Size = None, dtype: DTypeLike = float) -> Array:
+        return poisson(self.__key, lam, size, dtype)
+
+    # power(a[, size])
+    # Draws samples in [0, 1] from a power distribution with positive exponent a - 1.
+
+    # rayleigh([scale, size])
+    # Draw samples from a Rayleigh distribution.
+
+    # standard_cauchy([size])
+    # Draw samples from a standard Cauchy distribution with mode = 0.
+
+    # standard_exponential([size, dtype, method, out])
+    # Draw samples from the standard exponential distribution.
+
+    # standard_gamma(shape[, size, dtype, out])
+    # Draw samples from a standard Gamma distribution.
+
+    def standard_normal(self, size: Size = None, dtype: DTypeLike = float) -> Array:
+        """Draw samples from a standard Normal distribution (mean=0, stdev=1)."""
+        return normal(self.__key, _s(size), dtype)
+
+    # standard_t(df[, size])
+    # Draw samples from a standard Student's t distribution with df degrees of freedom.
+
+    # triangular(left, mode, right[, size])
+    # Draw samples from the triangular distribution over the interval [left, right].
+
+    def uniform(
+        self, low: int = 0, high: int = 1, size: Size = None, dtype: DTypeLike = float
+    ) -> Array:
+        """Draw samples from a Uniform distribution."""
+        return uniform(self.__key, _s(size), dtype, low, high)
+
+    # vonmises(mu, kappa[, size])
+    # Draw samples from a von Mises distribution.
+
+    # wald(mean, scale[, size])
+    # Draw samples from a Wald, or inverse Gaussian, distribution.
+
+    # weibull(a[, size])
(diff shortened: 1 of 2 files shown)


tests/_jax_rng.py
Saransh-cpp, line 1: Should this be in glass or should we publish https://github.com/glass-dev/rng-jax as a stand-alone PyPI package?

tests/conftest.py
```diff
@@ -151,8 +152,10 @@ def dc_inv(self, dc: NDArray[np.float64]) -> NDArray[np.float64]:
 
 
 @pytest.fixture(scope="session")
-def rng() -> np.random.Generator:
-    return np.random.default_rng(seed=42)
+def rng() -> np.random.Generator | Generator:
+    if array_lib != jax:  # how?
+        return np.random.default_rng(seed=42)
+    return Generator(seed=42)
```
Saransh-cpp, line 158: I have tried quite a few things to make this work, but I think the only thing that can be done here is [parameterizing the tests](https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization) that use this fixture.

I'll look at it more next week, but please let me know if you have any ideas in mind.

tests/_jax_rng.py
```diff
@@ -0,0 +1,274 @@
...
+        self,
+        a: RealArray,
+        scale: RealArray = 1.0,
+        size: Size = None,
+    ) -> Array:
+        """Draw samples from a Gamma distribution."""
+        return array(scale) * gamma(self.__key, a, _s(size, a, scale))
+
+    # geometric(p[, size])
+    # Draw samples from the geometric distribution.
```
Saransh-cpp, line 170: Can be populated as and when we start using these distributions.

Saransh-cpp:
Requesting a review to start discussions/get suggestions 🙂

tests/_jax_rng.py
ntessore: This might be a good argument against making it a separate package?

glass/rng_dispatcher.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def uniform(
+        self, low: int = 0, high: int = 1, size: Size = None, dtype: DTypeLike = float
+    ) -> Array:
+        """Draw samples from a Uniform distribution."""
+        return uniform(self.__key, _s(size), dtype, low, high)
+
+
+def rng(array: NDArray[Any] | Array) -> JAXGenerator | np.random.Generator:
+    """Dispatch RNG on the basis of the provided array or backend."""
```
Saransh-cpp, line 109:
```suggestion
    """Dispatch RNG on the basis of the provided array."""
```

glass/rng_dispatcher.py
```diff
@@ -0,0 +1,132 @@
...
+        low: int = 0,
+        high: int = 1,
+        size: Size = None,
+        dtype: Shaped[Array, ...] = float,
+    ) -> Array:
+        """Draw samples from a Uniform distribution."""
+        return uniform(self.__key, _s(size), dtype, low, high)
+
+
+def rng(array: NDArray[Any] | Array) -> JAXGenerator | np.random.Generator:
```
Saransh-cpp, line 125: The dispatching mechanism requires to dispatch on the basis of something. I looked around, and every function accessing an RNG accepts 1 or more arrays as arguments. So, this looks like the best solution to me.

The other solution is to check if a user has jax imported / installed and dispatch on the basis of that.

glass/rng_dispatcher.py
```diff
@@ -0,0 +1,132 @@
...
+    """Dispatch RNG on the basis of the provided array."""
+    if array.__array_namespace__().__name__ == "jax.numpy":
+        return JAXGenerator(seed=42)
+    return np.random.default_rng()
+
+
+UnifiedGenerator: TypeAlias = np.random.Generator | JAXGenerator
```
Saransh-cpp, line 132: Should this be in `_array_api_utils.py`?

tests/conftest.py
```diff
@@ -95,8 +96,22 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
...
+]
+
 # use this as a decorator for tests involving array API compatible functions
-array_api_compatible = pytest.mark.parametrize("xp", xp_available_backends.values())
+array_api_compatible = pytest.mark.parametrize(
+    "backend",
+    # backends must be matched with their corresponding RNGs
+    [(b, r) for b, r in zip(xp_available_backends.values(), rngs, strict=True)],
+    ids=xp_available_backends.values(),
+)
```
Saransh-cpp, line 114: Check tests running in #595 to see this in action.

Saransh-cpp (commented):
@paddyroddy @ntessore this should now be ready for a review. I want to run the API design through you both before writing any tests.

The API design that I have in mind:

For users:

```py
rng = glass.rng.JAXGenerator(seed=42)
glass.something(jax_array, rng=rng)

# or

rng = np.random.default_rng(seed=42)
glass.something(numpy_array, rng=rng)
```

For internal use:

```py
def something(some_array, rng=None):
  if rng is None:
    rng = glass.rng_dispatcher.rng(some_array)
  ...
```

paddyroddy:
> @paddyroddy @ntessore this should now be ready for a review. I want to run the API design through you both before writing any tests.
> 
> The API design that I have in mind:

I think this looks nice

tests/conftest.py
```diff
@@ -95,8 +96,22 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
...
```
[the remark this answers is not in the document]
paddyroddy: Could we change `b`/`r` to something more meaningful? I know `b` is `backend`, but I'm not sure what `r` is

glass/rng.py
```diff
@@ -0,0 +1,132 @@
...
+import numpy as np
+from jax.dtypes import issubdtype, prng_key
+from jax.numpy import broadcast_shapes, shape
+from jax.random import (
+    key,
+    normal,
+    poisson,
+    split,
+    uniform,
+)
```
paddyroddy, line 18: Personally, I'd rather we `import jax.random` here. It's getting confusing with methods called `normal`, as well as direct imports called `normal`

glass/rng.py
```diff
@@ -0,0 +1,132 @@
...
+        size: Size = None,
+        dtype: Shaped[Array, ...] = float,
+    ) -> Array:
+        """Draw samples from a Uniform distribution."""
+        return uniform(self.__key, _s(size), dtype, low, high)
+
+
+def rng_dispatcher(array: NDArray[Any] | Array) -> JAXGenerator | np.random.Generator:
+    """Dispatch RNG on the basis of the provided array."""
+    if array.__array_namespace__().__name__ == "jax.numpy":
```
paddyroddy, line 127: I know we're only thinking about NumPy/JAX. But is RNG only different for JAX? I assume not given the Array API not planning full support

glass/rng.py
```diff
@@ -0,0 +1,132 @@
...
+        return JAXGenerator(seed=42)
+    return np.random.default_rng()
+
+
+UnifiedGenerator: TypeAlias = np.random.Generator | JAXGenerator
```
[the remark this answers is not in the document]
paddyroddy: Would that make sense? Given that this typing is RNG related, it feels like it belongs here

CONTRIBUTING.md
```diff
@@ -90,13 +90,19 @@ decorating it with `@array_api_compatible`. This will `parameterize` the test to
...
 from tests.conftest import array_api_compatible
 
+if TYPE_CHECKING:
+    from types import ModuleType
+    from glass.rng_dispatcher import UnifiedGenerator
+
 
 @array_api_compatible
-def test_something(xp: types.ModuleType):
+def test_something(backend: tuple[ModuleType, UnifiedGenerator]):
```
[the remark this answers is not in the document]
ntessore: I don't think that's right?

```py
import jax
import numpy as np
import pytest

@pytest.fixture(params=["numpy", "jax"])
def xp(request):
    if request.param == "numpy":
        return np
    if request.param == "jax":
        return jax.numpy
    raise ValueError(request.param)

@pytest.fixture
def rng(xp):
    if xp is np:
        return np.random.default_rng(42)
    if xp is jax.numpy:
        return jax.random.key(42)
    raise ValueError("unknown xp")

def test_abc(xp, rng):
    print(xp, rng)
```
```console
% pytest -xvv a.py
a.py::test_abc[numpy] PASSED
a.py::test_abc[jax] PASSED
```

ntessore:
> The API design that I have in mind:

I like the design. Can we stick jax-specific functionality into a `glass.jax` namespace so that the optional dependency is clearly isolated?

Saransh-cpp:
pre-commit.ci autofix

CONTRIBUTING.md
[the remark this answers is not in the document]
Saransh-cpp: Thank you! I hadn't thought of passing a fixture as an arg to another fixture. This simplifies the whole infrastructure.

Saransh-cpp (commented):
Thanks for the review, @paddyroddy @ntessore! This should be ready again. I think I've incorporated all the suggestions, but I might've missed a couple.

glass/jax.py
```diff
@@ -0,0 +1,117 @@
...
+    if size is None:
+        if bcast:
+            return jnp.broadcast_shapes(*map(jnp.shape, bcast))  # type: ignore[no-any-return]
+        return ()
+    if isinstance(size, int):
+        return (size,)
+    return size
+
+
+class JAXGenerator:
```
Saransh-cpp, line 37: Looking at this again, the class can now be renamed to `Generator` -- `glass.jax.Generator` looks quite clean.

tests/conftest.py
```diff
@@ -95,8 +105,24 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
     msg = f"unsupported array backend: {GLASS_ARRAY_BACKEND}"
     raise ValueError(msg)
 
-# use this as a decorator for tests involving array API compatible functions
-array_api_compatible = pytest.mark.parametrize("xp", xp_available_backends.values())
+
+@pytest.fixture(
+    params=xp_available_backends.values(), ids=xp_available_backends.values()
+)
+def xp(request: pytest.FixtureRequest) -> types.ModuleType:
```
Saransh-cpp, line 112: This closes #611 too. I think the issue are so intertwined with each other that it is much easier to resolve both of them in a single PR.

Saransh-cpp:
The RTD failure looks unrelated.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "healpy>=1.15.0",
     "numpy>=2.1.1",
     "transformcl>=2022.8.9",
+    "typing_extensions",
```
paddyroddy, line 28: Just wonder whether we want to pin this for consistency? Not fixed on the specific version
```suggestion
    "typing_extensions>=4.13.2",
```

tests/conftest.py
paddyroddy: I think it would be nice to have docstrings for these fixtures. So at a glance I know what each one is doing.

tests/conftest.py
```diff
@@ -95,8 +105,24 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
...
+@pytest.fixture
+def urng(xp: types.ModuleType) -> UnifiedGenerator:
+    seed = 42
+    backend = xp.__name__
+    if backend == "jax.numpy":
+        return glass.jax.JAXGenerator(seed=seed)
+    if backend in {"numpy", "array_api_strict"}:
+        return np.random.default_rng(seed=seed)
+    msg = "the array backend in not supported"
+    raise NotImplementedError(msg)
```
paddyroddy, line 125: I like this approach

glass/jax.py
paddyroddy: I think this sounds better

tests/test_jax.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+
+import jax
+import jax.numpy as jnp
+from glass._array_api_utils import JAXGenerator
```
paddyroddy, line 5: Any idea what's going on with `isort` here? I would have expected

```python
import jax
import jax.numpy as jnp
import pytest

from glass._array_api_utils import JAXGenerator
```

tests/test_jax.py
```diff
@@ -0,0 +1,89 @@
...
+    assert len(subrngs) == 3
+    for subrng, subkey in zip(subrngs, subkeys, strict=False):
+        assert isinstance(subrng, JAXGenerator)
+        assert subrng.key == subkey
+
+
+def test_random() -> None:
+    rng = JAXGenerator(42)
+    key = rng.key
+    rvs = rng.random(size=10000)
```
paddyroddy, line 51: Could we use the `_` throughout for reading ease?

```suggestion
    rvs = rng.random(size=10_000)
```

tests/test_array_api_utils.py
```diff
@@ -1,8 +1,8 @@
-import jax.numpy as jnp
 import numpy as np
 import pytest
 
 import glass._array_api_utils
+import jax.numpy as jnp
```
paddyroddy, line 5: I'm very confused what's going on with `isort` here

paddyroddy (commented):
The tests look really nice 👏

tests/test_array_api_utils.py
Saransh-cpp: `isort` thinks the `jax` import is coming from `glass.jax` (which is weird). This made me think if naming a submodule `jax` is wise, but then `jax` has a sub-package named `numpy` and they face no issues, so it should be okay.

I have explicitly added `jax` as a third-party import for `isort` and the imports look good now!

paddyroddy: Ah, yeah, that would be it. That is weird though, might be worth raising an issue on `isort`/`ruff`?

tests/conftest.py
```diff
@@ -95,8 +105,36 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
...
+def xp(request: pytest.FixtureRequest) -> types.ModuleType:
+    """
+    Fixture for array backend.
+
+    Access array library functions using `xp.` in tests.
+    """
+    return request.param
+
+
+@pytest.fixture
```
paddyroddy, line 121: I tend to always do this with fixtures
```suggestion
@pytest.fixture(scope="session")
```

paddyroddy (approved):
One small suggestion, but approving

ntessore:
Could we test using `xp` and `urng` somewhere here? I suspect it would not look great in the logs because we use the modules itself as the labels of the `xp` fixture. Might make more sense to use the keys as the labels?

Saransh-cpp:
Oops, the ids were actually not required. Everything looks nice now. See #595's [test run](https://github.com/glass-dev/glass/actions/runs/14780524414/job/41498249921?pr=595) for logs.

tests/conftest.py
Saransh-cpp: Whoops, just ran the tests locally and everything works. I'm pretty sure I saw an error yesterday, but maybe it was coming from somewhere else. Thanks for the help!

tests/conftest.py
```diff
@@ -95,8 +105,34 @@ def _import_and_add_jax(xp_available_backends: dict[str, types.ModuleType]) -> N
...
+
+    Access the relevant RNG using `urng.` in tests.
+
+    Must be used with the `xp` fixture. Use `rng` for non array API tests.
+    """
+    seed = 42
+    backend = xp.__name__
+    if backend == "jax.numpy":
+        return glass.jax.Generator(seed=seed)
+    if backend in {"numpy", "array_api_strict"}:
```
Saransh-cpp, line 132: Another mildly annoying thing I have been investigating is that we can mix NumPy and JAX arrays because they are interoperable, but array API provides no such uniform interface. According to array API, the array libraries are in no way required to be interoperable, which means that `array_api_strict` is not interoperable with other libraries (including NumPy).

For our case, we use NumPy RNGs to generate input for `array_api_strict` tests (as there is no RNG API in `array_api_strict`). Taking an example:
```py
>>> import numpy as np; import array_api_strict
>>> rng = np.random.default_rng(42)
>>> a = rng.normal(10)
>>> a = rng.normal(size=(10,))
>>> a
array([-1.03998411,  0.7504512 ,  0.94056472, -1.95103519, -1.30217951,
        0.1278404 , -0.31624259, -0.01680116, -0.85304393,  0.87939797])
>>> a.dtype
dtype('float64')
>>> array_api_strict.linalg.eigvalsh(a)
/Users/saransh/Code/UCL/glass/.env/lib/python3.13/site-packages/array_api_strict/_linalg.py:152: UserWarning: You are comparing a array_api_strict dtype against a NumPy native dtype object, but you probably don't want to do this. array_api_strict dtype objects compare unequal to their NumPy equivalents. Such cross-library comparison is not supported by the standard.
  if x.dtype not in _floating_dtypes:
Traceback (most recent call last):
  File "<python-input-11>", line 1, in <module>
    array_api_strict.linalg.eigvalsh(a)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/Users/saransh/Code/UCL/glass/.env/lib/python3.13/site-packages/array_api_strict/_flags.py", line 384, in wrapper
    return func(*args, **kwargs)
  File "/Users/saransh/Code/UCL/glass/.env/lib/python3.13/site-packages/array_api_strict/_linalg.py", line 153, in eigvalsh
    raise TypeError('Only floating-point dtypes are allowed in eigvalsh')
TypeError: Only floating-point dtypes are allowed in eigvalsh
```
which means using `xp.asarray(urng.<distribution>(...)` is an absolute requirement in every test to make the `array_api_strict` tests pass. I was hoping we can get away without using `xp.asarray` (which we can in the case of array libraries), but we can't for the case of `array_api_strict`.

paddyroddy: Hmm... So, to clarify, if we weren't testing `array_api_strict` then we wouldn't need `xp.asarray`? In theory, even if we were testing against all array libraries manually (i.e. NumPy, JAX, CuPy, dask, etc.)?

Saransh-cpp: > So, to clarify, if we weren't testing array_api_strict then we wouldn't need xp.asarray

Yes, we won't need `xp.asarray`!

In theory, the array libraries are not required to be interoperable, but in practice, all of them are built to be interoperable. But actually, the question of interoperability popped up here because we can't generate a random `array_api_strict.array` as there is no RNG for that. We can generate random JAX, NumPy, CuPy, ..., arrays, but all of them will have to go through `array_api_strict.asarray` to be compatible with `array_api_strict` functions (in the tests, as no one would actually use `array_api_strict` for their scientific code).

See https://github.com/glass-dev/glass/pull/595#discussion_r2074364379 to see `asarray` requirement in action.

Saransh-cpp: For NumPy, the RNG generates a NumPy array and we use NumPy functions on the array in the tests - same for JAX and other future backends - so no interoperability discussions pop up.

ntessore: Sounds like the actual issue is that we don't have a RNG for array-api-strict. So why don't we wrap a NumPy RNG into `array_api_stricy.asarray()` inside the `urng` fixture, same as we do for JAX?

Saransh-cpp: That sounds doable. Looking at it 👀

Saransh-cpp:
This is now blocked by #619

tests/conftest.py
Saransh-cpp: @paddyroddy and I discussed this today, and adding a wrapper class sounds like the best solution + it would be better to add that class in a separate PR (I will create a new PR once this PR is merged).

CONTRIBUTING.md
```diff
@@ -96,10 +98,12 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     import types
+    from glass._array_api_utils import UnifiedGenerator
 
 
-def test_something(xp: types.ModuleType):
+def test_something(xp: ModuleType, urng: UnifiedGenerator) -> None:
```
Saransh-cpp, line 104:
```suggestion
def test_something(xp: types.ModuleType, urng: UnifiedGenerator) -> None:
```

Saransh-cpp (commented):
@ntessore @paddyroddy this should be ready for a review now. I will create a wrapper class for `array_api_strict` in a new PR.

paddyroddy (approved):
This looks like a nice change

ntessore (approved):
Great, thanks @Saransh-cpp for seeing this through!

Issue #616, opened 2025-05-06
Use `position_weights()` in galaxy redshift example


ntessore:
### Which Section of the Documentation Needs Improving?

The galaxy redshifts example.

### What Can be Improved About This Section

The example currently uses the shortcut
```py
clustering = glass.effective_cls(spectra, ngal)
```
which works, but only because the redshift distributions are normalised to unity here, and the bias is unity as well.

### How to Improve This Section

Use the more generally correct approach
```py
clustering_weights = glass.position_weights(ngal, bias=1.0)
clustering = glass.effective_cls(spectra, clustering_weights)
```

PR #617, opened 2025-05-07
gh-616: use `position_weights()` in galaxy redshift example

paddyroddy:
# Description

Fixing a current example where it works only because the redshift distributions are normalised to unity here, and the bias is unity as well. This change should have no practical effect, but is more correct.

Closes: #616

## Checks


Saransh-cpp (approved):
Looks good, thanks @paddyroddy!

ntessore (approved):
Thanks!

paddyroddy:
Docs failing due to https://github.com/sphinx-doc/sphinx/issues/13533

Issue #611, opened 2025-04-30
Make `xp` (or `backend`) a fixture


Saransh-cpp:
### Add your issue here

Following #610 

It might be better to not parametrize the tests with `@array_api_compatible` and instead just use a parametrized fixture:

```py
@pytest.fixture(params=[(b, r) for b, r in zip(xp_available_backends.values(), rngs, strict=True)], ids=xp_available_backends.values())
def backends(request):
  return request.param
```

This way we will not have to decorate every single function, but instead, just use `backends` as an argument.

```py
def test_something(backends):
  xp, rng = backends
```

paddyroddy:
I prefer the idea of this solution

PR #619, opened 2025-05-07
gh-611: xp fixture for array API tests

Saransh-cpp:
# Description

Remove `@array_api_compatible` decorator and introduce `xp` fixture.

Closes: #611

## Checks

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,16 +85,19 @@ GLASS_ARRAY_BACKEND=jax python -m pytest
 GLASS_ARRAY_BACKEND=all python -m pytest
 ```
 
-Moreover, one can mark a test to be compatible with the array API standard by
-decorating it with `@array_api_compatible`. This will `parameterize` the test to
-run on every array library specified through `GLASS_ARRAY_BACKEND` -
+Moreover, one can test if a function is compatible with the array API standard
+by using the `xp` fixture. This will `parametrize` the test to run on every
+array library specified through `GLASS_ARRAY_BACKEND` -
 
 ```python
-import types
-from tests.conftest import array_api_compatible
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import types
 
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -85,16 +85,19 @@ GLASS_ARRAY_BACKEND=jax python -m pytest
...
-decorating it with `@array_api_compatible`. This will `parameterize` the test to
-run on every array library specified through `GLASS_ARRAY_BACKEND` -
+Moreover, one can test if a function is compatible with the array API standard
+by using the `xp` fixture. This will `parametrize` the test to run on every
+array library specified through `GLASS_ARRAY_BACKEND` -
 
 ```python
-import types
-from tests.conftest import array_api_compatible
```
paddyroddy, line 93: Why do we need this?

paddyroddy (approved):
Nice, this looks really useful

CONTRIBUTING.md
Saransh-cpp: The import is required for the `TYPE_CHECKING` block. Removing the import gives -

```py
tests/conftest.py:16:12: TC004 Move import `types` out of type-checking block. Import is used for more than type hinting.
   |
15 | if TYPE_CHECKING:
16 |     import types
   |            ^^^^^ TC004
17 |
18 | from numpy.typing import NDArray
   |
   = help: Move out of type-checking block
```

See https://github.com/astral-sh/ruff/issues/7214 for more details.

Saransh-cpp: Ah, sorry, a bit confusing, I meant that `from __future__ import annotations` is required for the `TYPE_CHECKING` block to work. I get an error from ruff (attached above) if I remove `from __future__ import annotations`.

paddyroddy: Okay, understood. Just wonder if we could simplify the contributing guidelines with respect to typing, but that does kind of go against the point.

ntessore (approved):
I think this is a nice simplification 🎉

Issue #629, opened 2025-07-08
`array-api-strict` arrays now cannot be plugged into `np.testing` functions


Saransh-cpp:
### Describe the Bug

All the tests using `np.testing.assert_*` functions are failing with -
```py
E           TypeError: Interoperation with NumPy requires python >= 3.12. Please upgrade.
```

as the latest `array-api-strict` release has removed the interoperability support for Python < 3.12.

I digged a bit into SciPy, and they use their custom `assert_*` functions everywhere, which were recently moved to `array-api-extra` (https://github.com/data-apis/array-api-extra/issues/17). We can use these functions but they're private; hence, their API might change without any warning and break our tests again.

`array-api-strict` people are discussing about reinstating support for Python 3.10 and 3.11, and I have raised this concern here - https://github.com/data-apis/array-api-strict/issues/160.

So I think we just wait and handle this in September/October.

### To Reproduce

```python
$ pytest
```

### Expected Behaviour

Tests pass.

### Actual Behaviour

Tests fail.

### Version In Use

main

### Additional Context

```markdown
- Python version: <3.12
- Operating system: all
```

paddyroddy:
* [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) has `3.11` support until April 2026
* [SPEC0](https://scientific-python.org/specs/spec-0000) has `3.11` support until 2025 Q4

ntessore:
I still think SPEC0 is unreasonably aggressive in light of many real HPC infrastructures, so I would not be very happy about requiring 3.12 so early.

ntessore:
Yeah, it was more of a general reflection that even from December we should probably not enforce the 3.12 requirement even if both SPEC0 and array-api-strict demand it. I know people who are still on 3.9 and think it's too early to update to 3.10 😁

PR #632, opened 2025-07-08
gh-629: temporarily pin `array-api-strict`

paddyroddy:
# Description

A recent change (see #629) is causing the tests to fail. Following this suggestion idea https://github.com/data-apis/array-api-extra/pull/343#issuecomment-3020606203 and not wanting to drop `Python<3.12` this is a compromise.

## Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ examples = [
     "matplotlib",
 ]
 test = [
-    "array_api_strict<2.4",
+    "array_api_strict>=2.0,<2.4",
     "fitsio",
     "jax>=0.4.32",
     "pytest",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ examples = [
     "matplotlib",
 ]
 test = [
-    "array_api_strict>=2",
+    "array_api_strict<2.4",
     "fitsio",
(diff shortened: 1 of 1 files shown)


Saransh-cpp (approved):

pyproject.toml
```diff
@@ -60,7 +60,7 @@ examples = [
     "matplotlib",
 ]
 test = [
-    "array_api_strict>=2",
+    "array_api_strict<2.4",
```
ntessore, line 63: Do we not want to keep the `>=2` if there was a reason to have it? If not then good to go from me

ntessore (approved):

pyproject.toml
paddyroddy: If someone were to install it, then it would install `2.3.1`. @Saransh-cpp you implemented that bound. Do you feel strongly?

Saransh-cpp: Ah, I just realised why I did this. We use the `api_version` flag to adhere to a specific array API version -

https://github.com/glass-dev/glass/blob/4349efed157136a458248d6da3d64801a42a0ec6/tests/conftest.py#L72

and this flag was introduced in `2.0.0`, so yes, we should keep `>=2`.

Issue #633, opened 2025-07-16
Consider moving some `optional-dependencies` to `dependency-groups`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Only just learnt about this https://github.com/astral-sh/uv/issues/9011. Feels like some of the things (like `docs`) should definitely be in this other category.

PR #634, opened 2025-07-23
gh-633: switch to `dependency-groups`

paddyroddy:
# Description

Closes: #633

## Checks


paddyroddy:
> Agree with the changes, but since `examples` is something a user might want to install, shouldn't we leave that one in optional-dependencies perhaps?

That does make sense. I will put this one back.

Issue #405, opened 2024-11-11
Array API: port straightforward functions in `core.algorithm`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

- [ ] `core.algorithm.nnls`

PR #643, opened 2025-10-06
gh-408: porting straightforward functions in `shells`

connoraird:
# Description

Porting the straightforward functions in `fields`. 

Closes: #408 #658 #405

## Changelog entry

Added: Added array-api support to several functions in `fields.py`
Added: A class to `glass._array_api_utils` to wrap functions which don't yet exist in the array api spec or array-api-extra.

## Checks

### DIFF
--- a/glass/_array_api_utils.py
+++ b/glass/_array_api_utils.py
@@ -61,7 +61,8 @@ class Generator:
     __slots__ = ("rng",)
 
     def __init__(
-        self, seed: int | bool | NDArray[np.int_ | np.bool] | None = None
+        self,
+        seed: int | bool | NDArray[np.int_ | np.bool] | None = None,  # noqa: FBT001
     ) -> None:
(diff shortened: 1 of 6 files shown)


glass/_array_api_utils.py
```diff
@@ -101,3 +104,118 @@ def uniform(
 
 
 UnifiedGenerator: TypeAlias = np.random.Generator | glass.jax.Generator | Generator
+
+
+class GlassXPAdditions:
```
connoraird, line 109: I have begun wrapping functions which are not present in array-api-strict or array-api-extra in this class and then following the pattern we've used for rng to handle each of the three supported libraries separately. 

For array-api-strict, I follow the [pattern used by scipy](https://labs.quansight.org/blog/scipy-array-api) when handling `Compiled code`. Which is to copy to numpy, use the numpy implementation and then copy back. 

I'm not confident this is the correct approach so I want to ask for opinions (@paddyroddy) before I finish this implementation.

Issue #669, opened 2025-10-16
Add `plain` test of importing GLASS


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

As identified in #667, #614 essentially made `jax` a required dependency. This was missed because we are testing via:
- examples
- tests
- documentation

All of these have their own set of `dependency-groups`/`optional-dependencies`. Since `jax` was in all of these, it got missed that `jax` is actually required to run GLASS.

### Describe the Solution You'd Like

Create a simple `import glass` test

PR #614, opened 2025-05-05
gh-502: add array API tests and typing for `grf._transformations`

Saransh-cpp:
# Description

Add array API tests and typing for `grf._transformations`. Should be okay to merge after #610.

Fixes: #502

## Checks

- [X] Is your code passing linting?
- [X] Is your code passing tests?
- [X] Have you added additional tests (if required)?
- [ ] Have you modified/extended the documentation (if required)?
- [ ] Have you added a one-liner changelog entry above (if required)?

PR #670, opened 2025-10-16
gh-669: Add test to make sure that GLASS is tested on its own

paddyroddy:
# Description

Looks like #614 introduced `jax` as a mandatory dependency... This was a mistake. This PR doesn't solve the problem but adds a simple test to make sure that we avoid this happening in the future. `jax` was a dependency in all of our testing (tests, docs, examples) so this went missed. This PR just adds one simple check to make sure that GLASS is installed by itself.

> [!NOTE]
> Note this is intended to fail.

Closes: #669

## Checks


connoraird (approved):

ntessore:
@all-contributors please add @connoraird for code, review

Issue #676, opened 2025-10-17
I've just realised we are still using `UnifiedGenerator` in a few places but it is no longer defined


paddyroddy:
I've just realised we are still using `UnifiedGenerator` in a few places but it is no longer defined

_Originally posted by @connoraird in https://github.com/glass-dev/glass/issues/672#issuecomment-3414657497_

PR #677, opened 2025-10-17
gh-410: Port straightforward functions in `observations`

connoraird:
# Description

Ported straight forward functions in `observations`

> Some functionality previously implemented here has been split out into [PR-681](https://github.com/glass-dev/glass/pull/681) and [PR-680](https://github.com/glass-dev/glass/pull/680)

Closes: #410

## Changelog entry

Added: Ported several functions within `observations` and one in `lensing` to the array-api (#410)

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [x] Have you added additional tests (if required)?
- [ ] Have you modified/extended the documentation (if required)?
- [x] Have you added a one-liner changelog entry above (if required)?

PR #681, opened 2025-10-17
gh-676: Remove references to UnifiedGenerator

connoraird:
# Description

> this was split out from [PR-677](https://github.com/glass-dev/glass/pull/677)

Fixes: #676 

## Changelog entry

Fixed: Removed the references to UnifiedGenerator which no longer exists (https://github.com/glass-dev/glass/issues/676)

## Checks

### DIFF
--- a/tests/grf/test_transformations.py
+++ b/tests/grf/test_transformations.py
@@ -4,25 +4,21 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 
-import glass._array_api_utils as utils
 import glass.grf
 
 if TYPE_CHECKING:
     import types
 
+    from conftest import UnifiedGenerator
 
-def test_normal(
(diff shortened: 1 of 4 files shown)


tests/grf/test_transformations.py
```diff
@@ -4,21 +4,25 @@
 
 import numpy as np
 
+import glass._array_api_utils as utils
```
paddyroddy: Elsewhere we have been using `_utils`

```suggestion
import glass._array_api_utils as _utils
```

paddyroddy (changes requested):
As suggested previously, I do think there is benefit in specifying a single type from readability. We could specify the type in `conftest.py` and then use `from tests.conftest import UnifiedGenerator` in the tests

Issue #682, opened 2025-10-20
Use custom `isort` for the Array API


paddyroddy:
### Add your issue here

Makes sense to have their own section

PR #684, opened 2025-10-20
gh-682: create an `isort` section for Array API

paddyroddy:
# Description

Creates a custom section for `isort` which means we can group Array API packages together similar to the cosmology packages.

Closes: #682

## Checks


connoraird (approved):
Nice! A bit clearer.

Issue #686, opened 2025-10-20
Investigation removing upper pin on dependencies


paddyroddy:
### Add your issue here

A couple of dependencies have pins that might not need them any more

PR #689, opened 2025-10-20
gh-686: relax `array-api-strict` upper bound pin

paddyroddy:
# Description

Relaxes the upper bound pin in `array-api-strict` that was introduced in https://github.com/glass-dev/glass/pull/632. [2.4.1 allowed the use of 3.10 again](https://github.com/data-apis/array-api-strict/blob/main/docs/changelog.md#major-changes) so we can reinstate `np.testing`. Further, the updates didn't allow `pytest.approx` and changed a few other things.

Fixes: #686

## Changelog entry

Removed: The upper bound pin of `array-api-strict`

## Checks

### DIFF
--- a/tests/grf/test_solver.py
+++ b/tests/grf/test_solver.py
@@ -89,5 +89,9 @@ def test_monopole(cl, rng):
     gl, cl_out, _ = glass.grf.solve(cl, t, monopole=gl0, gltol=1e-8)
 
     assert gl[0] == gl0
-    with pytest.raises(AssertionError):
-        np.testing.assert_allclose(cl_out[0], cl[0])
+    np.testing.assert_raises(
+        AssertionError,
+        np.testing.assert_allclose,
+        cl_out[0],
+        cl[0],
+    )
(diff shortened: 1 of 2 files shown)


tests/grf/test_solver.py
```diff
@@ -84,9 +84,10 @@ def test_monopole(cl, rng):
...
     assert gl[0] != 0.0
-    assert cl_out[0] == pytest.approx(cl[0])
+    np.testing.assert_allclose(cl_out[0], cl[0])
 
     gl, cl_out, _ = glass.grf.solve(cl, t, monopole=gl0, gltol=1e-8)
 
     assert gl[0] == gl0
-    assert cl_out[0] != pytest.approx(cl[0])
+    with pytest.raises(AssertionError):
+        np.testing.assert_allclose(cl_out[0], cl[0])
```
connoraird: Is this how we're supposed to do this, seems strange

paddyroddy: You mean how to do `!=`? There's no given way. Can also `assert_raises`, will change it to that

paddyroddy: See https://stackoverflow.com/a/38507093/7359333

Issue #432, opened 2024-11-17
Array API: remove array mutation from functions/write JAX version of each array mutating function


Saransh-cpp:
### Add your issue here

Conversation from Slack -

>>>>> 2 quick things from my initial experiments -
>>>>> 
>>>>> We will need custom JAX implementations for most of the functions as JAX arrays are immutable. See how -
>>>>> - s2fft (each file has a corresponding *_jax file - Matt Price worked on this), and
>>>>> - awkward (all other backends have common reducers but the JAX backend has specialized reducers - I worked on this)
>>>>> were developed independently, but they both have one common pattern, a standard implementation of a particular function and then a JAX implementation of the same function.
>>>>> 
>>>>> Given that the project's first half is more focussed on GPUs and less on AD, I experimented with CuPy, but their Array API support is still experimental (and I might have also stumbled on a bug).
>>>>> 
>>>> I thought the only place in the code where we cannot trivially get around the mutation was in iternorm() but I haven't actually gone through and checked
>>>>
>>>> Oh yes, or remove all the lines of code that are mutating arrays. Maybe I should check for all such lines using ruff.
>>>>>
>>> So the mutation in iternorm would be quite organically fixed by the "deep dive" project outside the GPU work that I was mentioning. We definitely have a path forward there. 
>>>
>> My inclination would be that we change all functions to not mutate arrays and then we can use jax and everything else in one go. Is there any disadvantage of this approach?
>>
> Only in the way iternorm currently works, since it always needs to discard a large array and reallocate a fresh large array that's exactly the same. But probably negligible in the grand scheme of things, within GLASS at least (the code predates use in GLASS).

Saransh-cpp:
For array mutations, we can either rewrite functions to build copy of the original arrays, or we can write a separate implementation for `jax` (using the [`at` syntax](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.ndarray.at.html)).

The `at` syntax produces copies too, _**but**_ it applies changes in-place whenever the function is JIT compiled, which is a really nice performance advantage.

> None of the x.at expressions modify the original x; instead they return a modified copy of x. However, inside a [jit()](https://docs.jax.dev/en/latest/_autosummary/jax.jit.html#jax.jit) compiled function, expressions like x = x.at[idx].set(y) are guaranteed to be applied in-place.

I think having a separate JAX implementation makes the most sense here to leverage better memory management whenever functions are JIT compiled.

PR #699, opened 2025-10-22
gh-432: Removing array mutation from ported functions

connoraird:
# Description

Closes: #432

## Checks

### DIFF
--- a/glass/algorithm.py
+++ b/glass/algorithm.py
@@ -6,8 +6,6 @@ from typing import TYPE_CHECKING
 
 import array_api_compat
 
-import glass._array_api_utils as _utils
-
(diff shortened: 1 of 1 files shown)


connoraird:
Using the benchmark I've added here, it seems that performance has seriously degraded for numpy and array-api-strict. Therefore, we will want to either rethink the algorithm or place these changes in a separate function which is called if jax is used.

paddyroddy:
> Using the benchmark I've added here, it seems that performance has seriously degraded for numpy and array-api-strict. Therefore, we will want to either rethink the algorithm or place these changes in a separate function which is called if jax is used.

I _think_ we discussed this before. Would assume the latter @ntessore?

Issue #714, opened 2025-10-24
Currently require `jax` and `array-api-strict` to run tests always


paddyroddy:
### Describe the Bug

We are installing these two within the `noxfile` which makes sense. However, anyone running `pytest` locally has to have both. This isn't a big issue, but it's an annoying mismatch. I shall look into this.

### To Reproduce

```python
import glass
...
```

### Version In Use

2025.2

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #715, opened 2025-10-24
gh-714: ensure `pytest` and `nox` agree on test dependencies

paddyroddy:
# Description

This PR addressed the mismatch between `pytest` and `nox`. `nox` explicitly installs the given dependency (`jax`, `array-api-strict`), but if you were running from `pytest` you would need all backends even if only testing on `numpy`.

The `scipy` dependency has been introduced due to following through dependencies. ~So would suggest we don't merge this until https://github.com/ntessore/flt/issues/52 has been fixed.~

Fixes: #714

## Changelog entry

Fixed: `pytest` can now be run without having all backends installed

## Checks

### DIFF
--- a/docs/user/definitions.rst
+++ b/docs/user/definitions.rst
@@ -7,6 +7,7 @@ The *GLASS* code uses the following mathematical definitions.
 .. glossary::
 
    deflection
+   displacement
       The deflection :math:`\alpha` is a complex value with spin weight
       :math:`1`.  It describes the displacement of a position along a geodesic
       (i.e. great circle).  The angular distance of the displacement is the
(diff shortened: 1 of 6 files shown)


connoraird (commented):
Does this now mean certain tests will fail silently (or be skipped) if we don't have jax/array-api-strict installed?

paddyroddy:
> Does this now mean certain tests will fail silently (or be skipped) if we don't have jax/array-api-strict installed?

Locally (if using `pytest`), yes, but never in CI. The `nox` ensures the backends are always installed.

paddyroddy:
> Just tested locally, you still get a nice skipped message `SKIPPED (test requires array_api_strict)`

Yes, exactly. I feel like we either remove the `session.install` steps from `noxfile` or we do this. I like this approach in case someone is working on a system where they can't install `jax` but aren't blocked from testing. CI will still check for them.

Issue #732, opened 2025-10-30
Consistent use of new array types


paddyroddy:
### Add your issue here

We now have a series of custom types specified in `glass._types`. We will eventually move to them with the work in #67, but I wonder if it would be easier to move a bunch now. We can then iterate on this.

PR #733, opened 2025-10-30
gh-732: use the new array types from `glass.types`

paddyroddy:
# Description

As part of #67 we are gradually using the types created in #729. To ensure that we don't forget some and to simplify future PRs, this PR uses these types everywhere. This has the effect that we are very rarely import `NDArray`. This change should ensure that switching to `array-api-typing` is easier once it becomes available.

Fixes: #732

## Changelog entry

Changed: To Array API compatible typing so that users know to use any array library

## Checks

### DIFF
--- a/glass/points.py
+++ b/glass/points.py
@@ -51,8 +51,15 @@ import glass.arraytools
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Generator
+    from types import ModuleType
 
-    from glass._types import ComplexArray, DoubleArray, FloatArray, IntArray
+    from glass._types import (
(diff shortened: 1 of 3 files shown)


glass/user.py
```diff
@@ -60,7 +60,7 @@ def save_cls(
 
 def load_cls(
     filename: os.PathLike[str],
-) -> list[NDArray[np.float64] | Sequence[float]]:
+) -> Sequence[FloatArray | Sequence[float]]:
```
paddyroddy: This fixed `mypy`

Issue #739, opened 2025-10-30
Don't use `numpy` alias when only used for array backend


paddyroddy:
### Add your issue here

Sometimes we use `numpy` to set the `xp`. It doesn't make sense to use the alias `np` in this case.

PR #740, opened 2025-10-30
gh-739: don't use numpy alias when interacting with backend

paddyroddy:
# Description

Removing cases of `import numpy as np` when only using to set, i.e. `xp = np`.

Closes: #739

## Checks


glass/_array_api_utils.py
```diff
@@ -131,13 +131,13 @@ def __init__(
...
-        import numpy as np  # noqa: PLC0415
+        import numpy  # noqa: ICN001, PLC0415
 
         import array_api_strict  # noqa: PLC0415
 
         self.axp = array_api_strict
-        self.nxp = np
-        self.rng = np.random.default_rng(seed=seed)
+        self.nxp = numpy
+        self.rng = self.nxp.random.default_rng(seed=seed)
```
paddyroddy: `numpy.random` looks weird, thought this might read better?

glass/shells.py
```diff
@@ -852,7 +850,9 @@ def _uniform_grid(
 
     """
     if xp is None:
-        xp = np
+        import numpy  # noqa: ICN001, PLC0415
+
+        xp = numpy
```
connoraird: I know I added this but I've started now to not default to anything in other functions and instead fail if we don't provide a value for xp. what would we rather do?

paddyroddy: I think we should fail

Issue #747, opened 2025-10-31
nox should be pinned to Python 3.13 for all workflows


paddyroddy:
### Describe the Bug

We just hit this bug https://github.com/healpy/healpy/issues/1002 in #746. I think GitHub Actions have recently updated to Python 3.14 which `healpy` doesn't currently support. We should pin to Python 3.13.

### To Reproduce

```python
import glass
...
```

### Version In Use

2025.3.dev29+gdb06ad960.d20251031

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #748, opened 2025-11-01
gh-747: Force nox to default to 3.13

paddyroddy:
# Description

`healpy` doesn't currently support Python 3.14. A recent GitHub Actions update means that Python defaults to 3.14. This PR ensures that we are always working in the latest Python version that we are supporting.

Fixes: #747

## Changelog entry

Fixed: `nox` now runs in the latest supported Python version only

## Checks

### DIFF
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           cache-dependency-path: pyproject.toml
           cache: pip
-          python-version: 3.x
+          python-version: 3.13
 
       - name: Cache nox
         uses: actions/cache@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/test-examples-on-label.yml
```diff
@@ -14,4 +14,4 @@ jobs:
   test-examples-on-label:
     if: contains(github.event.pull_request.labels.*.name, 'run-examples')
     name: On label
-    uses: glass-dev/glass/.github/workflows/run-examples.yml@main
+    uses: ./.github/workflows/run-examples.yml
```
paddyroddy: This should have always been like this. Ran into issues where the updated workflow wasn't being run...

Issue #720, opened 2025-10-27
glass.regularized_spectra(..., method="nearest") does not warn when it doesn't converge


ntessore:
### Describe the Bug

When `glass.regularized_spectra(..., method="nearest")` does not converge in `niter` iterations, it silently returns a result that is potentially still indefinite.

We should add a warning if that happens, with advice for how to fix the problem.

For example, in `nearcorr()`, we could add:

```py
def nearcorr():
    ...
    # find the nearest correlation matrix
    for _ in range(niter):
        ...
    else:
        # nearest correlation matrix was not found
        warnings.warn(
            f"Nearest correlation matrix not found in {niter} iterations. "
            "The result may be invalid. Please run with a larger `niter` value, "
            "or run the function again on the returned result."
        )
```

### To Reproduce

```python
import glass
...
```

### Version In Use

latest

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #757, opened 2025-11-04
gh-720: add warning if `nearcorr` has no convergence

paddyroddy:
# Description

`nearcorr` would return a potentially indefinite array with no warning. This fixes that by adding a warning to the user.

Fixes: #720

## Changelog entry

Fixed: `nearcorr` will now communicate to the user if there are insufficient iterations

## Checks


connoraird (approved):
LGTM

Issue #756, opened 2025-11-04
Run regression tests for PRs


connoraird:
### Is Your Feature Request Related to a Problem? Please Describe

To ensure we are not causing performance regressions as we port to the array-api, we should run regression tests for each PR which alters relevant sections of the repo (src, dependency versions, etc).

### Describe the Solution You'd Like

A benchmarks repo ([glass-dev/glass-benchmarks](github.com/glass-dev/glass-benchmarks)) has been created for this repo as part of #709. 

- [ ] There should be a workflow which is triggered if relevant files are updated (src, dependencies, etc)
- [ ] This workflow would then trigger a run of the regression tests workflow in `glass-dev/glass-benchmarks`.
- [ ] We should not restrict merging in the event a regression test fails as there is likely to be some false positives due to the variance in the machine we are running these benchmarks on

PR #765, opened 2025-11-06
gh-756: Add workflow to trigger reusable workflow from benchmarks repo

connoraird:
# Description

Adds a workflow which runs a reusable workflow from the benchmarks repo

Closes: #756 

Depends on https://github.com/glass-dev/glass-benchmarks/pull/13

## Changelog entry

Added: Performance regression tests in github actions

## Checks

### DIFF
--- a/.github/workflows/regression-test.yml
+++ b/.github/workflows/regression-test.yml
@@ -15,5 +15,5 @@ jobs:
   regression-tests:
     uses: glass-dev/glass-benchmarks/.github/workflows/regression-test.yaml@v0
     with:
-      starting_revision: ${{github.event.pull_request.base.ref}}
-      end_revision: ${{github.event.pull_request.head.ref}}
+      starting_revision: ${{ github.event.pull_request.base.ref }}
+      end_revision: ${{ github.event.pull_request.head.ref }}


connoraird:
I don't think a reusable workflow is what we need as this will assume we want to use the src and tests from `glass-dev/glass` but we need to use that of the benchmarks repo. We need a workflow that triggers a github action in the benchmarks repo.

connoraird:
We might be abler to do something like this https://medium.com/hostspaceng/triggering-workflows-in-another-repository-with-github-actions-4f581f8e0ceb

.github/workflows/regression-test.yml
```diff
@@ -0,0 +1,19 @@
...
+      - ready_for_review
+      - reopened
+      - synchronize
+
+jobs:
+  regression-tests:
+    uses: glass-dev/glass-benchmarks/.github/workflows/regression-test.yaml@v0
+    with:
+      starting_revision: ${{github.event.pull_request.base.ref}}
+      end_revision: ${{github.event.pull_request.head.ref}}
```
paddyroddy: These are normally formatted like this

```suggestion
      starting_revision: ${{ github.event.pull_request.base.ref }}
      end_revision: ${{ github.event.pull_request.head.ref }}
```

paddyroddy (approved):
Wonderful

Issue #770, opened 2025-11-12
redshift_grid() requires backend


ntessore:
### Describe the Bug

The `glass.redshift_grid()` function seems to require an explicit backend now, which breaks an enormous amount of code. If no backend is provided, it should default to numpy arrays.

### To Reproduce

```python
import glass
...
```

### Version In Use

latest

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #771, opened 2025-11-12
gh-770: Default to numpy for redshift_grid and fixed_zbins if no xp is provided

connoraird:
# Description

Removed requirement for xp to be passed into redshift_grid and fixed_zbins. Now the default library used will be numpy.

Fixes: #770 

## Changelog entry

Fixed: redshift_grid and fixed_zbins no longer rely on xp being passed. Now defaults to numpy.

## Checks

### DIFF
--- a/tests/core/test_shells.py
+++ b/tests/core/test_shells.py
@@ -275,30 +275,18 @@ def test_redshift_grid_default_xp() -> None:
     # check num input
 
     num = 5
-    with pytest.warns(
-        UserWarning,
-        match="No array library has been provided",
-    ):
-        z = glass.redshift_grid(zmin, zmax, num=5)
+    z = glass.redshift_grid(zmin, zmax, num=5)
     assert z.size == num + 1
 
     # check dz input
 
     dz = 0.2
-    with pytest.warns(
-        UserWarning,
-        match="No array library has been provided",
-    ):
-        z = glass.redshift_grid(zmin, zmax, dz=dz)
+    z = glass.redshift_grid(zmin, zmax, dz=dz)
     assert z.size == math.ceil((zmax - zmin) / dz) + 1
 
     # check dz for spacing which results in a max value above zmax
 
-    with pytest.warns(
-        UserWarning,
-        match="No array library has been provided",
-    ):
-        z = glass.redshift_grid(zmin, zmax, dz=0.3)
+    z = glass.redshift_grid(zmin, zmax, dz=0.3)
     assert zmax < z[-1]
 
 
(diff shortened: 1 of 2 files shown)


connoraird:
> Thanks for looking into this! How about we create a `default_xp()` function that could evolve over time? (I am inspired by `np.random.default_rng()` here.)

Sounds like a great idea

glass/_array_api_utils.py
```diff
@@ -76,6 +78,11 @@ def import_numpy(backend: str) -> ModuleType:
         return numpy
 
 
+def default_xp() -> ModuleType:
+    """Returns the library backend we default to if none is specified by the user."""
+    return import_numpy()
```
paddyroddy: Would it be useful in this method to log to the user rather than above?

tests/core/test_shells.py
```diff
@@ -267,6 +267,29 @@ def test_partition(xp: types.ModuleType, uxpx: _utils.XPAdditions, method: str)
     np.testing.assert_allclose(xp.sum(part, axis=0), uxpx.trapezoid(fz, z))
 
 
+def test_redshift_grid_default_xp() -> None:
```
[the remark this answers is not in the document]
connoraird: Same point as above

glass/_array_api_utils.py
connoraird: As in, display a warning message to the user if they use a default library?

paddyroddy: Yeah

glass/_array_api_utils.py
```diff
@@ -36,15 +38,17 @@ class CompatibleBackendNotFoundError(Exception):
...
 
-    def __init__(self, missing_backend: str, users_backend: str) -> None:
+    def __init__(self, missing_backend: str, users_backend: str | None) -> None:
         self.message = (
-            f"{missing_backend} is required here as some functions required by GLASS "
-            f"are not supported by {users_backend}"
+            f"{missing_backend} is required here as "
+            "no alternative has been provided by the user."
+            if users_backend is None
+            else f"GLASS depends on functions not supported by {users_backend}"
```
paddyroddy: Do we need this as well as the logging in `default_xp`?

glass/_array_api_utils.py
```diff
@@ -76,6 +80,15 @@ def import_numpy(backend: str) -> ModuleType:
         return numpy
 
 
+def default_xp() -> ModuleType:
+    """Returns the library backend we default to if none is specified by the user."""
+    warnings.warn(
+        f"No array library has been provided for call to {inspect.stack()[1].function}",
```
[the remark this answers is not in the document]
ntessore: I think it's fine for users to use GLASS as a CPU code so I wouldn't warn on `default_xp()`, at least for the time being.

paddyroddy (approved):

PR #786, opened 2025-11-13
gh-781: Add benchmarks for glass.arraytools

connoraird:
# Description

> originally implemented in https://github.com/glass-dev/glass-benchmarks/pull/18

Add benchmark tests for `glass.arraytools`. These tests are essentially copied from tests/core with the sad paths removed and each test parameterised so we call the function being tested once per test.

Closes: #781

## Checks


paddyroddy (approved):
Welcome back tests

PR #788, opened 2025-11-13
gh-782: Add benchmarks for `glass.fields`/`glass.harmonics`

connoraird:
# Description

Add benchmark tests for `glass.fields` and `glass.harmonics` for any functions which have both been identified as worth benchmarking in #790 and do not return Generators. Generator functions will be handled in #809.

Closes: #782, #813

## Checks

### DIFF
--- a/tests/benchmarks/test_fields.py
+++ b/tests/benchmarks/test_fields.py
@@ -1,21 +1,21 @@
 from __future__ import annotations
 
-from random import randrange
+import random
 from typing import TYPE_CHECKING
-
-import numpy as np
 
 import glass.fields
 
 if TYPE_CHECKING:
     from types import ModuleType
 
+    from conftest import Compare
     from pytest_benchmark.fixture import BenchmarkFixture
 
 
 def test_getcl_lmax_0(
-    xp: ModuleType,
     benchmark: BenchmarkFixture,
+    compare: type[Compare],
+    xp: ModuleType,
 ) -> None:
     """Benchmarks for glass.fields.getcl with lmax of 0."""
     scale_factor = 1_000
@@ -26,8 +26,8 @@ def test_getcl_lmax_0(
         for j in range(i, -1, -1)
     ]
 
-    random_i = randrange(scale_factor)
-    random_j = randrange(scale_factor)
+    random_i = random.randrange(scale_factor)
+    random_j = random.randrange(scale_factor)
 
     # check slicing
     result = benchmark(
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -225,6 +225,8 @@ lint.per-file-ignores = {"__init__.py" = [
     "INP001", # implicit-namespace-package
     "PLR2004", # magic-value-comparison
     "S101", # assert
+    "S311", # Standard pseudo-random generators are not suitable for cryptographic purposes
+    "SLF001", # Private member accessed (in tests we want to access private members)
```
paddyroddy: Just use slug like elsewhere

```suggestion
    "S311", # suspicious-non-cryptographic-random-usage
    "SLF001", # private-member-access
```

tests/benchmarks/test_fields.py
```diff
@@ -0,0 +1,74 @@
...
+    """Benchmarks for glass.fields.getcl with lmax of 0."""
+    range_size = benchmark_scale_factor
+    # make a mock Cls array with the index pairs as entries
+    cls = [
+        xp.asarray([i, j], dtype=xp.float64)
+        for i in range(range_size)
+        for j in range(i, -1, -1)
+    ]
+
+    random_i = randrange(range_size)
```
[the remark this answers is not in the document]
connoraird: The tests in core run through a load of indices and check they are all accessed correctly but we can only call the function once if we want to benchmark so this randomly selects indices instead.

tests/benchmarks/test_harmonics.py
```diff
@@ -0,0 +1,37 @@
...
+    xp: ModuleType,
+    benchmark: BenchmarkFixture,
+    benchmark_scale_factor: int,
+) -> None:
+    """Benchmarks for glass.harmonics.multalm."""
+    # Call jax version of iternorm once jax version is written
+    if xp.__name__ == "jax.numpy":
+        pytest.skip("Arrays in multalm are not immutable, so do not support jax")
+
+    scale_factor = benchmark_scale_factor * 100
```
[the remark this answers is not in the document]
connoraird: I decided to remove it as we had a different value already in main.

paddyroddy:
🙏 remember labels

tests/benchmarks/test_fields.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import annotations
+
+from random import randrange
```
paddyroddy: To be consistent with eleswhere can we `import random` then use `random.randrange`? The benefit is at a glance you can see where it comes from.

tests/benchmarks/test_fields.py
```diff
@@ -0,0 +1,72 @@
...
+    result = benchmark(
+        glass.fields.getcl,
+        cls,
+        random_i,
+        random_j,
+        lmax=lmax,
+    )
+    expected = xp.zeros((lmax - 1,), dtype=xp.float64)
+    assert result.size == lmax + 1
+    np.testing.assert_allclose(result[2:], expected)
```
paddyroddy: `compare` fixture

Issue #794, opened 2025-11-17
Create wrapper functions for spherical harmonic transforms


ntessore:
### Is Your Feature Request Related to a Problem? Please Describe

We need to wrap the spherical harmonic transform and inverse transform in a way that will eventually work with multiple Array API backends.

### Describe the Solution You'd Like

Create wrapper functions for the spherical harmonic transforms:

* `glass.harmonics.transform()` wrapping `hp.map2alm()` currently
* `glass.harmonics.inverse_transform()` wrapping `hp.alm2map()` currently

Then we start migrating functions using the SHT to these functions.

The new functions should return and accept spherical harmonic coefficients in "GLASS order" (ell then m) not "healpy order" (m then ell). We have a function `glass.fields._glass_to_healpix_alm()` to do this in one direction.

paddyroddy:
I closed my attempt on this #805, after the work done in creating the `glass.healpix` module #968. Feels like we need to put more thought into how we want to do this.

PR #805, opened 2025-11-18
gh-794: create spherical harmonic transform wrapper functions

paddyroddy:
# Description

Ultimately, we need these transforms to be Array API compatible. Isolating them within the `glass.harmonics` module may make this easier for us further down the line.

Closes: #794

## Changelog entry

Added: Spherical harmonic transforms in the `harmonics` module

## Checks

### DIFF
--- a/glass/harmonics.py
+++ b/glass/harmonics.py
@@ -5,9 +5,9 @@ from __future__ import annotations
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
-import healpy as hp
-
 import array_api_compat
+
+import glass.healpix as hp
 
 if TYPE_CHECKING:
     from glass._types import ComplexArray, FloatArray


tests/core/test_harmonics.py
```diff
@@ -54,3 +58,83 @@ def test_multalm(compare: type[Compare], xp: ModuleType) -> None:
...
+    alm_result = glass.harmonics.transform(
+        simple_map,
+        lmax=lmax,
+        polarised_input=False,
+    )
+
+    compare.assert_allclose(alm_result, alm_expected, rtol=1e-15)
+
+
+def test_inverse_transform(compare: type[Compare], xp: ModuleType) -> None:
```
connoraird: Do we want to continue following the pattern of having multiple tests within one test function or do we want to have a single test function for each test case?

paddyroddy:
I'm going to close this following the work of #968. This whole idea needs further work @ntessore.

Issue #821, opened 2025-11-20
`glass.fields` is being imported inconsistently in `test_fields.py`


paddyroddy:
### Describe the Bug

There are functions in `__init__.py` that aren't being used in the way users would use them

### To Reproduce

```python
import glass
...
```

### Version In Use

2025.3.dev54+g7fc40ff9f.d20251119

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #822, opened 2025-11-20
gh-821: consistent importing of `glass.fields` in tests

paddyroddy:
# Description

Functions exposed in `__init__.py` aren't being used as such.

Closes: #821

## Checks


connoraird (approved):
This is the case in benchmarks. Would we rather just `glass.function`?

Issue #825, opened 2025-11-21
Create testing helper functions for array comparison


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

We want to ultimately reduce our dependence on NumPy as we move to Array API compatibility. I'd like to make a single file where NumPy is imported in the tests. This will replace calls like `np.testing.assert_array_equal` etc.

PR #826, opened 2025-11-21
gh-825: create an array comparison helper to replace `np.testing`

paddyroddy:
# Description

As we move to full Array API adoption, it would be nice to reduce the use of NumPy in tests as well as core functionality. This PR creates a helper class `Compare` in `conftest.py` to replicate some `np.testing.` functionality. In the future, this would mean it would be easy to change if a proper Array API agnostic testing facility became available.

Also tidy up some `glass` imports in the main GLASS. We should be doing the full path internally, but just `glass` (or submodules) in tests/examples.

Closes: #825

## Checks

### DIFF
--- a/tests/core/test_points.py
+++ b/tests/core/test_points.py
@@ -414,7 +414,6 @@ def test_displace_abs(
 
 
 def test_displacement(
-    compare: type[Compare],
     urng: UnifiedGenerator,
     xp: ModuleType,
 ) -> None:


connoraird:
> Regression test failing because `glass/_array_comparison.py` doesn't appear in `main` @connoraird. We should do something about this in the testing.

Should we not be creating _array_comparison in the testing directory? Then it would be available. Plus are we using in outside of tests?

paddyroddy:
> Should we not be creating _array_comparison in the testing directory? Then it would be available. Plus are we using in outside of tests?

So we could, but it is generally discouraged to import stuff from inside the `tests` directory.

The other thing I considered was putting it in `glass.arraytools` and have, e.g. `_assert_array_equal` but then we'd have to suppress warnings about private calls from `ruff` throughout.

tests/conftest.py
```diff
@@ -246,3 +246,55 @@ def shells() -> list[glass.RadialWindow]:
 def benchmark_scale_factor() -> int:
     """Scaling factor to increase the problem size for benchmarks."""
     return 10_000
+
+
+class Compare:
```
paddyroddy: This idea came from https://stackoverflow.com/a/42156088/7359333

Issue #810, opened 2025-11-20
Run regression tests only in latest support Python version


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Discussed with @connoraird. We are currently testing in all Python versions for the benchmarking. This isn't a problem currently as we are running in CI and GLASS is a public repo. However, the tests are flaky and it increases the likelihood of failure. This should mean we get at most 1 failure so we can run again if required. Hopefully this will be less of an issue once the tests get more stable in #791.

PR #830, opened 2025-11-24
gh-810: run regression tests only on latest Python

paddyroddy:
# Description

Performs the regression test only on the latest Python version. We've decided to do this because in reality we only care about modern versions for benchmarking. Further, as the tests are flaky, it makes sense to remove one extra case of potential failure.

Closes: #810

## Checks


connoraird (approved):
Looks good and tests are passing!

Issue #809, opened 2025-11-20
How to benchmark functions that return generators


connoraird:
### Is Your Feature Request Related to a Problem? Please Describe

Naively wrapping a function in the `benchmark` fixture does not provide a true benchmark of the function as each timed execution will only run until the first yield. 

### Describe the Solution You'd Like

A clear way to benchmark functions which return generators. Possibly wrapping the full test in the benchmark decorator?

### Describe Alternatives You've Considered

We could instead not benchmark functions that return generators or stop returning generators from functions.

paddyroddy:
What are the results if you were to run that?

connoraird:
> What are the results if you were to run that?

The tests passes and on my local machine it takes about 55 us

paddyroddy:
I meant the table output

PR #832, opened 2025-11-26
gh-809: Add benchmarks for fields functions which return Generators

connoraird:
# Description

Adds benchmarks for functions within `fields.py` that return Generators.

The approach I have employed is to benchmark the call to the function being tested, which creates the Generator as well as looping through the Generator until it is exhausted. 

> Note: exhausting a generator often causes an exception to be raised, therefore, I have had to catch these exceptions and just pass on them.

Closes: #809 

## Checks

### DIFF
--- a/tests/benchmarks/test_fields.py
+++ b/tests/benchmarks/test_fields.py
@@ -35,10 +35,7 @@ def test_iternorm_no_size(
     array_in = [xp.asarray(x) for x in xp.arange(10_000, dtype=xp.float64)]
 
     def function_to_benchmark() -> list[Any]:
-        generator = glass.fields.iternorm(
-            k,
-            (x for x in array_in),
-        )
+        generator = glass.fields.iternorm(k, iter(array_in))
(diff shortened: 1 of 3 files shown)


tests/benchmarks/test_fields.py
```diff
@@ -3,16 +3,278 @@
...
+        )
+        return _consume_generator(generator)
+
+    gaussian_fields = benchmark(function_to_benchmark)
+
+    assert gaussian_fields[0].shape == (hp.nside2npix(nside),)
+
+
+def _nth_triangular_number(n: int) -> int:
+    return int((n * (n + 1)) / 2)
```
paddyroddy: re-use `glass.fields._inv_triangle_number`?

paddyroddy (changes requested):

tests/benchmarks/test_fields.py
paddyroddy: It's also only used once so does it even need to be a function?

Issue #846, opened 2025-12-01
Investigate replacing `healpy` with `ducc0`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Faster than `healpy` and now Windows compatible

paddyroddy:
Blocked by #794

PR #662, opened 2025-10-15
gh-661: Add Python 3.14 support

paddyroddy:
# Description

Adds Python 3.14 support, reflecting the release last week.

Closes: #661

Added: Python 3.14 support

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [ ] Have you added additional tests (if required)?
- [ ] Have you modified/extended the documentation (if required)?
- [x] Have you added a one-liner changelog entry above (if required)?

PR #847, opened 2025-12-01
gh-846: replace `healpy` dependence with `ducc0`

paddyroddy:
# Description

See if it's possible to switch from `healpy` to `ducc0` in the Python code. `ducc0` is now Windows compatible so this should be easy for people to install.

Closes: #846

## Changelog entry

Changed: Changed from `ducc0` as a backend for SHT instead of `healpy`

## Checks


paddyroddy:
Will have to come back to this

paddyroddy:
Given that we now have 3.14 support in #662 and the uncertainty in what happens with spherical harmonic transforms wrapper https://github.com/glass-dev/glass/issues/794#issuecomment-3542923528, I will close this attempt.

Issue #855, opened 2025-12-04
Test failures on main


connoraird:
### Describe the Bug

There is an intermittent benchmark test failure occuring on main within CI with python 3.14 for this test `tests/benchmarks/test_fields.py::test_generate[numpy-None-4]`. We had previsouly tried to fix this in #836 but the problem still persists. See [this build](https://github.com/glass-dev/glass/actions/runs/19890136481/job/57013559116).

#### TL;DR

In the `test_filed.py::test_generate` benchmark test we consume the Generator returned by `fields.generate` and sometimes the number of successful yields is 4 and sometime it is 7. The exception that is raised comes from this code block in `fields.iternorm`,

```py
if xp.any(s < 0):
    msg = "covariance matrix is not positive definite"
    raise ValueError(msg)
```

### To Reproduce

Run the [Test workflow](https://github.com/glass-dev/glass/actions/workflows/test.yaml) on main.

### Expected Behaviour

The tests always pass

### Actual Behaviour

They intermittently fail.

### Version In Use

d71c3d8336532711254e26598dd2d1746e3b6c70

### Additional Context

```markdown
- Python version: 3.14
```

PR #856, opened 2025-12-04
gh-855: Add verification of the expected error message when consuming generators.

connoraird:
# Description

Adds a check that, if an exception is raised whilst consuming a generator, the exception has the message we expect. This is done to ensure we are not hiding any other exceptions we are unaware of.

Refs: #855

## Checks

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -166,7 +166,6 @@ def iternorm(
         a_np = np.asarray(a, copy=True)
         einsum_result_np = np.einsum("...i,...i", a_np, a_np)
         s = x[..., 0] - xp.asarray(einsum_result_np, copy=True)
-        print(s)  # noqa: T201
         if xp.any(s < 0):
             msg = "covariance matrix is not positive definite"
             raise ValueError(msg)


tests/conftest.py
```diff
@@ -302,6 +302,8 @@ class GeneratorConsumer:
     @staticmethod
     def consume(
         generator: Generator[Any],
+        *,
+        valid_exception: str = "No exception should have been thrown",
```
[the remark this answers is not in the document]
connoraird: Either would be fine. I thought the default message would make it easier to determine what the issue was (i.e. we forgot to specify the expected message)

tests/fixtures/helper_classes.py
```diff
@@ -110,6 +109,8 @@ class GeneratorConsumer:
     @staticmethod
     def consume(
         generator: Generator[Any],
+        *,
+        valid_exception: str = "No exception should have been thrown",
```
paddyroddy: Still not sure about this

connoraird: Updated

glass/fields.py
```diff
@@ -166,6 +166,7 @@ def iternorm(
         a_np = np.asarray(a, copy=True)
         einsum_result_np = np.einsum("...i,...i", a_np, a_np)
         s = x[..., 0] - xp.asarray(einsum_result_np, copy=True)
+        print(s)  # noqa: T201
```
paddyroddy:
```suggestion
```

paddyroddy (approved):
Nice

Issue #866, opened 2025-12-05
Exclude imports from coverage report


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

Missed in #865. It might make sense to also exclude imports from the coverage import. Files like `glass/cosmomology.py` report zero coverage although the class is excluded due to the imports.

PR #865, opened 2025-12-04
gh-861: be more careful with coverage exclusion

paddyroddy:
# Description

Follow up of #859 with branch name change. In the [coverage.py documentation](https://coverage.readthedocs.io/en/latest/excluding.html#advanced-exclusion) they recommend the following

```toml
[tool.coverage.report]
exclude_also = [
    'def __repr__',
    'if self.debug:',
    'if settings.DEBUG',
    'raise AssertionError',
    'raise NotImplementedError',
    'if 0:',
    'if __name__ == .__main__.:',
    'if TYPE_CHECKING:',
    'class .*\bProtocol\):',
    '@(abc\.)?abstractmethod',
]
```

we are doing something similar, but it's not working. This PR aims to address that.

Closes: #861

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [ ] Have you added additional tests (if required)?
- [ ] Have you modified/extended the documentation (if required)?
- [ ] Have you added a one-liner changelog entry above (if required)?

PR #867, opened 2025-12-05
gh-866: exclude imports from coverage report

paddyroddy:
# Description

Following the work in #865, this PR further excludes imports. Currently, `glass/cosmology.py` reports 0% coverage, which isn't true. This is a bit overkill but should make the percentage more clear.

Have also adjusted the `@overload` decorator to avoid the case of, i.e. `@overloaded`,

Closes: #866

## Checks

### DIFF
--- a/tests/benchmarks/test_points.py
+++ b/tests/benchmarks/test_points.py
@@ -6,7 +6,7 @@ import numpy as np
 import pytest
 
 import glass
-from tests.conftest import xp_available_backends
+import tests.conftest
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


connoraird (commented):
The coverage seems to have decreased. Does this not remove the lines from the percentage?

paddyroddy:
> The coverage seems to have decreased. Does this not remove the lines from the percentage?

So I think it does remove the lines, but the percentage will depend on what was covered. Here, the only noticeable decrease is for the benchmarks. If we look at, e.g. `glass/algorithm.py`. I think it makes sense that the coverage has "decreased".

#### Before 8.25%

Manually counting I see:
* 4 import lines covered
* 4 non-import lines covered
* 75 lines not covered

8/75=10.7% 🤔 might be something to do with branch coverage

coverage before change

#### After 4.3%

Manually counting I see:
* 0 import lines covered
* 4 non-import lines covered
* 75 lines not covered

4/75=5.3% 🤔 might be something to do with branch coverage

coverage after change

paddyroddy:
> So the import lines are still counted in the total number of lines to be covered.

This statement is incorrect

connoraird (approved):

Issue #870, opened 2025-12-08
Split up the `conftest.py`


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

The `conftest.py` has become a bit unwieldy. We could consider splitting it up, either through different `conftest.py`. Alternatively, we could use `pytest_plugins` as discussed in this article https://swiety-python.blogspot.com/2016/07/pytest-splitting-conftest-file.html.

Issue #871, opened 2025-12-09
Remove need to import conftest


connoraird:
### Describe the Bug

We are currently importing `xp_available_backends` from `conftest` in order to filter tests for certain backends without simply using `pytest.skip`. It is not recommended to import conftest. So, instead, we should add new fixtures to handle this desired functionality

### To Reproduce

```python
Run the benchmark tests.
```

### Expected Behaviour

Tests pass without importing anything from `conftest` and `test_displacement_jax` runs for jax only and is not skipped.

### Version In Use

main

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
I think it's even more general than that. It's discouraged to import from within `tests` generally.

PR #872, opened 2025-12-09
gh-870: split up `conftest.py` into separate files

paddyroddy:
# Description

Our `conftest.py` has become a bit unwieldy, so this PR leverages `pytest_plugins` to simplify this. I have split into four different files for now.

Closes: #870

## Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -229,27 +229,23 @@ lint.per-file-ignores = {"__init__.py" = [
     "TRY003", # raise-vanilla-args
 ], "glass/grf/_*" = [
     "SLF001", # private-member-access
+], "tests/*" = [
+    "INP001", # implicit-namespace-package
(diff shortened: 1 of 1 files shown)


tests/benchmarks/test_points.py
```diff
@@ -179,7 +183,7 @@ def _benchmark_displacement(
     "xp",
     [
         xp
-        for name, xp in tests.conftest.xp_available_backends.items()
+        for name, xp in tests.fixtures.array_backends.xp_available_backends.items()
```
connoraird: This is an alternative to #873. Which do you prefer?

paddyroddy: https://github.com/glass-dev/glass/issues/871#issuecomment-3636426632

paddyroddy:
This is still failing due to #871. I will merge this anyway.

PR #873, opened 2025-12-09
gh-871: Filter xp in tests without having to import conftest

connoraird:
# Description

The solution here is not as good as I hoped it would be -- I still have to skip a test when jax is not imported which matches the implementation in `core/test_array_api_utils.py`

Closes: #871

## Checks

### DIFF
--- a/tests/benchmarks/test_points.py
+++ b/tests/benchmarks/test_points.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
+import importlib.util
 from typing import TYPE_CHECKING
 
 import numpy as np
 import pytest
 
 import glass
-import tests.conftest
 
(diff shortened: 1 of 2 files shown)


paddyroddy:
Based on what we discussed yesterday, do we want to skip JAX entirely? Or put the JAX tests in their own folder `tests/benchmarks/jax`?

connoraird:
> Based on what we discussed yesterday, do we want to skip JAX entirely? Or put the JAX tests in their own folder `tests/benchmarks/jax`?

I think there are benefits of having jax benchmarks, for example if there is a desire to produce some sort of report on the performance of glass. However, they are not needed for our regression testing purposes and will fail if we try to run against an old version which does not support jax.

Issue #883, opened 2025-12-11
More efficient code using list comprehension


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

There are a few cases where we could get a small speed boost by tweaking a few functions

PR #886, opened 2025-12-11
gh-883: Small algorithmic changes

paddyroddy:
# Description

This PR introduces a few changes like using list comprehension etc.

Closes: #883

## Checks

### DIFF
--- a/glass/grf/_transformations.py
+++ b/glass/grf/_transformations.py
@@ -33,7 +33,9 @@ class Normal:
         x: AnyArray,
         /,
     ) -> AnyArray | NotImplementedType:
-        return x if type(other) is Normal else NotImplemented
+        if type(other) is Normal:
+            return x
+        return NotImplemented
 
     def icorr(
         self,
@@ -41,7 +43,9 @@ class Normal:
         x: AnyArray,
         /,
     ) -> AnyArray | NotImplementedType:
-        return x if type(other) is Normal else NotImplemented
+        if type(other) is Normal:
+            return x
+        return NotImplemented
 
     def dcorr(
         self,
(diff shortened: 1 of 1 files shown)


glass/fields.py
```diff
@@ -888,8 +888,8 @@ def glass_to_healpix_spectra(spectra: Sequence[T]) -> list[T]:
     """
     n = nfields_from_nspectra(len(spectra))
 
-    comb = [(i, j) for i, j in spectra_indices(n)]
-    return [spectra[comb.index((i + k, i))] for k in range(n) for i in range(n - k)]
+    comb = {tuple(idx_tuple): pos for pos, idx_tuple in enumerate(spectra_indices(n))}
+    return [spectra[comb[(i + k, i)]] for k in range(n) for i in range(n - k)]
```
paddyroddy: O(n) rather than O(n^2)

glass/grf/_transformations.py
```diff
@@ -33,29 +33,23 @@ def corr(
...
     def dcorr(
         self,
         other: Transformation,
         x: AnyArray,
         /,
     ) -> AnyArray | NotImplementedType:
-        if type(other) is Normal:
-            return 1.0 + (0 * x)
-        return NotImplemented
+        return 1.0 + (0 * x) if type(other) is Normal else NotImplemented
```
connoraird: Does this not cause this calculation on the left to be done no matter the result of the condition? Whereas before it would only be computed if the condition was met?

glass/grf/_transformations.py
```diff
@@ -99,10 +93,7 @@ def corr(
         if type(other) is Lognormal:
             return self.lamda * other.lamda * xp.expm1(x)
```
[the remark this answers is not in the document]
paddyroddy: There's no `else`

PR #904, opened 2025-12-16
gh-432: Run test_multalm with jax

connoraird:
# Description

Runs test_multalm with jax as it does not mutate any arrays.

Refs: #432 

## Checks


paddyroddy (approved):
Good spot

Issue #910, opened 2025-12-17
Displacement is ambiguous


ntessore:
### Describe the Bug

The output of `glass.displacement()` is not clear regarding what "east" and "west" mean in terms of geometry. This needs to be checked.

### To Reproduce

```python
import glass
...
```

### Version In Use

main

### Additional Context

```markdown
- Python version:
- Operating system:
```

PR #911, opened 2025-12-17
gh-910: consistent definition of displacement

ntessore:
# Description

Consistently defines the displacement in `glass.displace()`, `glass.displacement()`, and the documentation.

In short, displacement is now everywhere defined as on a compass, like suggested by the docs. Going "east" means increasing longitude, and going "west" means decreasing longitude.

I also changed the way `displacement()` is computed so that it no longer divides by zero, and added a test to that effect.

Furthermore, I added a test for checking that `displace()` and `displacement()` are consistent with each other.

Finally, I added a test for `displacement()` wherein I painstakingly compute random points with known displacement in a manner that is completely independent of the implemented method (by manually constructing a rotation matrix and rotating points).

Fixes: #910

## Changelog entry

Fixed: `glass.displace()` and `glass.displacement()` are now fully consistent

## Checks

### DIFF
--- a/tests/benchmarks/test_points.py
+++ b/tests/benchmarks/test_points.py
@@ -16,7 +16,6 @@ if TYPE_CHECKING:
 
     from glass._types import UnifiedGenerator
     from tests.fixtures.helper_classes import (
-        Compare,
         DataTransformer,
         GeneratorConsumer,
     )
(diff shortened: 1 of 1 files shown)


ntessore:
pre-commit.ci autofix

Issue #912, opened 2025-12-17
Use `math.pi` rather than `xp.pi`


paddyroddy:
### Add your issue here

To ease in array API transition

ntessore:
Ok! Do we want to use `math.pi` in new code then or is `xp.pi` still good? I don't mind either way but it would be nice to have guidance in place, even if it's "use whatever".

(Why does `xp.pi` exist at all? Is there any situation in which using `xp.pi` gives a different result than `math.pi`? Does jax complain when it's running in single-precision mode and you divide by `math.pi` or something?)

PR #913, opened 2025-12-17
gh-912: use `math.pi`

paddyroddy:
# Description

Consistent use of `math.pi` over `xp.pi` to aid in Array API porting.

Closes: #912

## Checks


paddyroddy:
> There are some instances of `np.pi` in the examples, if that matters?

I did see that. I guess it makes sense too.

paddyroddy:
> I did see that. I guess it makes sense too.

Actually, won't bother.

Issue #915, opened 2025-12-18
Array backend import error fails silently


connoraird:
### Describe the Bug

If we request `ARRAY_BACKEND=all` and importing a backend fails, this `ImportError` is swallowed silently and the test continue without that backend. This can cause us to not test certain backends via the CI if we are unable to import them.

### To Reproduce

Run the tests with `ARRAY_BACKEND=all` using python 3.10.

### Expected Behaviour

If a backend is requested for testing, including if requested via `ARRAY_BACKEND=all`, and that backend fails to be imported, an error should be logged.

### Actual Behaviour

The failed backends are swallowed silently and just the successfully imported backends are ran for the tests.

### Version In Use

2025.4.dev23+g6a68a4e25

### Additional Context

```markdown
- Python version: 3.10
```

PR #917, opened 2025-12-18
gh-915: Fail tests if array backend not found

connoraird:
# Description

Stops swallowing ImportErrors for `ARRAY_BACKEND=all`

Closes: #915

## Checks

### DIFF
--- a/glass/_array_api_utils.py
+++ b/glass/_array_api_utils.py
@@ -1,6 +1,6 @@
 """
 Array API Utilities for glass.
-============================
+==============================
 
 This module provides utility functions and classes for working with multiple array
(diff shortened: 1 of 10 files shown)


connoraird:
@paddyroddy This prevents the tests from running and fails early. I could instead fail each test by setting to backend to None and then asserting that the backend is not None in the xp Fixture?

paddyroddy:
> @paddyroddy This prevents the tests from running and fails early. I could instead fail each test by setting the backend to None and then asserting that the backend is not None in the xp Fixture?

Let's discuss later, not sure I fully follow

PR #919, opened 2025-12-18
gh-903: Port galaxies.py to support jax

connoraird:
# Description

Ports functions in galaxies.py which mutate arrays to now use `xpx.at...` so they support jax.

Closes: #903 

## Changelog entry

Added: jax support for `glass.galaxies.redshifts_from_nz` and `glass.galaxies.galaxy_shear `.

## Checks

### DIFF
--- a/tests/fixtures/array_backends.py
+++ b/tests/fixtures/array_backends.py
@@ -40,6 +40,7 @@ def _check_version(lib: str, array_api_compliant_version: str) -> None:
     ------
     ImportError
         If the installed version is not compliant with the array API standard.
+
     """
     lib_version = packaging.version.Version(importlib.metadata.version(lib))
     if lib_version < packaging.version.Version(array_api_compliant_version):
@@ -100,6 +101,7 @@ def xp(request: pytest.FixtureRequest) -> ModuleType:
     Fixture for array backend.
 
     Access array library functions using `xp.` in tests.
+
     """
     return request.param  # type: ignore[no-any-return]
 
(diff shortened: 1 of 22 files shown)


connoraird:
1. Coveralls reduction I assume is because of the removal of array-api-strict
2. I believe this could be a genuine regression in performance

tests/fixtures/array_backends.py
```diff
@@ -110,7 +110,7 @@ def xp(request: pytest.FixtureRequest) -> ModuleType:
     params=[
         xp
         for name, xp in xp_available_backends.items()
-        if name not in {"array-api-strict", "jax.numpy"}
+        if name not in {"array_api_strict", "jax.numpy"}
```
connoraird: Noticed this was the wrong string to match against (see #922). I have implemented in a separate PR here https://github.com/glass-dev/glass/pull/928

paddyroddy (approved):
Looks good. How do we know it's a genuine reduction this time?

paddyroddy:
That would make sense. Thankfully not a huge regression.

```rst
Performance has regressed:
	test_galaxy_shear[numpy-True] (0001_f906eb0) - Field 'mean' has failed PercentageRegressionCheck: 9.996939723 > 5.000000000
	test_galaxy_shear[numpy-False] (0001_f906eb0) - Field 'mean' has failed PercentageRegressionCheck: 8.899524405 > 5.000000000
```

paddyroddy (approved):
Let's wait until @ntessore has looked at this

paddyroddy:
> Could this be a regression due to use of `urng` instead of `rng`? Or is `xpx.at()` generally that much slower than index access?

So [urngb](https://github.com/glass-dev/glass/blob/17e8e653b26cc8750d115f28e47d253fc802e1e5/tests/benchmarks/test_galaxies.py#L68) was already in use in that benchmark, and isn't used by `galaxy_shear`. So it could be the `xpx.set` call. This is quite a small regression, so I'm not worried personally.

Issue #921, opened 2025-12-18
Pin `pre-commit` to lowest supported version


paddyroddy:
### Describe the Bug

#603 pinned `pre-commit.ci` which fixed an annoying bug, but his preventing @ntessore from using `pre-commit` locally.

### To Reproduce

```python
import glass
...
```

### Version In Use

2025.4.dev27+geae29620c

### Additional Context

```markdown
- Python version:
- Operating system:
```

Issue #923, opened 2025-12-18
Use `ty` for type checking support


paddyroddy:
### Add your issue here

`ty` has become stable https://astral.sh/blog/ty. We should move towards it as `mypy` is difficult to work with and hopefully `ty` is better (and definitely faster).

PR #924, opened 2025-12-18
gh-921: pin `pre-commit` to oldest support Python version

paddyroddy:
# Description

Pin the mypy Python version. I have also updated the `additional_dependencies`.

This seems to have degraded `mypy` support, not full sure why. We should fix these at some point, but for now have suppressed them. Some of this might be fixed by #923.

Closes: #921

## Changelog entry

Changed: `pre-commit` is now run with the lowest supported Python version

## Checks

### DIFF
--- a/tests/fixtures/array_backends.py
+++ b/tests/fixtures/array_backends.py
@@ -107,14 +107,28 @@ def xp(request: pytest.FixtureRequest) -> ModuleType:
 
 
 @pytest.fixture(
-    params=[xp for name, xp in xp_available_backends.items() if name != "jax.numpy"],
+    params=[
+        xp
+        for name, xp in xp_available_backends.items()
+        if name not in {"array-api-strict", "jax.numpy"}
+    ],
     scope="session",
 )
 def xpb(request: pytest.FixtureRequest) -> ModuleType:
     """
-    Fixture for array backend.
+    Fixture for array backend to be used in benchmarks.
 
-    Access array library functions using `xp.` in tests.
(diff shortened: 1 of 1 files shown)


tests/benchmarks/test_lensing.py
```diff
@@ -8,14 +8,13 @@
...
-    from typing import Never
 
     from conftest import Compare
     from pytest_benchmark.fixture import BenchmarkFixture
-
-    from cosmology import Cosmology
+    from typing_extensions import Never
 
     from glass._types import FloatArray, UnifiedGenerator
+    from glass.cosmology import Cosmology
```
paddyroddy: This is unrelated, I just spotted it

ntessore:
Ok, I think this is due to a mismatch in intended use and our use: the pre-commit docs say that `default_language_version` sets `language_version`, and `language_version` is the version used to _run_ pre-commit hooks (see [here](https://pre-commit.com/#overriding-language-version)). By setting the default to any version, we are telling pre-commit to look for that exact Python interpreter to run hooks, which is not what we want. So we should remove the pre-commit instruction entirely.

What we want to do is tell `ruff` and `mypy` to run their checks against a specific Python version. Otherwise, they assume our `requires-python` setting, which is 3.10 currently. If we want to check against Python 3.14 (which I think we want), we can use the tool-specific settings in pyproject.toml:

```toml
[tool.mypy]
python_version = "3.14"

[tool.ruff]
target-version = "3.14"
```

This is now completely decoupled from the environment that a developer might have on their machine.

PR #953, opened 2026-01-07
gh-906: Port `observations.vmap_galactic_ecliptic` to array-api

connoraird:
# Description

Ports `observations.vmap_galactic_ecliptic` to the array api by simply converting to the requested array namespace before returning. I have done it this way as we would have been unable to remove np completely from the function due to the use of `healpy` and since none of the inputs are arrays.

Closes: #906 

## Changelog entry

Added: Ported `glass.vmap_galactic_ecliptic` to the array-api.

## Checks


paddyroddy (approved):
Looks good

PR #954, opened 2026-01-08
gh-908: Port `shapes.py` to support jax

connoraird:
# Description

Removes mutations from three functions in `shapes.py` using `array_api_extra.at`.

Closes: #908

## Changelog entry

Added: Support for `jax.numpy` in `glass.ellipticity_ryden04`, `glass.ellipticity_gaussian` and `glass.ellipticity_intnorm`

## Checks

### DIFF
--- a/tests/core/test_shapes.py
+++ b/tests/core/test_shapes.py
@@ -167,11 +167,6 @@ def test_ellipticity_intnorm(
     urng: UnifiedGenerator,
     xp: ModuleType,
 ) -> None:
-    if xp.__name__ == "jax.numpy":
-        pytest.skip(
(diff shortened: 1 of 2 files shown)


tests/core/test_shapes.py
```diff
@@ -64,11 +64,6 @@ def test_triaxial_axis_ratio(urng: UnifiedGenerator, xp: ModuleType) -> None:
 
 
 def test_ellipticity_ryden04(urng: UnifiedGenerator, xp: ModuleType) -> None:
```
paddyroddy: Have verified these three tests have associated benchmarks

paddyroddy (approved):
This is nice. `xpx.set` makes things so straightforward!

Issue #382, opened 2024-10-25
Consider using `ragged` arrays to handle `cls` data


paddyroddy:
### Is Your Feature Request Related to a Problem? Please Describe

https://github.com/scikit-hep/ragged is designed to handle ragged arrays in an Array API compliant manner, i.e.

```python
>>> import ragged
>>> a = ragged.array([[[1.1, 2.2, 3.3], []], [[4.4]], [], [[5.5, 6.6, 7.7, 8.8], [9.9]]])
>>> a
ragged.array([
    [[1.1, 2.2, 3.3], []],
    [[4.4]],
    [],
    [[5.5, 6.6, 7.7, 8.8], [9.9]]
])
```

paddyroddy:
Moving to blocked due to uncertainty behind #403, more specifically the use of `s2ftt`.

paddyroddy:
@Saransh-cpp you have more experience with `ragged`. Do you have any chance to make a comment about this? From the investigation in #973, it looks like it might be possible to use `ragged`.

Saransh-cpp:
I think I have a solution.

Given that Awkward takes the underlying NumPy array as a view, it can be mutated (through the NumPy array, as in my comment above), but, given that Awkward copies Python objects (such as a List of NumPy arrays), they are pretty much immutable. However, you can go into Ragged's (Awkward's) low-level API to actually mutate the array:

```py
a = [np.array([1, 2, 3]), np.array([1, 2])]
r = ragged.asarray(a)

# r -> ragged.array([
#     [1, 2, 3],
#     [1, 2]
# ])

# will give you the upstream Awkward array - Ragged arrays are just specialised Awkward arrays
r._impl
# <Array [[1, 2, 3], [1, 2]] type='2 * var * int64'>

# the XML-like representation of an Awkward array
# notice how the content is contiguous, and the offsets store the how to look at it in the memory (giving that ragged functionality to the user)
r._impl.layout
# <ListOffsetArray len='2'>
#     <offsets><Index dtype='int64' len='3'>[0 3 5]</Index></offsets>
#     <content><NumpyArray dtype='int64' len='5'>[1 2 3 1 2]</NumpyArray></content>
# </ListOffsetArray>

# and finally mutate the array
# you will need to use `r._impl.layout.offsets.data` to navigate to specific indices of the array if you want to maintain the nice indexing syntax
r._impl.layout.content.data[3] = 999
# r -> ragged.array([
#     [1, 2, 3],
[...]

PR #973, opened 2026-01-15
gh-382: Test ragged arrays

connoraird:
# Description

Tests using ragged array which appears to cause an error as it does not work with `xpx.at...` (see [issue comment](https://github.com/glass-dev/glass/issues/382#issuecomment-3755263508))

Refs: #382


connoraird:
Closing as this was a useful experiment but will never be merged

Issue #998, opened 2026-01-23
Decide on how we emulate `len(x)` in the array api


connoraird:
### Add your issue here

The array api does not support `len(x)`. Therefore, we must choose one of `x.shape[0]` or `x.size`

PR #1032, opened 2026-01-29
gh-998: Change all references to `len` to `.shape[0]`

paddyroddy:
# Description

The Array API spec doesn't have a definition of `len()`. This PR changes `.size` references to `.shape[0]`. It also changes some current `len()` to `.shape[0]`.

Closes: #998

## Checks

### DIFF
--- a/glass/galaxies.py
+++ b/glass/galaxies.py
@@ -152,7 +152,7 @@ def redshifts_from_nz(
         )
         total += count_out[k]
 
-    assert total == redshifts.shape[0]  # noqa: S101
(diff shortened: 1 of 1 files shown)


paddyroddy:
> So are we not sure if they used to be `len` before they were converted to `size`?

I wasn't clear what you were asking. I guess not.

paddyroddy:
> If we trust our tests then this is okay

I probably don't... Will check tomorrow.

paddyroddy:
> I probably don't... Will check tomorrow.

I've been more cautious and reverted all of them unless `cl` or in tests.

PR #1047, opened 2026-03-04
gh-1026: remove `np.asarray` from `generate`

### LINKED ISSUE #1026: Remove `np.assaray` from `generate`
### Add your issue here

Once we have merged and released https://github.com/glass-dev/transformcl/pull/6, we can upgrade to the new version and remove the `np.asarray` call from `generate`. See https://github.com/glass-dev/glass/pull/979#issuecomment-3816584549 for details.

paddyroddy:
# Description

Following https://github.com/glass-dev/transformcl/pull/6 we can now safely remove the `np.asarray` call within `generate`. This is because `cltovar` is now Array API compatible.

Closes: #1026

## Checks


connoraird (approved):
~Looks good~ One query below

glass/fields.py
connoraird: Can we now remove `xp.asarray`?

connoraird (changes requested):
Actually, one comment

glass/fields.py
paddyroddy: Looking at `cltovar` we should be able to

connoraird (approved):
Nice!

PR #1065, opened 2026-05-11
gh-1063: Default to numpy for all unrecognised array backends

### LINKED ISSUE #1063: Default to NumPy if a function is not supported

Currently, we throw an exception if the library used is not one of NumPy, Jax or array-api-strict.

To allow the use of any array backend, we should instead default to a library such as NumPy if we do not recognise the provided library as a library which has implemented the function in question.

_No response_

### Additional Context

_No response_

connoraird:
# Description

- Adds logic to default to numpy for any given backend which is not explicitly supported by GLASS.
- Skips a test which relies on `array_api_strict` if it isn't installed.
- ~Also adds CuPy as a library for testing to check this new functionality.~

Closes: #1063

## Changelog entry

Changed: Instead of raising an exception, we now default to NumPy if an array backend is not supported.

## Checks

### DIFF
--- a/glass/_array_api_utils.py
+++ b/glass/_array_api_utils.py
@@ -47,9 +47,10 @@ class CompatibleBackendNotFoundError(Exception):
         super().__init__(self.message)
 
 
-def import_numpy(backend: str | None = None) -> ModuleType:
+def default_xp(backend: str | None = None) -> ModuleType:
     """
-    Import the NumPy module, raising a helpful error if NumPy is not installed.
+    Returns the library backend we default to (NumPy) if none is specified by the user.
+    Raises a helpful error if NumPy is not installed.
 
     Parameters
     ----------
@@ -78,11 +79,6 @@ def import_numpy(backend: str | None = None) -> ModuleType:
         raise CompatibleBackendNotFoundError("numpy", backend) from err
     else:
         return numpy
-
-
-def default_xp() -> ModuleType:
-    """Returns the library backend we default to if none is specified by the user."""
-    return import_numpy()
 
 
 class xp_additions:  # noqa: N801
@@ -123,11 +119,6 @@ class xp_additions:  # noqa: N801
         -------
             Integrated result.
 
-        Raises
-        ------
-        NotImplementedError
-            If the array backend is not supported.
-
         Notes
         -----
         See https://github.com/glass-dev/glass/issues/646
@@ -143,17 +134,12 @@ class xp_additions:  # noqa: N801
         if xp.__name__ == "numpy":
             return xp.trapezoid(y, x=x, dx=dx, axis=axis)
 
-        if xp.__name__ == "array_api_strict":
-            np = import_numpy(xp.__name__)
-
-            # Using design principle of scipy (i.e. copy, use np, copy back)
-            y_np = np.asarray(y, copy=True)
-            x_np = np.asarray(x, copy=True)
-            result_np = np.trapezoid(y_np, x_np, dx=dx, axis=axis)
-            return xp.asarray(result_np, copy=True)
-
-        msg = "the array backend in not supported"
-        raise NotImplementedError(msg)
+        # If any other backend use default
+        dxp = default_xp(xp.__name__)
+        y_dxp = dxp.asarray(y, copy=True)
+        x_dxp = dxp.asarray(x, copy=True)
+        result_dxp = dxp.trapezoid(y_dxp, x_dxp, dx=dx, axis=axis)
+        return xp.asarray(result_dxp, copy=True)
 
     @staticmethod
     def interp(  # noqa: PLR0913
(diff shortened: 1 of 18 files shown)


glass/_array_api_utils.py
```diff
@@ -47,9 +47,10 @@ def __init__(self, missing_backend: str, users_backend: str | None) -> None:
         super().__init__(self.message)
 
 
-def import_numpy(backend: str | None = None) -> ModuleType:
+def default_xp(backend: str | None = None) -> ModuleType:
     """
-    Import the NumPy module, raising a helpful error if NumPy is not installed.
+    Returns the library backend we default to (NumPy) if none is specified by the user.
+    Raises a helpful error if NumPy is not installed.
```
paddyroddy: Do we want to specify NumPy here? In future, it's possible we might default to another backend I guess.

glass/_array_api_utils.py
```diff
@@ -80,11 +81,6 @@ def import_numpy(backend: str | None = None) -> ModuleType:
         return numpy
 
 
-def default_xp() -> ModuleType:
-    """Returns the library backend we default to if none is specified by the user."""
-    return import_numpy()
```
paddyroddy, line 85: For example, here we never said NumPy in the docstring.

glass/_array_api_utils.py
```diff
@@ -143,17 +134,12 @@ def trapezoid(
         if xp.__name__ == "numpy":
             return xp.trapezoid(y, x=x, dx=dx, axis=axis)
 
-        if xp.__name__ == "array_api_strict":
-            np = import_numpy(xp.__name__)
-
-            # Using design principle of scipy (i.e. copy, use np, copy back)
```
paddyroddy, line 149: This comment has been lost (and elsewhere). Do we want it removed?

glass/_rng.py
```diff
@@ -48,42 +48,39 @@ def rng_dispatcher(*, xp: ModuleType) -> UnifiedGenerator:
...
+    NumPy random number generator returning Arrays of the given backend.
 
     This class wraps NumPy's random number generator and returns arrays compatible
-    with array_api_strict.
+    with the provided backend.
 
     """
 
-    __slots__ = ("ap", "np", "rng")
+    __slots__ = ("np", "rng", "xp")
```
paddyroddy, line 63: The order here doesn't matter right? Assume you've gone trusty alphabetical? In which case I'm a fan.

tests/core/test_array_api_utils.py
```diff
@@ -22,29 +22,31 @@ def test_rng_dispatcher_numpy() -> None:
 
 @pytest.mark.skipif(not HAVE_JAX, reason="test requires jax")
 def test_rng_dispatcher_jax() -> None:
-    import jax.numpy as jnp
+    import jax.numpy as jnp  # ty: ignore[unresolved-import]
```
paddyroddy, line 25: Trying to work out why we need now 🤔

tests/core/test_healpix.py
```diff
@@ -316,6 +319,7 @@ def test_query_strip_float64(
     compare.assert_array_equal(old, new)
 
 
+@pytest.mark.skipif(not HAVE_ARRAY_API_STRICT, reason="test requires array_api_strict")
```
paddyroddy, line 322: Do we need this now from the work of this PR or is this something previously missed?

paddyroddy (commented):
Some Qs

glass/_array_api_utils.py
connoraird: I thought about this too. I wasn't sure how to remove it but I'll have a go.

glass/_array_api_utils.py
connoraird: Do you think it's useful?

glass/_rng.py
connoraird: I think this was a linter auto-fix

tests/core/test_array_api_utils.py
connoraird: Yes, I'm not sure why this wasn't a problem before. I guess we haven't ran ty against this file before? Shoudl we just ignore that rule everywhere?

tests/core/test_healpix.py
connoraird: I think this was previously missed. I just tried to run with ARRAY_BACKEND=numpy and got an error.

tests/core/test_array_api_utils.py
paddyroddy: It's probably just that `ty` has updated since we last ran it. I think leave `ty` on for now as it's still early in development.

glass/_array_api_utils.py
paddyroddy: Not really, it's fairly self-explanatory

glass/_rng.py
paddyroddy: Even better!

tests/core/test_healpix.py
paddyroddy: Not sure how we missed that. Are we missing something in our CI for this?

glass/_array_api_utils.py
connoraird: Updated

paddyroddy (approved):
Looks good

tests/core/test_healpix.py
connoraird: If we always install array-api-strict, then it'll pass.

paddyroddy: I know. But we could have a separate test to ensure we're testing it properly. I'm thinking something along the lines of https://github.com/glass-dev/glass/pull/670

PR #1064, opened 2026-05-11
gh-984: Add config to run on Archer2

connoraird:
# Description

- Adds submission scripts and documentation of how to run the regression tests on Archer2.

## Testing

- I have tested the archer2 script by running a regression test with fd80e8fb20648602d956ac399e1568d90c11d113 as the starting version and main as the final version and received an error due to the change in the glass modules which causes benchmarks for fd80e8fb20648602d956ac399e1568d90c11d113 to fail.
- I have also ran comparing `main` to `connor/run-on-archer2` which runs as expected.

Refs: #984
Closes: #984 

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [ ] Have you added additional tests (if required)?
- [x] Have you modified/extended the documentation (if required)?
- [ ] Have you added a one-liner changelog entry above (if required)?

PR #1068, opened 2026-05-12
gh-995: Allow filtering a single benchmark test

### LINKED ISSUE #995: Add the ability to run regression tests only on certain tests

When you have a failure, you have to run over the whole code locally. This is fine for CI, but a pain locally.

_No response_

_No response_

### Additional Context

_No response_

connoraird:
# Description

- Allows the passing of pytest flags to the benchmark and regression tests nox commands

Fixes: #995 

## Changelog entry

Added: Ability to run benchmarks/regression tests for a single test

## Checks

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -64,10 +64,10 @@ def _check_revision_count(
         msg = f"{expected_count} revision(s) not provided"
         raise ValueError(msg)
 
-    if len(session_posargs) != expected_count:
+    if len(session_posargs) < expected_count:
         msg = (
             f"Incorrect number of revisions provided ({len(session_posargs)}), "
-            f"expected {expected_count}"
+            f"expected at least {expected_count}"
         )
         raise ValueError(msg)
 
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -239,7 +239,12 @@ def benchmarks(session: nox.Session) -> None:
 
     # overwrite current package with specified revision
     session.install(f"git+{GLASS_REPO_URL}@{revision}")
-    session.run("pytest", BENCH_TESTS_LOC)
+    session.run(
+        "pytest",
+        BENCH_TESTS_LOC,
+        *SHARED_BENCHMARK_FLAGS,
```
paddyroddy, line 245: Why weren't we using these before?

paddyroddy (changes requested):
These need to be documented somewhere.

Also do you think it's clearest having one function rather than `benchmark` and `benchmark_all`? I know it's not very DRY, but I'm just thinking of other users that aren't us

noxfile.py
connoraird: I think we just missed this.

connoraird:
> Also do you think it's clearest having one function rather than `benchmark` and `benchmark_all`? I know it's not very DRY, but I'm just thinking of other users that aren't us

I think it's clearer to have `benchmarks` and pass through the pytest args as that is what we do eslewhere, for example, in `tests.

paddyroddy:
> I think it's clearer to have `benchmarks` and pass through the pytest args as that is what we do eslewhere, for example, in `tests.

Okay, fair enough

paddyroddy (approved):
Looks good

connoraird:
I missed out a call to `_setup_array_backend(session)` in the noxfile function  `benchmarks`. I'll add it as part of #1064

PR #1073, opened 2026-05-15
gh-1070: Replace `pre-commit.ci` with manual workflow

### LINKED ISSUE #1070: Switch from `pre-commit.ci` to a manual workflow
### Add your issue here

Discussed with @connoraird whilst tackling #1069. Want a consistent environment to what devs are running.

paddyroddy:
# Description

Removes @pre-commit-ci and replaces it with a manual workflow. This has been so that we are getting consistent results locally and in CI, namely by running `nox -s lint`. For now this will fail until #1072 is merged. I have also updated @dependabot to update the @pre-commit hooks that was being done before. Have also documented the use of `prek` which is a faster version of `pre-commit` with a friendlier community.

We're experiencing a strange bug in the documentation build https://github.com/vale-cli/vale-action/issues/156 which is unrelated to this PR.

Closes: #1070

## Changelog entry

Changed: Replaces `pre-commit.ci` with the use of `prek` via `nox`

## Checks


.pre-commit-config.yaml
```diff
@@ -1,12 +1,4 @@
 ---
```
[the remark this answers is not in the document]
paddyroddy: Already uninstalled the app

noxfile.py
```diff
@@ -74,13 +74,13 @@ def _check_revision_count(
 
 @nox_uv.session(
     # ty requires all dependencies to be installed to be the most effective
-    uv_all_extras=True,
     uv_all_groups=True,
+    uv_no_install_project=True,
     uv_sync_locked=False,
```
connoraird: Just to sanity check, is this installing all the dependencies needed for `ty` to fully understand the typing?

connoraird (approved):
Looks good. Just one question

noxfile.py
paddyroddy: I've put it back to

```python
uv_all_groups=True,
uv_sync_locked=False,
```

Not sure if we need to install the project itself, but better to be safe

paddyroddy:
Docs failure coming from https://github.com/vale-cli/vale-action/issues/156

PR #1083, opened 2026-05-19
gh-1076: Use array comparison functions from `array-api-extra`

### LINKED ISSUE #1076: backend-agnostic `Compare`
### Add your issue here

Hey @paddyroddy, I am hoping that you will be able to address this note soon:

https://github.com/glass-dev/glass/blob/7e620f7202fdccd8a9021bc76348baae14ba6567/tests/fixtures/helper_classes.py#L18-L24

Draft PRs using what we would like to expose in `xpx.testing` in SciPy and sklearn are at https://github.com/scipy/scipy/pull/25143 and https://github.com/scikit-learn/scikit-learn/pull/34019. It would be cool if someone could check whether what we have right now would be sufficient for glass, or whether more is needed.

paddyroddy:
# Description

As can be seen in https://github.com/scipy/scipy/pull/25143 and https://github.com/scikit-learn/scikit-learn/pull/34019, `array-api-extra` now provides array comparison methods. This means we should be able to remove our custom `Compare` helper class.

Closes: #1076

## Changelog entry

Changed: Used `array_api_extra.testing` for array comparison in tests

## Checks


tests/core/test_harmonics.py
```diff
@@ -4,15 +4,15 @@
...
 
+from array_api_extra._lib._testing import xp_assert_close
+
 import glass.harmonics
 
 if TYPE_CHECKING:
     from types import ModuleType
 
-    from tests.fixtures.helper_classes import Compare
-
```
prady0t, line 13: Here and everywhere else, we can call `xp_assert...` inside the class itself, so `compare.assert_allclose` works but internally calls `xp_assert_close`. This would help avoid changing lines everywhere.

tests/core/test_healpix.py
```diff
@@ -152,7 +151,7 @@ def test_ang2pix(  # noqa: PLR0913
     phis = healpix_inputs.latitudes(max_phi, rng=urng)
     old = healpix.ang2pix(healpix_inputs.nside, thetas, phis, lonlat=lonlat)
     new = hp.ang2pix(healpix_inputs.nside, thetas, phis, lonlat=lonlat, xp=xp)
-    compare.assert_array_equal(old, new)
+    xp_assert_equal(xp.asarray(old), new)
```
prady0t: Here and everywhere else, instead of changing it everywhere, you can modify the compare class functions first to convert a Python scalar or list into an array, then call the array api extra function with these inputs. See here for example (done in this SciPy PR):

```py
def _convert_scalar_to_array(x, xp):
    if isinstance(x, (list, tuple)) or type(x) in (
        int,
        float,
        complex,
        bool,
    ):
        return xp.asarray(x)
    return x

def xp_assert_close(actual, desired, *, rtol=None, atol=0, check_dtype=True,
                    check_shape=True, check_0d=False, err_msg='', xp=None):
```

tests/core/test_harmonics.py
paddyroddy: Yes but the `Compare` class was always a hack, so I'd rather do the big changes once

tests/core/test_healpix.py
paddyroddy: I feel if we did this then we could have just stuck with our initial implementation

tests/benchmarks/test_fields.py
```diff
@@ -119,13 +120,12 @@ def function_to_benchmark() -> list[Any]:
     j, a, s = results[0]
     assert j is None
     assert a.shape == (0,)
-    compare.assert_allclose(xpb.asarray(s), 1.0)
+    xp_assert_close(s, xpb.ones_like(s))
```
lucascolley: the `check_shape=False` parameter should be an alternative here

tests/core/test_healpix.py
lucascolley: the advantage of converting array-likes to arrays is that you gain the strictness of checking that the return namespace is what you expect

lucascolley:
public API is now available on array-api-extra `main` and documented at https://data-apis.org/array-api-extra/api-testing.html

paddyroddy:
> public API is now available on array-api-extra `main` and documented at [data-apis.org/array-api-extra/api-testing.html](https://data-apis.org/array-api-extra/api-testing.html)

Nice! I'll change to that and wait for the release for this PR

tests/benchmarks/test_fields.py
paddyroddy: I'm not sure if that will work in this instance.

```python
type(s) # <class 'numpy.ndarray'>
print(s) # 1.0
xpx_testing.assert_close(s, 1.0) # array_namespace requires at least one non-scalar array input
xpx_testing.assert_close(s, 1.0, check_shape=False) # array_namespace requires at least one non-scalar array input
```

paddyroddy: This works though `xpx_testing.assert_close(s, xpb.asarray(1.0))`

prady0t: Setting `check_shape` to false won't be helpful if s is scalar too. We don't allow both inputs to be scalar in our api.

lucascolley: oh yeah, sorry, I forgot that we decided to require that both `actual` and `desired` be arrays. Let's persevere with that as long as there aren't any horrible cases.

prady0t: Yes. @paddyroddy  You can look at how we are dealing with this in scipy for reference [here](https://github.com/scipy/scipy/pull/25143/changes#:~:text=def%20_convert_scalar_to_array(x%2C%20xp):).

lucascolley: but I would recommend avoiding SciPy's workarounds if you can help it :)

tests/core/test_galaxies.py
```diff
@@ -275,7 +281,7 @@ def test_gaussian_phz(
 
     assert phz.__array_namespace__() == xp
     assert phz.shape == (10,)
-    compare.assert_array_equal(z, phz)
+    xpx.testing.assert_equal(z * xp.ones_like(phz), phz)
```
paddyroddy: @lucascolley I've gone through and used the `check_shape=False` idea as suggested. For this line I previously had `xpx.testing.assert_equal(z * xp.ones_like(phz), phz)`, naively I thought `xpx.testing.assert_equal(xp.asarray(z), phz, check_shape=False)` would work, but I ran into this:

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/paddy/arc/euclid/glass/.venv/lib/python3.14/site-packages/array_api_extra/testing.py", line 860, in assert_equal
    actual, desired, xp, np = _check_ns_shape_dtype(
                              ~~~~~~~~~~~~~~~~~~~~~^
        actual, desired, check_dtype, check_shape, check_scalar, xp
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/paddy/arc/euclid/glass/.venv/lib/python3.14/site-packages/array_api_extra/testing.py", line 651, in _check_ns_shape_dtype
    assert actual_size == desired_size, msg
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: sizes do not match: 1 != 10
```

or even `xpx.testing.assert_equal(xp.asarray([z]), phz, check_shape=False)` which gets the same thing. I've settled on `xpx.testing.assert_equal(xp.broadcast_to(1.0, phz.shape), phz)` which seems to work. This looks to be caused by https://github.com/data-apis/array-api-extra/blob/c784fb6ef66d07828c1d3df58c7ba9afbfca543f/src/array_api_extra/testing.py#L648-L651, is this expected behaviour?

tests/benchmarks/test_fields.py
paddyroddy: I've resolved this now, thanks both. Have raised one comment which acted slightly different to how I expected here https://github.com/glass-dev/glass/pull/1083#discussion_r3323975442.

tests/core/test_galaxies.py
lucascolley: Yes, I would recommended constructing the desired array with the desired size in this case.

Feel free to open an issue for a `check_size=False` if the pattern occurs frequently and you feel strongly!

connoraird:
Failing tests

paddyroddy:
> Failing tests

I know, that's because `array-api-extra` hasn't done a release. Can you review anyway?

tests/benchmarks/test_arraytools.py
```diff
@@ -56,7 +55,7 @@ def test_cumulative_trapezoid_1d(
     ct = benchmark(glass.arraytools.cumulative_trapezoid, f, x)
 
     # Compare to int64 as old versions of glass round to int64 if `dtype` is not passed.
-    compare.assert_allclose(
+    xpx.testing.assert_close(
```
connoraird: Should this be assert equal since these are integers? 

Same question for the other comparisons of integers.

tests/benchmarks/test_lensing.py
```diff
@@ -38,8 +38,8 @@ def test_multi_plane_matrix(
     mat = glass.multi_plane_matrix(shells, cosmo)
     deltas = urngb.random((len(shells), 10))
 
-    compare.assert_array_equal(mat, xpb.tril(mat))
-    compare.assert_array_equal(xpb.triu(mat, k=1), 0)
+    xpx.testing.assert_equal(mat, xpb.tril(mat))
+    xpx.testing.assert_equal(xpb.triu(mat, k=1), xpb.asarray(0.0), check_shape=False)
```
connoraird, line 42: Should this be `assert_close` since it is a float?

Same question for other float comparisons.

tests/core/grf/test_transformations.py
```diff
@@ -2,26 +2,23 @@
...
 
-def test_normal(
-    compare: type[Compare],
-    urng: UnifiedGenerator,
-) -> None:
+def test_normal(urng: UnifiedGenerator) -> None:
     t = glass.grf.Normal()
     x = urng.standard_normal(10)
-    compare.assert_array_equal(t(x, 1.0), x)
+    xpx.testing.assert_equal(t(x, 1.0), x)
```
connoraird, line 18: Does this need to be `assert_close`? Same below.

tests/core/test_fields.py
```diff
@@ -542,27 +520,27 @@ def test_enumerate_spectra(
...
-def test_spectra_indices(
-    compare: type[Compare],
-    xp: ModuleType,
-) -> None:
-    compare.assert_array_equal(glass.spectra_indices(0), xp.zeros((0, 2)))
-    compare.assert_array_equal(glass.spectra_indices(0, xp=xp), xp.zeros((0, 2)))
-    compare.assert_array_equal(glass.spectra_indices(1, xp=xp), [[0, 0]])
-    compare.assert_array_equal(
+def test_spectra_indices(xp: ModuleType) -> None:
+    xpx.testing.assert_equal(glass.spectra_indices(0), np.zeros((0, 2), dtype=np.int64))
```
connoraird, line 510: Could we add a comment here to explain that we're testing that `glass.spectra_indices` defaults to NumPy? At first, I was confused why we needed NumPy.

tests/core/test_galaxies.py
```diff
@@ -281,7 +281,7 @@ def test_gaussian_phz(
 
     assert phz.__array_namespace__() == xp
     assert phz.shape == (10,)
-    compare.assert_array_equal(z, phz)
+    xpx.testing.assert_equal(xp.broadcast_to(xp.asarray(1.0), phz.shape), phz)
```
connoraird, line 284: `assert_close`?

tests/core/test_healpix.py
```diff
@@ -201,28 +199,27 @@ def test_get_nside(
...
+        pol=pol,
+        use_pixel_weights=use_pixel_weights,
+    )
+    new = hp.map2alm(
+        kappa,
+        lmax=healpix_inputs.lmax,
+        pol=pol,
+        use_pixel_weights=use_pixel_weights,
     )
+    xpx.testing.assert_equal(xp.asarray(old), new)
```
connoraird, line 222: Just a reminder to sanity check these should be `assert_equal` rather than `assert_close`

tests/core/test_healpix.py
```diff
@@ -294,12 +285,11 @@ def test_pixwin(
 
     assert len(old) == len(new)
     for i in range(len(old)):
-        compare.assert_array_equal(old[i], new[i])
+        xpx.testing.assert_equal(xp.asarray(old[i], dtype=xp.float64), new[i])
```
connoraird, line 288: `assert_close`?

Same below?

tests/core/test_points.py
```diff
@@ -9,6 +9,7 @@
 
 import glass
 import glass.healpix as hp
+import glass.points
```
connoraird, line 12: Do we still need `import glass` if we have this?

connoraird (changes requested):
I stopped writing the same comment after a while but I think double checking the types of what we are comparing and making sure it should be `assert_close` or `assert_equal` would be good.

tests/benchmarks/test_arraytools.py
paddyroddy: Good spot. Was focusing on just replacements here, but makes sense to adjust.

tests/benchmarks/test_lensing.py
paddyroddy: I disagree with this one. Whilst we are comparing floats, we are comparing `0.0` with `0.0`. We are allowing potential failures if we change this to `assert_close`.

tests/core/grf/test_transformations.py
paddyroddy: Again I don't see a benefit of changing this

tests/core/test_fields.py
paddyroddy: That confused me too, so sure

tests/core/test_galaxies.py
paddyroddy: Same again

tests/core/test_points.py
paddyroddy: We still need it for `effective_bias` etc.

paddyroddy:
I've done a find-and-replace and then iterated to include more `assert_equal` calls. But I don't agree that because something is `float` that we should necessarily use `assert_close`. If you consider [the NumPy implementation](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) it's to do with equality up to a tolerance. If that tolerance is `0`, then they may as well be equal.

paddyroddy:
Have merged in #1089

lucascolley:
exams done, please do give https://github.com/data-apis/array-api-extra/releases/tag/v0.10.2 a test!

lucascolley:
oops, made it to PyPI this time https://pypi.org/project/array-api-extra/

connoraird (approved):

paddyroddy:
Merging now. Hope the exams went well @lucascolley!

PR #1089, opened 2026-05-21
gh-1085: refactor iternorm

### LINKED ISSUE #1085: Create JAX implementation of `iternorm`
### Add your issue here

Discussed in #999, we are going to have two `iternorm` implementations. One for JAX and one for everything else.

ntessore:
# Description

An attempt to

* simplify `iternorm()` in a way that makes it clear this is the same algorithm as in Appendix A of [arXiv:2302.01942](https://arxiv.org/pdf/2302.01942),
* add tests that check for correctness of the algorithm, and
* make `iternorm()` Array API compatible.

The function now takes only an iterable of covariance matrix rows. It will figure out the correct shape of the output from the input. It will also figure out the number of correlations to keep from each covariance matrix row individually, which means that the function can now dynamically forget previous correlations.

The function now yields a single scaling vector that can be applied to the stack of standard normal random variates to get the correctly correlated normal variate.

Closes: #1085

## Changelog entry

Changed: simplified `glass.iternorm()` that is now Array API compatible

## Checks

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -6,7 +6,7 @@ import itertools
 import math
 import sys
 import warnings
-from collections import deque
+import collections
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
@@ -386,7 +386,7 @@ def _generate_grf(
     z_size = n * (n + 1) // 2
 
     # store standard normal random variates
-    y = deque()
+    y = collections.deque()
 
     # sample the Gaussian fields iteratively
     for w in iternorm(cov):
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -9,7 +9,6 @@ import warnings
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
-import numpy as np
 import transformcl
 
 import array_api_compat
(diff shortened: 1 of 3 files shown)


ntessore:
So far I have refactored `iternorm()` in a way that should work with all Array API backends. The cyclic overwriting of rows is gone from the code; instead the internal matrix is rebuilt as per the paper.

I am getting a ton of `ty` errors locally, which come from parts of the code I haven't touched.

ntessore:
@paddyroddy @connoraird This is ready for pre-review. I don't know how much the repeated `xp.concat` will affect performance, considering all of the other heavy stuff that goes into a simulation. There might also be a better way to assemble the matrix. Happy to hear your suggestions.

glass/fields.py
```diff
@@ -91,94 +91,79 @@ def nfields_from_nspectra(nspectra: int) -> int:
...
+    for i, row in enumerate(cov):
+        # number of correlations, this will determine matrix size
+        k = row.shape[-1] - 1
+        if k < 0:
+            raise ValueError("empty covariance matrix")
+
+        # check for first iteration
+        if i == 0:
+            # extract input array backend
+            xp = row.__array_namespace__()
```
connoraird, line 125: I think we should be able to move this out of the loop by adding this at the start of `iternorm`. Seems to be faster as well after some very basic testing

```py
xp = array_api_compat.array_namespace(*cov, use_compat=False)
```

ntessore: But this consumes `cov`, which will generally be an iterator, not a sequence. We can go back to peeking the first element of `cov` but then we have to handle the case where `cov` is empty.

glass/fields.py
```diff
@@ -6,10 +6,10 @@
 import math
 import sys
 import warnings
+from collections import deque
```
paddyroddy: Can we change this to an import? If I were to read `deque` I would assume it was one of our functions. This is different to `Sequence` which is capitalised and a standard type so it gives me some idea.

```suggestion
import collections
```

glass/fields.py
```diff
@@ -91,94 +91,88 @@ def nfields_from_nspectra(nspectra: int) -> int:
...
-        The vector for iterative sampling.
-    standard_deviation
-        The standard deviation for iterative sampling.
-
-    Raises
-    ------
-    TypeError
-        If the covariance matrix is not broadcastable to the given size.
-    ValueError
-        If the covariance matrix is not positive definite.
```
paddyroddy, line 126: We still have a `ValueError`. Can this be reinstated?

paddyroddy: In fact we now have various `ValueError`s

glass/fields.py
```diff
@@ -388,30 +382,30 @@ def _generate_grf(
...
 
-    # working arrays for the iterative sampling
+    # alm size
     z_size = n * (n + 1) // 2
-    y = xp.zeros((z_size, ncorr), dtype=xp.complex128)
 
-    # generate the conditional normal distribution for iterative sampling
-    conditional_dist = iternorm(ncorr, cov, size=n)
+    # store standard normal random variates
+    y = deque()
```
paddyroddy:
```suggestion
    y = collections.deque()
```

glass/fields.py
```diff
@@ -388,30 +382,30 @@ def _generate_grf(
...
-        for i in range(ncorr):
-            alm += glass.harmonics.multalm(y[:, i], a[:, i])
+        # how many correlations have missing variates
+        mis = w.shape[-1] - len(y)
 
-        # store the standard normal in y array at the indicated index
-        if j is not None:
-            y = xpx.at(y)[:, j].set(z)
+        # compute the correlated variate
+        alm = sum(glass.harmonics.multalm(z, w[..., i + mis]) for i, z in enumerate(y))
```
paddyroddy, line 415: Given that `multalm` is meant to return a `ComplexArray`

```suggestion
        alm = xp.sum(glass.harmonics.multalm(z, w[..., i + mis]) for i, z in enumerate(y))
```

tests/benchmarks/test_fields.py
```diff
@@ -25,25 +25,17 @@ def test_iternorm_no_size(
...
 
-    results = benchmark(function_to_benchmark)
-    j, a, s = results[0]
-
-    assert isinstance(j, int)
-    assert a.shape == (k,)
-    assert s.shape == ()
-    assert s.dtype == xpb.float64
-    assert s.shape == ()
+    benchmark(function_to_benchmark)
```
paddyroddy: I believe @connoraird put these in just to make sure the benchmark isn't really fast because it didn't run

tests/core/test_fields.py
```diff
@@ -26,117 +26,91 @@ def not_triangle_numbers() -> list[int]:
...
-    size = (3,)
-
-    generator = glass.iternorm(
-        k,
-        (
-            xp.asarray(arr)
-            for arr in [
+@pytest.mark.parametrize("k", [0, 1, 2], ids=["k=0", "k=1", "k=2"])
+@pytest.mark.parametrize("test_nd", [False, True], ids=["0d", "nd"])
+def test_iternorm(k, test_nd, xp: ModuleType, compare: type[Compare]) -> None:
```
paddyroddy: Would be great to include types and be consistent with other tests
```suggestion
def test_iternorm(
    compare: type[Compare],
	k: int,
	test_nd: bool,
	xp: ModuleType,
) -> None:
```

tests/core/test_fields.py
```diff
@@ -26,117 +26,91 @@ def not_triangle_numbers() -> list[int]:
...
+        cov[..., i, i::-1][..., : min(i, k) + 1]
+        for i in range(cov.shape[-1])
+    )
 
-    assert j == 1
-    assert a.shape == (3, 2)
-    assert s.shape == (3,)
+    # check rows against explicit calculation
+    # see https://arxiv.org/pdf/2302.01942, Appendix A
+    # ruff: disable[N806]
```
paddyroddy, line 74: Is there a reason why you'd be against `sn` or `s_n`?

paddyroddy (commented):
Few comments

glass/fields.py
ntessore: I guess it depends on your interpretation of

> This section should be used judiciously, i.e., only for errors that are non-obvious or have a large chance of getting raised.

I can add a general comment that this will be raised for invalid covariance matrices.

glass/fields.py
ntessore: No, it is done on purpose like this. It should be as if calling `+` on each item in the generator expression. `xp.sum()` would first construct the full array of intermediate values, then add them together, which is unnecessarily large.

tests/core/test_fields.py
ntessore: I dislike the capitals as much as you do, but unfortunately we have both $A_n$ and $a_n$ in the paper, and I wanted to match the notation exactly.

tests/benchmarks/test_fields.py
connoraird: Yes exactly. Simple assertions which are unlikely to change (i.e. to fail the benchmark tests) are important to make sure we're actually running the scenario we think we are.

glass/fields.py
connoraird: Ah, okay. I thought enumerate fully consumed a generator. I'll do some research.

connoraird: I've just checked an enumerate doesn't consume the generator so your point makes sense.

Should we then be passing cov as a generator in a test?

tests/benchmarks/test_fields.py
```diff
@@ -69,57 +59,31 @@ def test_iternorm_specify_size(
...
-    array_in = [xpb.stack([x]) for x in xpb.ones(1_000, dtype=xpb.float64)]
-
-    def function_to_benchmark() -> list[Any]:
-        generator = glass.iternorm(k, iter(array_in))
-        return generator_consumer.consume(generator)
+    array_in = [xpb.asarray([x]) for x in xpb.ones(1_000, dtype=xpb.float64)]
 
-    results = benchmark(function_to_benchmark)
+    def function_to_benchmark() -> None:
+        generator = glass.iternorm(array_in)
```
connoraird, line 73: For example, one of these tests should pass `array_in` as a generator i.e. `iter(array_in)`

glass/fields.py
ntessore: Yep, I am passing a generator expression in `test_iternorm()` for that purpose.

tests/benchmarks/test_fields.py
ntessore: I put the generator test into the unit tests because I thought it's better to only test performance in the benchmark part.

glass/fields.py
paddyroddy: Where did you get that from? We have just been consistent with the rest of the code base. We should standardise our approach.

glass/fields.py
paddyroddy: Makes sense

tests/core/test_fields.py
paddyroddy: Fair enough

tests/core/test_fields.py
```diff
@@ -26,117 +26,96 @@ def not_triangle_numbers() -> list[int]:
...
+            [
+                [1.0, 0.2, 0.1],
+                [0.2, 0.5, 0.2],
+                [0.1, 0.2, 0.3],
+            ],
+        )
 
-    j, a, s = result
+    generator = glass.iternorm(
+        # for each covariance row, start at the diagonal diagonal and go left
```
connoraird:
```suggestion
        # for each covariance row, start at the diagonal and go left
```

connoraird (changes requested):
One small comment and needs merge conflicts resolving.

tests/core/test_fields.py
```diff
@@ -26,117 +26,96 @@ def not_triangle_numbers() -> list[int]:
...
+        compare.assert_allclose(An.mT @ An, Sninv)
+        an = (An @ cn[..., None])[..., 0]
+        sn = xp.sqrt(vn - xp.vecdot(an, an))
+        # make sure a only has k correlations
+        assert w.shape[-1] <= min(n, k) + 1
+        # split scaling vector up into a and s
+        a, s = w[..., :-1], w[..., -1]
+        # cannot compare a directly, only a^T @ a
+        compare.assert_allclose(xp.vecdot(a, a), xp.vecdot(an, an))
+        compare.assert_allclose(s, sn)
```
paddyroddy, line 102: This seems quite complex logic for a test, perhaps it could be its own function?

paddyroddy (approved):
Happy as it is, one comment otherwise

tests/core/test_fields.py
ntessore: Yeah, it's quite intricate, so I could definitely move it to a function. That being said, the function would
* need some sort of consistent typing between `xp` and the three arrays it would receive as input, which gave me a headache just now;
* be used exactly once in the code, so it's maybe in the end making things more complicated instead of less?

ntessore:
@paddyroddy @connoraird Making the complicated test prettier seems too much pain for too little gain so I'd propose it like it is

connoraird (approved):

tests/core/test_fields.py
paddyroddy: Yeah, may not be worth it

PR #1091, opened 2026-05-21
gh-1090: move `ty` from `pre-commit` to `nox`

### LINKED ISSUE #1090: Remove `ty` from `pre-commit` and place into `nox`
### Add your issue here

Request from @ntessore, to avoid confusing contributors

paddyroddy:
# Description

Developers currently have to have everything installed which might be disruptive. This solution means that we no longer run `ty` via `prek` explicitly, but rather run it through `nox -s lint`. CI uses `nox` so it will be unchaged there.

Closes: #1090

## Changelog entry

Changed: `ty` is now run from `nox` rather than `pre-commit`

## Checks

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -20,7 +20,7 @@ concurrency:
 jobs:
   prek:
     if: github.event.pull_request.draft == false
-    name: Run prek
+    name: Run linting
     permissions:
       contents: read
     runs-on: ubuntu-slim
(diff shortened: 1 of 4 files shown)


ntessore:
Thanks @paddyroddy. Looks great. Would it make sense to have ty in a separate "Linting/Type checking" test?

paddyroddy:
> Looks great. Would it make sense to have ty in a separate "Linting/Type checking" test?

I feel like all being part of the same `lint` job makes more sense to me. Ideally this would be part of a single `prek` call, but that causes dependencies issues as discussed.

ntessore (approved):

PR #1092, opened 2026-05-27
gh-1087: Rename benchmarks dir to regression

### LINKED ISSUE #1087: Rename benchmark tests to regressions
# Description

In preparation for #1086, we should rename the benchmark tests to regressions and remove anything to do with running them as a single benchmark as they are only intented to be used for regression testing.

connoraird:
# Description

Updates all mention of the existing benchmark tests to use the term regression tests.

Closes: #1087 

## Changelog entry

Changed: Renamed the benchmarks tests to regression tests

## Checks


.github/workflows/test.yaml
```diff
@@ -18,13 +18,13 @@ concurrency:
     ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
 
 env:
-  BENCH_COV_NAME: cov-bench
+  REGRESSION_COV_NAME: cov-regression
```
paddyroddy: I made the old one `BENCH` rather than `BENCHMARK` so that it was roughly the same length as `CORE`. Could we use, e.g. `REG` throughout?

CONTRIBUTING.md
```diff
@@ -252,46 +252,34 @@ uv run nox -s docs -- serve
...
 # Regression tests
 uv run nox -s regression-tests -- <initial-state-revision> \
 <revision-to-compare> -k test_fields
 ```
 
 <!-- prettier-ignore -->
 > [!TIP]
-> Benchmark tests should do minimal assertions on what is returned to
-> make sure that the function is working as expected. The main goal of benchmark
+> Regression tests should do minimal assertions on what is returned to
```
paddyroddy, line 279: Have we clarified that point with Nicolas? There's been pushback

noxfile.py
```diff
@@ -26,9 +26,9 @@
     "array_api_strict": "array-api-strict>=2",
     "jax": "jax>=0.4.32",
 }
-BENCH_TESTS_LOC = pathlib.Path("tests/benchmarks")
+REGRESSION_TESTS_LOC = pathlib.Path("tests/regression")
```
paddyroddy: Same comment here about length of `BENCH`

paddyroddy (changes requested):
A few small comments

CONTRIBUTING.md
connoraird: @ntessore Do you agree with the statement in this file? 

I think perhaps this could be a separate issue?

paddyroddy: Sure if no reply then let's do that

.github/workflows/test.yaml
```diff
@@ -18,13 +18,13 @@ concurrency:
     ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
 
 env:
-  BENCH_COV_NAME: cov-bench
+  REG_COV_NAME: cov-regression
```
paddyroddy: I meant throughout on this, i.e. `cov-reg`, `tests-reg`, etc.

paddyroddy (commented):
Small request

CONTRIBUTING.md
ntessore: I think we agreed that benchmarks should verify that the expected amount of work has been done?

paddyroddy (approved):
Cheers

Issue #1088, opened 2026-05-20
Add realistic benchmarks


connoraird:
# Description

Add a new benchmarks test dir which contains larger, more realistic examples similar to those in the examples dir.

## Tasks

- [ ] We should start with the `1-basic/lensing.ipynb` example
- [ ] These benchmarks should be able to be run locally and on Archer2

PR #1125, opened 2026-06-08
gh-1124: Allow running glass offline

### LINKED ISSUE #1124: Allow benchmarking without internet
### Describe the Bug

Soem `healpy` functions such as `healpy.pixwin` and `healpy.map2alm` require an interenter connection to pull down pixel weight data files. However, to allow benchmarking on Archer2, we need to prevent HTTP calls happening mid execution. Therefore, we should add a global option to define a path to the `healpy` pixel weight data files which we then pass to `healpy` to prevent the download.

### To Reproduce

Run the `test_discretized_cls` core test without internet and it will fail.

```sh
# Clear any downloaded cache
rm -rf $HOME/.astropy/cache/download/url

# Run the test
uv venv
uv sync --group test
source .venv/bin/activate
pytest tests/core -k test_discretized_cls
```

### Expected Behaviour

Test should pass without internet

### Actual Behaviour

```sh
$ pytest tests/core -k test_discretized_cls
...
else:  # No success
            if not sources:
                raise KeyError(
                    f"No sources listed and file {remote_url} not in cache! "
                    "Please include primary URL in sources if you want it to be "
                    "included as a valid source."
                )
            elif len(sources) == 1:
                raise errors[sources[0]]
            else:
>               raise urllib.error.URLError(
                    f"Unable to open any source! Exceptions were {errors}"
                ) from errors[sources[0]]
E               urllib.error.URLError: <urlopen error Unable to open any source! Exceptions were {'https://healpy.github.io/healpy-data/pixel_window_functions/pixel_window_n0004.fits': URLError(gaierror(8, 'nodename nor servname provided, or not known. requested URL: https://healpy.github.io/healpy-data/pixel_window_functions/pixel_window_n0004.fits')), 'http://www.astropy.org/astropy-data/pixel_window_functions/pixel_window_n0004.fits': URLError(gaierror(8, 'nodename nor servname provided, or not known. requested URL: https://healpy.github.io/healpy-data/pixel_window_functions/pixel_window_n0004.fits'))}>

.venv/lib/python3.14/site-packages/astropy/utils/data.py:1586: URLError
============================================================================================================================================ short test summary info =============================================================================================================================================
SKIPPED [1] tests/core/test_jax.py:3: tests require jax
FAILED tests/core/test_fields.py::test_discretized_cls[numpy] - urllib.error.URLError: <urlopen error Unable to open any source! Exceptions were {'https://healpy.github.io/healpy-data/pixel_window_functions/pixel_window_n0004.fits': URLError(gaierror(8, 'nodename nor servname provided, or not known. requested URL: https://healpy.github.io/healpy-data/pixel_window...
================================================================================================================================== 1 failed, 1 skipped, 165 deselected in 0.22s ==================================================================================================================================
```

### Version In Use

2026.2.dev42+ga77d3e033

### Additional Context

```markdown
- Python version: 3.14
- Operating system: MacOS
```

connoraird:
# Description

Removes the requirement for glass to be ran with internet access by allowing a local path to healpy datafile to be specified through an environment variable

Fixes: #1124 

## Changelog entry

Fixed: An internet connection is no longer required to run glass.

## Checks

### DIFF
--- a/glass/healpix.py
+++ b/glass/healpix.py
@@ -22,9 +22,11 @@ if TYPE_CHECKING:
     from glass._types import ComplexArray, DTypeLike, FloatArray, IntArray
 
 
-HEALPY_DATAPATH = os.environ.get("HEALPY_DATAPATH")
-if HEALPY_DATAPATH is not None and not pathlib.Path(HEALPY_DATAPATH).is_dir():
-    raise ValueError(f"Healpy datapath not found at '{HEALPY_DATAPATH}'")
+def _get_healpy_datapath() -> str | None:
+    healpy_datapath = os.environ.get("HEALPY_DATAPATH")
+    if healpy_datapath is not None and not pathlib.Path(healpy_datapath).is_dir():
+        raise ValueError(f"Healpy datapath not found at '{healpy_datapath}'")
+    return healpy_datapath
 
 
 def alm2map(  # noqa: PLR0913
@@ -295,7 +297,7 @@ def map2alm(
     return xp.asarray(
         healpy.map2alm(
             inputs,
-            datapath=HEALPY_DATAPATH,
+            datapath=_get_healpy_datapath(),
             lmax=lmax,
             pol=pol,
             use_pixel_weights=use_pixel_weights,
@@ -375,7 +377,7 @@ def pixwin(
     """
     xp = _utils.default_xp() if xp is None else xp
 
-    output = healpy.pixwin(nside, datapath=HEALPY_DATAPATH, lmax=lmax, pol=pol)
+    output = healpy.pixwin(nside, datapath=_get_healpy_datapath(), lmax=lmax, pol=pol)
     return (
         tuple(xp.asarray(out, dtype=xp.float64) for out in output)
         if pol
--- a/glass/healpix.py
+++ b/glass/healpix.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import os
+import pathlib
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
@@ -18,6 +20,13 @@ if TYPE_CHECKING:
     from types import ModuleType
 
     from glass._types import ComplexArray, DTypeLike, FloatArray, IntArray
+
+
+def _get_healpy_datapath() -> str:
+    healpy_datapath = os.environ.get("HEALPY_DATAPATH")
+    if healpy_datapath is None or pathlib.Path(healpy_datapath).is_dir():
+        raise ValueError(f"Healpy datapath not found at '{healpy_datapath}'")
+    return healpy_datapath
 
 
 def alm2map(  # noqa: PLR0913
@@ -274,8 +283,15 @@ def map2alm(
         if isinstance(maps, Sequence)
         else np.asarray(maps)
     )
+
     return xp.asarray(
-        healpy.map2alm(inputs, lmax=lmax, pol=pol, use_pixel_weights=use_pixel_weights),
+        healpy.map2alm(
+            inputs,
+            datapath=_get_healpy_datapath(),
+            lmax=lmax,
+            pol=pol,
+            use_pixel_weights=use_pixel_weights,
+        ),
     )
 
 
@@ -341,7 +357,7 @@ def pixwin(
     """
     xp = _utils.default_xp() if xp is None else xp
 
-    output = healpy.pixwin(nside, lmax=lmax, pol=pol)
+    output = healpy.pixwin(nside, datapath=_get_healpy_datapath(), lmax=lmax, pol=pol)
     return (
         tuple(xp.asarray(out, dtype=xp.float64) for out in output)
         if pol
--- a/glass/healpix.py
+++ b/glass/healpix.py
@@ -22,11 +22,9 @@ if TYPE_CHECKING:
     from glass._types import ComplexArray, DTypeLike, FloatArray, IntArray
 
 
-def _get_healpy_datapath() -> str | None:
-    healpy_datapath = os.environ.get("HEALPY_DATAPATH")
-    if healpy_datapath is not None and not pathlib.Path(healpy_datapath).is_dir():
-        raise ValueError(f"Healpy datapath not found at '{healpy_datapath}'")
-    return healpy_datapath
+HEALPY_DATAPATH = os.environ.get("HEALPY_DATAPATH")
+if HEALPY_DATAPATH is not None and not pathlib.Path(HEALPY_DATAPATH).is_dir():
+    raise ValueError(f"Healpy datapath not found at '{HEALPY_DATAPATH}'")
 
 
 def alm2map(  # noqa: PLR0913
@@ -297,7 +295,7 @@ def map2alm(
     return xp.asarray(
         healpy.map2alm(
             inputs,
-            datapath=_get_healpy_datapath(),
+            datapath=HEALPY_DATAPATH,
             lmax=lmax,
             pol=pol,
             use_pixel_weights=use_pixel_weights,
@@ -377,7 +375,7 @@ def pixwin(
     """
     xp = _utils.default_xp() if xp is None else xp
 
-    output = healpy.pixwin(nside, datapath=_get_healpy_datapath(), lmax=lmax, pol=pol)
+    output = healpy.pixwin(nside, datapath=HEALPY_DATAPATH, lmax=lmax, pol=pol)
     return (
         tuple(xp.asarray(out, dtype=xp.float64) for out in output)
         if pol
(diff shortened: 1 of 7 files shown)


glass/healpix.py
```diff
@@ -246,6 +255,16 @@ def map2alm(
...
 
+    If you are running in an offline environment, you must provide a datapath to local
+    healpy datafiles. To download these files:
+
+        git clone --depth 1 https://github.com/healpy/healpy-data
+        cd healpy-data
+        bash download_weights_8192.sh
+
+    and set datapath to the root of the repository. If this method is used, the only
+    supported values of nside are 2^n for n in the range 5-13.
```
paddyroddy, line 266: Do we think we need to document this anywhere else or not?

paddyroddy (commented):
Is there a way we could explicitly test this? A situation where we pretend not to have internet?

glass/healpix.py
connoraird: I have tested this locally, but do you mean through CI?

paddyroddy: Yes, like a unit test or something

connoraird: > Do we think we need to document this anywhere else or not?

Not sure where else we would document it. 

> like a unit test or something

Not sure how to turn off the internet connection in pytest 🤔 We could probably set that env var in a test which won't be no internet but will test the env var functionality. We'd have to pull the data though.

connoraird: Would we want to do something like pulling the data from the healpy-data github mid test, only if we have an internet connection? Seems weird that to test not having an internet connection, we would need an internet connection.

connoraird: I've now added some tests which pull this data before using the env vars.

tests/fixtures/healpy_data.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Pull the healpy data and returns the path to the new folder.
+
+    Also performs cleanup by removing the cloned dir.
+    """
+    subprocess.run(  # noqa: S602
+        """
+        git clone --depth 1 https://github.com/healpy/healpy-data;
+        cd healpy-data;
+        bash download_weights_8192.sh;
```
paddyroddy: Do we need/benefit from this?

tests/core/test_healpix.py
```diff
@@ -190,14 +190,51 @@ def test_get_nside(
 
 
 @pytest.mark.parametrize(
```
paddyroddy: Is there much benefit of a `parametrize` with only one input?

tests/core/test_healpix.py
```diff
@@ -190,14 +190,51 @@ def test_get_nside(
...
     [
-        (False, False),
-        (False, True),
-        (True, False),
-        (True, True),
+        (False, (78,)),
     ],
 )
+def test_map2alm_with_pulled_data(
+    add_healpy_datapath_to_env: pytest.FixtureDef,  # noqa: ARG001
```
paddyroddy: We've usually typed these cases like this:

```suggestion
    add_healpy_datapath_to_env: Generator,  # noqa: ARG001
```

Did you go for `pytest.FixtureDef` as it's a bit clearer what is going on?

paddyroddy: Does this make sense to be a fixture? Would it make more sense as a decorator? Or does it not matter? Referring to [ARG001](https://docs.astral.sh/ruff/rules/unused-function-argument/).

tests/core/test_healpix.py
```diff
@@ -190,14 +190,51 @@ def test_get_nside(
...
+    expected_shape: tuple[int],
+    healpix_inputs: type[HealpixInputs],
+    pol: bool,  # noqa: FBT001
+    urng: UnifiedGenerator,
+) -> None:
+    """Tests running map2alm offline works as expected."""
+    kappa = urng.normal(size=12_288)  # nside=32
```
paddyroddy: Or even use 

```python
npix = healpy.nside2npix(32)
```


```suggestion
    npix = 12 * 32 ** 2
    kappa = urng.normal(size=npix)
```

tests/core/test_healpix.py
```diff
@@ -190,14 +190,51 @@ def test_get_nside(
...
+        lmax=healpix_inputs.lmax,
+        pol=pol,
+        use_pixel_weights=True,
+    )
+    assert result.shape == expected_shape
+
+
+@pytest.mark.parametrize("pol", [False, True])
+def test_map2alm_with_pulled_data_wrong_path(
+    invalid_healpy_datapath: str,
```
paddyroddy, line 211: Based on what I said above about typing fixtures, here you've gone with `str`

tests/core/test_healpix.py
```diff
@@ -190,14 +190,51 @@ def test_get_nside(
...
+    healpix_inputs: type[HealpixInputs],
+    pol: bool,  # noqa: FBT001
+    urng: UnifiedGenerator,
+) -> None:
+    """Tests running map2alm offline incorrectly doesn't fallback to a HTTP request."""
+    kappa = urng.normal(size=12_288)  # nside=32
```
paddyroddy: Same `npix` comment

paddyroddy (commented):
A few thoughts

tests/fixtures/healpy_data.py
connoraird: Definitely, the first two lines. ~The last line, it depends if a user would likely use nside=8192~ You were right, removed it

tests/core/test_healpix.py
connoraird: Nope, will remove

tests/core/test_healpix.py
connoraird: Ah okay, not sure. I'm just not familiar with decorators but I can have a look into that. I need the teardown feature. Need to check decorators allow that.

tests/core/test_healpix.py
connoraird: Because we use this value.

tests/core/test_healpix.py
connoraird: Yes, I thought it made more sense. Plus we never use the value anyway. The only reason it's a generator is because I call yield to run the teardown steps

tests/core/test_healpix.py
paddyroddy: Oh I see, I can see some logic behind that. Although, of course, you usually "use" fixtures.

tests/core/test_healpix.py
connoraird: I think using a fixture might be better because I am using a different scope for the `healpy_datapath` fixture to make it only run once per session.

paddyroddy: Should we prefix it with `_` as [suggested by ruff](https://docs.astral.sh/ruff/rules/unused-function-argument/)?

connoraird: Yep, sure

tests/core/test_healpix.py
```diff
@@ -189,15 +189,47 @@ def test_get_nside(
...
-@pytest.mark.parametrize(
-    ("pol", "use_pixel_weights"),
-    [
-        (False, False),
-        (False, True),
-        (True, False),
-        (True, True),
-    ],
-)
+@pytest.mark.usefixtures("_add_healpy_datapath_to_env")
```
paddyroddy, line 192: Ah, yeah I much prefer this 👍 think that's what you're _meant_ to do.

tests/core/test_healpix.py
```diff
@@ -189,15 +189,47 @@ def test_get_nside(
...
+def test_map2alm_with_pulled_data(
+    healpix_inputs: type[HealpixInputs],
+    urng: UnifiedGenerator,
+) -> None:
+    """Tests running map2alm offline works as expected."""
+    npix = healpy.nside2npix(32)
+    kappa = urng.normal(size=npix)
+    result = hp.map2alm(
```
paddyroddy, line 200: This looks odd for anyone reading, but seeing as we're testing the behaviour of `import glass.healpix as hp` here, it makes sense to only call that for the method involved, in this case `map2alm`.

tests/fixtures/healpy_data.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Pull the healpy data and returns the path to the new folder.
+
+    Also performs cleanup by removing the cloned dir.
+    """
+    subprocess.run(  # noqa: S602
+        """
+        git clone --depth 1 https://github.com/healpy/healpy-data;
+        cd healpy-data;
+        """,  # noqa: S607
```
paddyroddy, line 27: I realised the confusion I cost before. Every time I've use `subprocess.run()` syntax I've used `subprocess.run(["a", "b", "c"])` rather than `subprocess.run("abc")` syntax. I thought this was just a comment.

paddyroddy (approved):
Looks great

connoraird:
I have decided to merge even with the failing examples so I can continue to make progress on #1088 and we can investigate any persistent failures later.

PR #1132, opened 2026-06-10
gh-1131: Attempt to make reg tests stable

### LINKED ISSUE #1131: Regression tests are flaky again
### Add your issue here

# Description

We are getting consistent regression test failure for the tests `test_generate_grf`, `test_generate` and `test_galaxy_shear`. We should therefore, attempt to make these tests stable or label as unstable.

connoraird:
# Description

- Marks some tests as unstable
- Increases the number of achieved iteration in `test_generate[_grf]` tests
- Forces reg tests to be triggered if the tests themselves are updated

Closes: #1131

## Checks


tests/regression/test_fields.py
```diff
@@ -124,8 +124,8 @@ def test_generate_grf(  # noqa: PLR0913
     """Regression tests of glass.fields._generate_grf with positional arguments only."""
     n = 100
     nth_triangular_number = int((n * (n + 1)) / 2)
-    gls: AngularPowerSpectra = [xpb.ones(10) for _ in range(nth_triangular_number)]
-    nside = 16
+    gls: AngularPowerSpectra = [xpb.zeros(10) for _ in range(nth_triangular_number)]
+    nside = 32
```
paddyroddy, line 131: Why changing from `ones` to `zeros`?

connoraird: This prevents the exception "covariance matrix is not positive definite" from being raised, thus allowing us to consume the full generator.

paddyroddy: Why weren't we doing that before?

connoraird: I'm not sure, I guess it depends if passing all zeros simulates a real situation worth benchmarking?

paddyroddy: It doesn't sound obvious. How long does it take? Should we consult Nicolas on this?

connoraird: Probably. @ntessore. Does making gls all zeros make this benchmark not realistic?

ntessore: No, it will produce Gaussian random draws which are all zero, but the amount of work done is still the same.

Another option is

    [xpb.ones(10) if i == j else xpb.zeros(10) for i in range(10) for j in range(i, -1, -1)]

connoraird: Okay, So then I guess we are okay to merge? (assuming an approving review)

paddyroddy: I think I'd prefer Nicolas' suggestion as it reflects a real world case

connoraird: I've updated to use the new example

paddyroddy (approved):

paddyroddy:
Still got
```
test_generate[numpy-1] (0001_6d0144c) - Field 'mean' has failed PercentageRegressionCheck: 7.222508142 > 5.000000000
```
in #1133

PR #1137, opened 2026-06-16
gh-1021: Add decorator for np conversions

### LINKED ISSUE #1021: Decorator for input output conversion array-api porting

> Created from this [discussion](https://github.com/glass-dev/glass/pull/979/changes#r2736258232).

There are several functions which have proven difficult to port to the array api. When this is the case we have resorted to simply converting inputs to Numpy and outputs back to the input backend.

It would be good to have this functionality added as a decorator which would prevent us from having to manually convert inputs and outputs.

prady0t:
# Description

Adds a NumPy decorator to perform the internal NumPy conversion and change back.

Closes: #1021

## Checks


glass/_array_api_utils.py
```diff
@@ -559,3 +562,53 @@ def tril_indices(
         # If any other backend use default
         dxp = default_xp(xp.__name__)
         return tuple(xp.asarray(arr) for arr in dxp.tril_indices(n, k=k, m=m))
+
+def numpy_fallback(func):
```
prady0t: @lucascolley This could be more generalised for the standard. Currently, I feel the limitation is the inability to specify parameters such as `dtype` for xp.asarray. 

@ntessore @paddyroddy Would love some feedback.

lucascolley: cc @betatim, I wonder whether there is any overlap between this and https://github.com/scikit-learn/scikit-learn/pull/34324 which would be nice to upstream into `xpx.numpy`.

ntessore: This is pretty much out of my purview, and @paddyroddy is on leave currently, so it might be a bit before he can answer.

In my opinion, which as I said doesn't carry much weight here, I don't think this is something where GLASS should have to innovate. We're in the cosmological simulations business, not the array business 🙂

betatim: There isn't a lot of context in the PR description, what I figure from a quick browse: this PR adds a decorator that converts all "arrays" that are passed to a function to numpy arrays, then calls the wrapped function and converts any "arrays" in the output back to the original namespace.

I'm not sure how much the work in https://github.com/scikit-learn/scikit-learn/pull/34324 is useful for others. It feels somewhat scikit-learn specific. The tricky (imho) part is deciding when to convert to numpy, which I guess is a library specific thing? At least in scikit-learn the decision to convert to numpy is not a simple "this function is hard to support". It depends on the global config, the specific constructor arguments of the class and what the type of the input array is. Once you know you want to convert to numpy it feels like you are "home free".

For this library it looks like most of the code of the decorator deals with finding arrays in the input arguments and such infrastructure. One thing I'm wondering is what to do if the inputs use different array namespaces or if one of the inputs is a torch array and another a `list`. But maybe these are things aren't an issue here? In scikit-learn we allow a mix of things :-/

Conclusion, I wonder if we can make a generic fallback decorator for `xpx` that doesn't get mega complex and at the same time supports most of the different behaviours users would want. If the answer is "yes we can" then I think it would be a useful tool (even if I don't think we'd use it in scikit-learn).

glass/grf/_solver.py
```diff
@@ -80,11 +82,8 @@ def solve(  # noqa: PLR0912, PLR0913
     :func:`glass.grf.compute`: Direct computation for band-limited spectra.
 
     """
-    xp = cl.__array_namespace__()
-
     # This function is difficult to port to the Array API so for now we work
     # in NumPy and ultimately convert back at the end of it.
```
paddyroddy: We would need to remove this comment too

glass/_array_api_utils.py
paddyroddy: Some typing and a docstring would be good here

paddyroddy (commented):
This seems like a really nice approach and is the kind of thing I was thinking of. Long-term I definitely wouldn't want the code in this repo as it is complex, and our current approach would be easier for contributors to understand. Do we think we've captured all edge cases here? I assume the decorator is smart enough to detect which variables need to be converted and which don't?

glass/_array_api_utils.py
prady0t: Thanks for the review. I've added some in the latest commit.

prady0t (commented):
Currently, the state of the decorator is:

- Goes through all the function parameters (args and kwargs) and collects those which have `__array_namespace__` attribute. 
- If `xp` is in kwargs, it uses its value. If value is none, then use `default_xp()`. `xp` value overwrites parameter's `__array_namespace__` attribute value.
- If no `xp` parameter and no parameter has `__array_namespace__ ` attribute, returns the function as is.
- Parameters are converted to numpy
- Depending on what the original function returns, the wrapper recursively converts (so that it can support tuple, list and dict) back to the `xp`. 

There are two places where we cannot use this decorator: `pixwin` and the `query_strip` function, as we are specifying `dtype` during conversion. Something that this decorator cannot do yet.

glass/healpix.py
paddyroddy, line 39: This is missing `@numpy_fallback`

glass/_array_api_utils.py
```diff
@@ -559,3 +562,88 @@ def tril_indices(
...
+
+        def collect_arrays(obj):
+            if hasattr(obj, "__array_namespace__"):
+                list_of_xp.append(obj)
+            elif isinstance(obj, dict):
+                for v in obj.values():
+                    collect_arrays(v)
+            elif isinstance(obj, Sequence) and not isinstance(obj, (str, bytes)):
+                for item in obj:
+                    collect_arrays(item)
```
prady0t: I've made the decorator even smarter! Now, it's able to extract xp if the given arguments are a collection (or Sequence) of xp. Similarly, NumPy conversions are also being done recursively. This is helpful in the places commented below.

glass/healpix.py
```diff
@@ -61,29 +61,17 @@ def alm2map(  # noqa: PLR0913
...
-            pixwin=pixwin,
-            pol=pol,
-        ),
+    return healpy.alm2map(
+        alms,
+        nside,
+        inplace=inplace,
+        lmax=lmax,
+        pixwin=pixwin,
+        pol=pol,
```
prady0t, line 77: Instance 1

glass/healpix.py
```diff
@@ -282,26 +262,13 @@ def map2alm(
...
-            lmax=lmax,
-            pol=pol,
-            use_pixel_weights=use_pixel_weights,
-        ),
+    return healpy.map2alm(
+        maps,
+        datapath=_get_healpy_datapath(),
+        lmax=lmax,
+        pol=pol,
+        use_pixel_weights=use_pixel_weights,
```
prady0t, line 275: 2

glass/healpix.py
```diff
@@ -451,15 +419,13 @@ def randang(
         A tuple ``theta, phi`` of mathematical coordinates.
 
     """
-    xp = ipix.__array_namespace__()
-
     theta, phi = healpix.randang(
         nside,
-        np.asarray(ipix),
+        ipix,
```
prady0t, line 428: 3

glass/healpix.py
```diff
@@ -497,8 +464,4 @@ def rotate_map_pixel(self, m: FloatArray) -> FloatArray:
             Map in the new reference frame
 
         """
-        xp = m.__array_namespace__()
-
-        return xp.asarray(
-            healpy.Rotator(coord=self.coord).rotate_map_pixel(np.asarray(m)),
-        )
+        return healpy.Rotator(coord=self.coord).rotate_map_pixel(m)
```
prady0t, line 471: 4

glass/_array_api_utils.py
prady0t: @paddyroddy gentle ping for a review.

paddyroddy: Sorry I've been on leave

array-api-tests
paddyroddy, line 1: Did you mean to add this @prady0t? Tests are failing. Should this be a submodule?

glass/healpix.py
```diff
@@ -147,17 +134,14 @@ def almxfl(
...
-
-    return xp.asarray(
-        healpy.almxfl(
-            np.asarray(alm),
-            np.asarray(fl),
-            inplace=inplace,
-        ),
+    return healpy.almxfl(
+        np.asarray(alm),
+        np.asarray(fl),
```
paddyroddy: Given `alm` and `fl` here are arrays, do they need the `np.asarray`?

glass/healpix.py
```diff
@@ -187,18 +171,15 @@ def ang2pix(
...
-        healpix.ang2pix(
-            nside,
-            np.asarray(theta),
-            np.asarray(phi),
-            lonlat=lonlat,
-        ),
+    return healpix.ang2pix(
+        nside,
+        np.asarray(theta),
+        np.asarray(phi),
```
paddyroddy: Is the `np.asarray` required for `theta`/`phi` here because they could be floats?

glass/healpix.py
```diff
@@ -225,14 +206,12 @@ def ang2vec(
         A normalised 3-vector pointing in the same direction as ``ang``.
 
     """
-    xp = _utils.default_xp() if xp is None else xp
-
     x, y, z = healpix.ang2vec(
         np.asarray(theta),
         np.asarray(phi),
```
paddyroddy: Same here https://github.com/glass-dev/glass/pull/1137/changes#r3674365356

glass/healpix.py
```diff
@@ -252,6 +231,7 @@ def get_nside(m: FloatArray) -> int:
     return int(healpy.get_nside(np.asarray(m)))
```
paddyroddy: Is the `np.asarray` required here? `m` is an array

paddyroddy: I guess the decorator isn't bening used here currently

glass/_array_api_utils.py
```diff
@@ -21,8 +21,11 @@
 ]
 
 import functools
+from functools import wraps
```
paddyroddy: This pedantic, but I don't really like `from` imports and try to use them sparingly. Helps to know where something comes from at a glance. So could we change it to `import functools; functools.wraps(func)`?

glass/_array_api_utils.py
```diff
@@ -563,3 +566,88 @@ def tril_indices(
...
+    Returns
+    -------
+        Wrapped function that accepts Array API arrays and returns results in
+        the corresponding array namespace.
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        list_of_xp = []
+
+        def collect_arrays(obj):
```
paddyroddy: A (short) docstring for each of the internal methods would be good here.

glass/_array_api_utils.py
```diff
@@ -563,3 +566,88 @@ def tril_indices(
...
+            if xp is None:
+                xp = array_api_compat.array_namespace(*list_of_xp, use_compat=False)
+        else:
+            xp = array_api_compat.array_namespace(*list_of_xp, use_compat=False)
+
+        def to_numpy(obj):
+            if hasattr(obj, "__array_namespace__"):
+                return np.asarray(obj)
+            if isinstance(obj, tuple):
+                return tuple(to_numpy(x) for x in obj)
```
paddyroddy, line 626: For example this function is called `to_numpy` but is outputting tuples etc. so not obvious what it is doing

paddyroddy:
Have fixed the linting in b9aacdb, hope you don't mind.

paddyroddy:
~@connoraird looks like the regression tests don't work on PRs from forks.~

Fixed in #1176.

prady0t:
> Have fixed the linting in https://github.com/glass-dev/glass/commit/b9aacdb6abb9cff879dfc49f22dc390d99a9b5f2, hope you don't mind.

Ofc not. I've applied the changes; thanks for all the suggestions. I must have forgotten to remove some of `np.asarray` which now I have removed.

glass/healpix.py
```diff
@@ -116,13 +104,12 @@ def alm2map_spin(
         List of 2 out maps in RING scheme as arrays.
 
     """
-    xp = array_api_compat.get_namespace(*alms, use_compat=False)
-
     inputs = [np.asarray(alm) for alm in alms]
```
paddyroddy: Just realised that `alms` is a `Sequence[FloatArray]` so surely `alm` would be a `FloatArray`. So we should be safe to remove the manual `np.asarray` here too?

paddyroddy (approved):
This looks great! Sorry it has taken so long. We've discussed this offline, and we are happy to merge this. Long-term we're wondering if it would make sense in something like https://github.com/data-apis/array-api-extra, perhaps as the `array_api_extra.numpy` module suggested above.

paddyroddy:
Having some regression test failures, but I'm 99% sure its noise. If I keep running them then hopefully it'll fix itself.

paddyroddy:
I'm merging this now (@connoraird), we need to get the regression tests stable again. Thanks @prady0t!

PR #1143, opened 2026-06-23
gh-1116: add resample_shapes()

### LINKED ISSUE #1116: Resampling galaxy shapes
Add functionality to sample galaxy shapes by randomising a catalogue of existing shapes.

This should optionally correct for the variance coming from cosmic shear (i.e. the catalogue with shear has the same variance before and after resampling).

ntessore:
# Description

Adds a new function `glass.resample_shapes()` that takes an array of ellipticity and randomises them. The function can optionally compensate for the intrinsic variance `varg` or `vargamma` of the (reduced) shear field (i.e., the variance of the ellipticities will be equal to the input array after applying a (reduced) shear).

Closes: #1116

## Changelog entry

Added: A new function `glass.resample_shapes()` for resampling an existing catalogue of ellipticities.

## Checks

### DIFF
--- a/tests/core/test_shapes.py
+++ b/tests/core/test_shapes.py
@@ -11,7 +11,7 @@ import glass
 if TYPE_CHECKING:
     from types import ModuleType
 
-    from glass._types import UnifiedGenerator
+    from glass._types import ComplexArray, UnifiedGenerator
 
 
 def test_triaxial_axis_ratio(
@@ -252,23 +252,30 @@ def test_ellipticity_intnorm(
         glass.ellipticity_intnorm(1, 0.71, xp=xp)
 
 
-def test_resample_shapes(
+@pytest.fixture
+def epsilon(
     urng: UnifiedGenerator,
     xp: ModuleType,
-) -> None:
-    """Test for resample_shapes() without compensation."""
-    # generate an array of random ellipticities
-    n = 1000
+) -> ComplexArray:
+    """Generate uniform random epsilon ellipticities."""
+    n = 1_000_000
     r = xp.sqrt(urng.uniform(0.0, 1.0, n))
     phi = urng.uniform(0.0, 2 * xp.pi, n)
-    epsilon = r * xp.exp(1j * phi)
+    return r * xp.exp(1j * phi)
 
+
+def test_resample_shapes(
+    epsilon: ComplexArray,
+    urng: UnifiedGenerator,
+) -> None:
+    """Test for resample_shapes() without compensation."""
     # resulting absolute values should match identically
     result = glass.resample_shapes(epsilon, rng=urng)
     xpx.testing.assert_close(abs(result), abs(epsilon), rtol=1e-10)
 
 
 def test_resample_shapes_varg(
+    epsilon: ComplexArray,
     urng: UnifiedGenerator,
     xp: ModuleType,
 ) -> None:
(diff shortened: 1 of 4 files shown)


tests/core/test_shapes.py
```diff
@@ -250,3 +250,87 @@ def test_ellipticity_intnorm(
...
+    result = glass.resample_shapes(epsilon, varg=varg, rng=urng)
+
+    # sample g with given variance
+    g = urng.normal(
+        scale=xp.sqrt(xp.asarray(varg / 2)),
+        size=(n, 2),
+    ) @ xp.asarray([1, 1j])
+
+    # apply g to result
+    result = (result + g) / (1 + result * xp.conj(g))
```
connoraird, line 296: Why are we editing the result in the test? Is there a way to check the result returned from `glass.resample_shapes` as is?

tests/core/test_shapes.py
```diff
@@ -250,3 +250,87 @@ def test_ellipticity_intnorm(
...
+def test_resample_shapes(
+    urng: UnifiedGenerator,
+    xp: ModuleType,
+) -> None:
+    """Test for resample_shapes() without compensation."""
+    # generate an array of random ellipticities
+    n = 1000
+    r = xp.sqrt(urng.uniform(0.0, 1.0, n))
+    phi = urng.uniform(0.0, 2 * xp.pi, n)
+    epsilon = r * xp.exp(1j * phi)
```
connoraird: Could we wrap this into a fixture to deduplicate? Could just be in this file if that is the only place it's used

tests/core/test_shapes.py
```diff
@@ -250,3 +250,87 @@ def test_ellipticity_intnorm(
...
+    result = glass.resample_shapes(epsilon, vargamma=vargamma, rng=urng)
+
+    # sample gamma with given variance
+    gamma = urng.normal(
+        scale=xp.sqrt(xp.asarray(vargamma / 2)),
+        size=(n, 2),
+    ) @ xp.asarray([1, 1j])
+
+    # apply gamma to result
+    result = result + gamma
```
connoraird, line 325: I'm probably misunderstanding, but should this be applied to `epsilon` not `result`?

tests/core/test_shapes.py
ntessore: No, we cannot check without doing the operation on the results. The compensation ensures the results have the same variance as the input after these specific operations have been carried out, and we can only check this in the aggregate because it's a statistical effect.

tests/core/test_shapes.py
ntessore: I guess we could do that, but it seems a bit overkill to save less than 8 lines of code.

tests/core/test_shapes.py
ntessore: See above: in real life, we get a catalogue *epsilon* that contains this specific effect already. We then randomise this catalogue. If we now put the effect back on top, we'd get a larger variance than in the real catalogue. The compensation accounts for that, but to check, we need to apply this operation to the results again.

tests/core/test_shapes.py
connoraird: I think there's value in it if we care about the problem size being the same for all of these tests. Otherwise, I take your point and we can leave as is.

tests/core/test_shapes.py
connoraird: I understand

connoraird (approved):
Looks good. I was also experiencing the failing examples, I'm not sure what's causing it.

PR #1146, opened 2026-06-23
gh-1144: Alias xbp and urngb to xp and urng in regression tests

### LINKED ISSUE #1144: Alias xpb in regression tests to xp

It is confusing to switch between `xp` in core tests and `xpb` in regression tests. Therefore, we could alias `xp` to the trimmed down versions using the [suggested approach from pytest](https://docs.pytest.org/en/stable/how-to/fixtures.html#override-a-fixture-on-a-directory-conftest-level).

We add a fixture to `tests/regression/conftest.py` which aliases xp to the filtered backends:

```py
@pytest.fixture
def xp(xpb: ModuleType) -> ModuleType:
    """Alias of Fixture for array backend to be used in regression tests."""
    return xpb
```

_No response_

### Additional Context

_No response_

connoraird:
# Description

Changes `xpb` and `urngb` to `xp` and `urng` respectively in regression tests to reduce confusion when swapping between core and regression tests.

Closes: #1144 

## Checks

### DIFF
--- a/tests/regression/conftest.py
+++ b/tests/regression/conftest.py
@@ -19,7 +19,7 @@ def xp(xpb: ModuleType) -> ModuleType:
 
 
 @pytest.fixture
-def urngb(xpb: ModuleType) -> UnifiedGenerator:
+def urng(xpb: ModuleType) -> UnifiedGenerator:
     """
     Fixture for a unified RNG interface to be used in regression tests.
 
--- a/tests/regression/conftest.py
+++ b/tests/regression/conftest.py
@@ -4,11 +4,28 @@ from typing import TYPE_CHECKING
 
 import pytest
 
+from glass import _rng
+
 if TYPE_CHECKING:
     from types import ModuleType
+
+    from glass._types import UnifiedGenerator
 
 
 @pytest.fixture
 def xp(xpb: ModuleType) -> ModuleType:
     """Alias of Fixture for array backend to be used in regression tests."""
     return xpb
+
+
+@pytest.fixture
+def urngb(xpb: ModuleType) -> UnifiedGenerator:
+    """
+    Fixture for a unified RNG interface to be used in regression tests.
+
+    Access the relevant RNG using `urngb.` in tests.
+
+    Must be used with the `xpb` fixture.
+
+    """
+    return _rng.rng_dispatcher(xp=xpb)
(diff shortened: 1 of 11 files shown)


tests/regression/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+if TYPE_CHECKING:
+    from types import ModuleType
+
+    from glass._types import UnifiedGenerator
+
+
+@pytest.fixture
+def xp(xpb: ModuleType) -> ModuleType:
+    """Alias of Fixture for array backend to be used in regression tests."""
+    return xpb
```
paddyroddy, line 18: Why not just move `xpb` here and rename it to `xp`?

paddyroddy (changes requested):

tests/regression/conftest.py
connoraird: Sure, can do. I think I thought about this being something we essentially never changed and was just a bit of boilerplate, but what you've suggested probably makes more sense.

connoraird: I remembered the reason we can't do this. To move `xpb` here we would need to know `xp_available_backends` outside of a fixture which we don't . Unless you're happy with importing from a fixtures module `from tests.fixtures.array_backends import xp_available_backends`?

paddyroddy: Ah, I see the issue. We are doing `from tests.fixtures.` in various test files, but all in `TYPE_CHECKING` blocks. Tbh I think it's fine to do the import, and probably clearer than what we have here.

connoraird: I believe this [CI run](https://github.com/glass-dev/glass/actions/runs/28869978760/job/85629856545?pr=1146) shows that we need to do it the old way where we alias xbp as xp.

paddyroddy: Ah, yeah this is because importing relies on `__init__.py` which doesn't exist, and shouldn't according to `pytest`

tests/regression/conftest.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from glass import _rng
```
paddyroddy, line 7: Reminder to myself to change this

paddyroddy (approved):
Looks good

PR #1148, opened 2026-06-23
gh-1147: Create xp_no_jax

### LINKED ISSUE #1147: xpb used to mean everything but jax
### Add your issue here

# Description

In `tests/core/test_fields.py` we are using `xpb` to refer to all backends excluding JAX. However, this is not what `xpb` is. `xpb` is instead just numpy at the moment and will expand to include JAX in the future.

We should therefore add  `xp_no_jax` which is `xp` with JAX removed.

connoraird:
# Description

Add `xp_no_jax` to replace `xpb` when used as `xp` excluding JAX

Closes: #1147

## Checks

### DIFF
--- a/tests/core/test_fields.py
+++ b/tests/core/test_fields.py
@@ -163,13 +163,13 @@ def test_cls2cov_jax(jnp: ModuleType) -> None:
         xpx.testing.assert_close(cov2, cov3)
 
 
-def test_cls2cov_no_jax(xpb: ModuleType) -> None:
+def test_cls2cov_no_jax(xp_no_jax: ModuleType) -> None:
     # check output values and shape
 
     nl, nf, nc = 3, 2, 2
 
     generator = glass.cls2cov(
-        [xpb.asarray([1.0, 0.5, 0.3]), None, xpb.asarray([0.7, 0.6, 0.1])],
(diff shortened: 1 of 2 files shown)


ntessore:
Isn't this much cleaner handled by a `skip` at test level instead of changing the fixture?

paddyroddy (approved):
Looks good to me, not sure about the name though

paddyroddy:
> Isn't this much cleaner handled by a `skip` at test level instead of changing the fixture?

I think if we're not careful then that would skip the whole test. We're interested in skipping for certain backends, but still running the test. Could be done in another way potentially though.

ntessore:
> > Isn't this much cleaner handled by a `skip` at test level instead of changing the fixture?
> 
> I think if we're not careful then that would skip the whole test. We're interested in skipping for certain backends, but still running the test. Could be done in another way potentially though.

We use this fixture in exactly one test. So we could just do:

```py
def test_cls2cov_no_jax(xpb: ModuleType) -> None:
    if xpb.__name__ == "jax.numpy":
        pytest.skip("test does not work with jax")

    ...
```

paddyroddy:
Ah, yes that makes sense. I was thinking of a more complex example modifying `xp` fixture itself.

PR #1154, opened 2026-07-01
gh-1153: Add temporary fix for NumPy in examples

### LINKED ISSUE #1153: Pin NumPy version for examples due to CAMB bug
### Describe the Bug

We are experiencing this bug https://github.com/cmbant/CAMB/issues/203 which we can avoid locally by pinning NumPy to `<2.4.0`.

### To Reproduce

```python
import glass
...
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Version In Use

2026.3.dev12+g3b1b967f5

### Additional Context

```markdown
- Python version:
- Operating system:
```

paddyroddy:
# Description

The examples are currently failing as they rely on CAMB which experiencing this bug cmbant/CAMB#203. By putting an upper bound on NumPy for the examples we should be able to minimise disruption until this is fixed.

Fixes: #1153

## Changelog entry

Fixed: The examples that run with CAMB no longer have a NumPy version bug

## Checks

### DIFF
--- a/.github/workflows/test-examples-on-push.yaml
+++ b/.github/workflows/test-examples-on-push.yaml
@@ -13,6 +13,7 @@ on:
       - .github/workflows/run-examples.yaml
       - examples/**
       - glass/**
+      - pyproject.toml
     types:
       - opened
       - ready_for_review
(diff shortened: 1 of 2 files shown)


connoraird (approved):
Looks good

PR #1162, opened 2026-07-07
gh-1159: Add `CosmologyWithOmegaM` protocol

### LINKED ISSUE #1159: Remove HasOmegaM from Cosmology protocol
Remove `HasOmegaM` from the `Cosmology` protocol and only add it as a requirement to `DensityWeight`.

The time-dependent matter density is not very useful in general and only used in `DensityWeight`, which isn't used very often.

We can create a `CosmologyWithOmegaM` protocol intersection for this.

paddyroddy:
# Description

Adds a `CosmologyWithOmegaM` protocol as `OmegaM` isn't used that often.

Closes: #1159

## Changelog entry

Added: A `CosmologyWithOmegaM` protocol

## Checks

### DIFF
--- a/glass/cosmology.py
+++ b/glass/cosmology.py
@@ -16,8 +16,15 @@ class Cosmology(
     cosmology.api.HasInverseComovingDistance[AnyArray, AnyArray],  # ty: ignore[invalid-type-arguments]
     cosmology.api.HasLittleH[AnyArray],  # ty: ignore[invalid-type-arguments]
     cosmology.api.HasOmegaM0[AnyArray],  # ty: ignore[invalid-type-arguments]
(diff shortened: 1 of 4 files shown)


connoraird (commented):
Similar to my comments on other PRs, should we be updating the fixture type for `test_density_weight` to be `CosmologyWithOmegaM`?

paddyroddy:
> Similar to my comments on other PRs, should we be updating the fixture type for `test_density_weight` to be `CosmologyWithOmegaM`?

I'm not 100% sure, this is what I originally had, but Copilot said to change it.

paddyroddy:
@connoraird realise now I hadn't made a new fixture, the latest change does that now so hopefully passes.

connoraird (approved):
Looks good

paddyroddy:
Have re-added the `typing.Protocol` and will merge, so this doesn't get very stale.

PR #1163, opened 2026-07-08
gh-1158: Add `CosmologyWithGrowthFactor` protocol

### LINKED ISSUE #1158: Remove HasGrowthFactor from Cosmology protocol
Remove `HasGrowthFactor ` from the `Cosmology` protocol and only add it as a requirement to `_kappa_ia_nla()`.

The growth factor is difficult to provide in general and only required for the `_kappa_ia_nla()` function, which isn't even publicly callable.

We can create a `CosmologyWithGrowthFactor` protocol intersection for this.

paddyroddy:
# Description

Adds a `CosmologyWithGrowthFactor` protocol as `growth_factor` isn't used that often.

Closes: #1158

## Changelog entry

Added: A `CosmologyWithGrowthFactor` protocol

## Checks

### DIFF
--- a/glass/cosmology.py
+++ b/glass/cosmology.py
@@ -10,7 +10,6 @@ from glass._types import AnyArray
 class Cosmology(
     cosmology.api.HasComovingDistance[AnyArray, AnyArray],  # ty: ignore[invalid-type-arguments]
     cosmology.api.HasCriticalDensity0[AnyArray],  # ty: ignore[invalid-type-arguments]
-    cosmology.api.HasGrowthFactor[AnyArray, AnyArray],  # ty: ignore[invalid-type-arguments]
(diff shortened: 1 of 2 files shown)


connoraird (commented):
`_kappa_ia_nla` isn't used in the repo and has been hidden. Can we instead just delete it and remove the growth factor from the Cosmology interface?

paddyroddy:
> `_kappa_ia_nla` isn't used in the repo and has been hidden. Can we instead just delete it and remove the growth factor from the Cosmology interface?

Yes I know (#411)... @ntessore want it to stay. What we really need is some tests and for it be exposed.

connoraird (approved):
Looks good

paddyroddy:
Have re-added the `typing.Protocol` and will merge, so this doesn't get very stale.

PR #1168, opened 2026-07-16
gh-1167: Prepare for lazy imports in `3.15`

### LINKED ISSUE #1167: Prepare for lazy imports
### Add your issue here

[PEP 810](https://peps.python.org/pep-0810/) introduces the concept of `lazy` imports, this feature comes in Python `3.15`. I had thought we would have to wait until `3.15` is the lowest version, but turns out `__lazy_modules__` exists now, and enables backwards compatibility. So it doesn't do anything for now, but once we enable `3.15` it will work automatically.

paddyroddy:
# Description

Uses the `__lazy_modules__` in the `glass/glass` module to speed up use elsewhere once Python `3.15` is released.

Closes: #1167

## Checks

### DIFF
--- a/glass/fields.py
+++ b/glass/fields.py
@@ -2,6 +2,12 @@
 
 from __future__ import annotations
 
+__lazy_modules__ = [
+    "array_api_compat",
+    "array_api_extra",
+    "transformcl",
+]
+
 import collections
 import itertools
 import math
--- a/glass/_array_api_utils.py
+++ b/glass/_array_api_utils.py
@@ -16,6 +16,10 @@ integration, interpolation, and linear algebra.
 
 from __future__ import annotations
 
+__lazy_modules__ = [
+    "array_api_compat",
+]
+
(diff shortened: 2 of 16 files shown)


glass/fields.py
```diff
@@ -2,6 +2,12 @@
 
 from __future__ import annotations
 
+__lazy_modules__ = [
+    "array_api_compat",
+    "array_api_extra",
+    "transformcl",
```
connoraird, line 8: Why do we lazy import some modules but not others such as math or sys?

paddyroddy: That was me. I used https://github.com/henryiii/flake8-lazy to do it automatically. But I figured:
- I don't think we should be doing it in `tests` or `examples`
- I figured system libraries like `math` would be quick to import so little benefit
- Similarly I didn't include `glass`

connoraird (approved):

PR #1169, opened 2026-07-21
gh-1145: Default to wrapping rng if backend is not known

### LINKED ISSUE #1145: rng backend isn't validated
### Add your issue here

# Description

When we have an optional `rng` parameter in a function, we do not currently validate its array backend. This could mean we have mixed array backends within a single function.

connoraird:
# Description

- Alters the existing `glass.rng.Generator` to allow it to wrap more than just NumPy RNGs.
- Adds a new RNG protocol to ensure the wrapped rng in `glass.rng.Generator` defines the required methods.
- Updates how rngs are handles within functions to ensure they always return the required array backend.

Closes: #1145 

## Changelog entry

Changed: `glass.rng.Generator` class now wraps returned values ensuring all methods return the correct array backend.

## Checks

### DIFF
--- a/glass/galaxies.py
+++ b/glass/galaxies.py
@@ -74,6 +74,9 @@ def _draw_nz(
         Redshifts sampled from the given source distribution.
 
     """
+    xp = array_api_compat.array_namespace(count, z, nz, use_compat=False)
+    rng = glass.rng.Generator(rng=rng, xp=xp)
+
     # compute the CDF
     cdf = glass.arraytools.cumulative_trapezoid(nz, z)
     cdf /= cdf[-1]
@@ -157,8 +160,7 @@ def redshifts_from_bins(
     xp = array_api_compat.array_namespace(bins, z, *nz_values, use_compat=False)
 
     # get default RNG if not given
-    if rng is None:
-        rng = glass.rng.rng_dispatcher(xp=xp)
+    rng = glass.rng.Generator(rng=rng, xp=xp)
 
     # tally the bins
     bin_label, _, bin_index, bin_count = xp.unique_all(bins)
@@ -234,9 +236,7 @@ def redshifts_from_nz(
             stacklevel=2,
         )
 
-    # get default RNG if not given
-    if rng is None:
-        rng = glass.rng.rng_dispatcher(xp=xp)
+    rng = glass.rng.Generator(rng=rng, xp=xp)
 
     # bring inputs' leading axes into common shape
     dims, *rest = glass.arraytools.broadcast_leading_axes((count, 0), (z, 1), (nz, 1))
@@ -420,8 +420,7 @@ def gaussian_phz(  # noqa: PLR0913
     sigma_0_arr = xp.asarray(sigma_0)
 
     # get default RNG if not given
-    if rng is None:
-        rng = glass.rng.rng_dispatcher(xp=xp)
+    rng = glass.rng.Generator(rng=rng, xp=xp)
 
     # Ensure lower and upper are arrays that have the same shape and type
     lower_arr = xp.asarray(0.0 if lower is None else lower, dtype=xp.float64)
--- a/glass/galaxies.py
+++ b/glass/galaxies.py
@@ -159,7 +159,6 @@ def redshifts_from_bins(
 
     xp = array_api_compat.array_namespace(bins, z, *nz_values, use_compat=False)
 
-    # get default RNG if not given
     xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # tally the bins
@@ -419,7 +418,6 @@ def gaussian_phz(  # noqa: PLR0913
     z_arr = xp.asarray(z)
     sigma_0_arr = xp.asarray(sigma_0)
 
-    # get default RNG if not given
     xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # Ensure lower and upper are arrays that have the same shape and type
--- a/glass/galaxies.py
+++ b/glass/galaxies.py
@@ -75,7 +75,7 @@ def _draw_nz(
 
     """
     xp = array_api_compat.array_namespace(count, z, nz, use_compat=False)
-    rng = glass.rng.Generator(rng=rng, xp=xp)
+    xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # compute the CDF
     cdf = glass.arraytools.cumulative_trapezoid(nz, z)
@@ -83,7 +83,7 @@ def _draw_nz(
 
     # sample redshifts and return result
     return uxpx.interp(
-        rng.uniform(0.0, 1.0, size=count),
+        xrng.uniform(0.0, 1.0, size=count),
         cdf,
         z,
     )
@@ -160,7 +160,7 @@ def redshifts_from_bins(
     xp = array_api_compat.array_namespace(bins, z, *nz_values, use_compat=False)
 
     # get default RNG if not given
-    rng = glass.rng.Generator(rng=rng, xp=xp)
+    xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # tally the bins
     bin_label, _, bin_index, bin_count = xp.unique_all(bins)
@@ -168,7 +168,7 @@ def redshifts_from_bins(
     # sample the redshifts from each nz
     # concatenate into runs of redshifts for each bin
     redshifts = xp.concat([
-        _draw_nz(int(k), z, nz_values[nz_keys.index(x)], rng=rng)
+        _draw_nz(int(k), z, nz_values[nz_keys.index(x)], rng=xrng)
         for x, k in zip(bin_label, bin_count, strict=True)
     ])
 
@@ -236,7 +236,7 @@ def redshifts_from_nz(
             stacklevel=2,
         )
 
-    rng = glass.rng.Generator(rng=rng, xp=xp)
+    xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # bring inputs' leading axes into common shape
     dims, *rest = glass.arraytools.broadcast_leading_axes((count, 0), (z, 1), (nz, 1))
@@ -259,7 +259,7 @@ def redshifts_from_nz(
                 int(count_out[k]),
                 z_out_slice,
                 nz_out_slice,
-                rng=rng,
+                rng=xrng,
             ),
         )
         total += count_out[k]
@@ -420,7 +420,7 @@ def gaussian_phz(  # noqa: PLR0913
     sigma_0_arr = xp.asarray(sigma_0)
 
     # get default RNG if not given
-    rng = glass.rng.Generator(rng=rng, xp=xp)
+    xrng = glass.rng.Generator(rng=rng, xp=xp)
 
     # Ensure lower and upper are arrays that have the same shape and type
     lower_arr = xp.asarray(0.0 if lower is None else lower, dtype=xp.float64)
@@ -432,7 +432,7 @@ def gaussian_phz(  # noqa: PLR0913
 
     sigma = xp.add(1, z_arr) * sigma_0_arr
     dims = sigma.shape
-    zphot = xp.asarray(rng.normal(z_arr, sigma))
+    zphot = xp.asarray(xrng.normal(z_arr, sigma))
 
     # Check for valid user input
     if (lower_arr.ndim == upper_arr.ndim != 0) and not (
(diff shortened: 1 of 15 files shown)


tests/core/test_array_api_utils.py
```diff
@@ -44,6 +44,14 @@ def test_init() -> None:
     assert isinstance(rng, glass.rng.Generator)
 
 
+@pytest.mark.skipif(not HAVE_ARRAY_API_STRICT, reason="test requires array_api_strict")
+def test_init_mix_of_backends() -> None:
+    import array_api_strict
+
+    rng = glass.rng.Generator(rng=np.random.default_rng(), xp=array_api_strict)
+    assert isinstance(rng, glass.rng.Generator)
```
paddyroddy: So to check the purpose of this PR. We are allowing a mix of RNG and `xp`? Should we be testing that more explicitly? Like here, is `rng` NumPy or `array-api-strict`? Does it matter?

glass/galaxies.py
```diff
@@ -74,13 +74,16 @@ def _draw_nz(
         Redshifts sampled from the given source distribution.
 
     """
+    xp = array_api_compat.array_namespace(count, z, nz, use_compat=False)
```
paddyroddy, line 77: This line is just because we need an `xp` to be used within the new `Generator` class?

glass/galaxies.py
```diff
@@ -420,8 +420,7 @@ def gaussian_phz(  # noqa: PLR0913
     sigma_0_arr = xp.asarray(sigma_0)
 
     # get default RNG if not given
```
paddyroddy: This comment has been left here. Can we remove it everywhere?

glass/rng.py
```diff
@@ -36,12 +36,13 @@ def default_rng(
     if xp.__name__ == "jax.numpy":
         import glass.jax  # noqa: PLC0415
 
-        return glass.jax.Generator(seed=seed)
+        return glass.jax.Generator(seed=seed)  # ty: ignore[invalid-return-type]
```
paddyroddy: Should we be changing the output type above from `UnifiedGenerator?`

glass/rng.py
```diff
@@ -36,12 +36,13 @@ def default_rng(
...
 
-    if xp.__name__ == "numpy":
-        return xp.random.default_rng(seed=seed)
+    import numpy as np  # noqa: PLC0415
 
-    return Generator(xp=xp, seed=seed)
+    rng = np.random.default_rng(seed=seed)
+
+    return rng if xp.__name__ == "numpy" else Generator(rng=rng, xp=xp)  # ty: ignore[invalid-return-type]
```
paddyroddy: I think what we had before is clearer, i.e.

```python
if xp.__name__ == "numpy":
    return rng

import numpy as np  # noqa: PLC0415

rng = np.random.default_rng(seed=seed)
return Generator(rng=rng, xp=xp)
```

glass/rng.py
```diff
@@ -63,18 +64,20 @@ def rng_dispatcher(*, xp: ModuleType) -> UnifiedGenerator:
...
-    __slots__ = ("np", "rng", "xp")
+    __slots__ = ("default_dtype", "rng", "xp")
 
     def __init__(
         self,
+        *,
         xp: ModuleType,
+        rng: UnifiedGenerator | None = None,
         seed: int | IntArray = SEED,
     ) -> None:
```
paddyroddy, line 79: Given that they're now all keyword, how about


```suggestion
        rng: UnifiedGenerator | None = None,
        seed: int | IntArray = SEED,
        xp: ModuleType,
```

glass/rng.py
```diff
@@ -88,11 +91,12 @@ def __init__(
...
-        import numpy  # noqa: ICN001, PLC0415
-
         self.xp = xp
-        self.np = numpy
-        self.rng = self.np.random.default_rng(seed=seed)
+        self.default_dtype = xp.float64
+        if rng is None:
+            self.rng = default_rng(seed=seed, xp=xp)
+        else:
+            self.rng = rng
```
paddyroddy, line 104: Whereas this, I think it's okay to do

```suggestion
        self.rng = default_rng(seed=seed, xp=xp) if rng is None else rng
```

glass/rng.py
```diff
@@ -117,8 +121,8 @@ def random(
             Array of random floats.
 
         """
-        dtype = dtype if dtype is not None else self.np.float64
-        return self.xp.asarray(self.rng.random(size, dtype, out))  # ty: ignore[no-matching-overload]
+        dtype = dtype if dtype is not None else self.default_dtype
+        return self.xp.asarray(self.rng.random(size, out), dtype=dtype)
```
paddyroddy: I like it, but why can we do this now and why couldn't we before?

glass/rng.py
```diff
@@ -190,8 +194,8 @@ def standard_normal(
             Array of samples from the standard normal distribution.
 
         """
-        dtype = dtype if dtype is not None else self.np.float64
-        return self.xp.asarray(self.rng.standard_normal(size, dtype, out))  # ty: ignore[no-matching-overload]
+        dtype = dtype if dtype is not None else self.default_dtype
+        return self.xp.asarray(self.rng.standard_normal(size, out), dtype=dtype)
```
paddyroddy: I like it, but why can we do this now and why couldn't we before?

glass/rng.py
paddyroddy, line 81: I'm still not overly comfortable with `Protocols`. We are duplicating the docstrings here. Is this expected behaviour? Can you find any examples which show duplication Vs not?

glass/points.py
```diff
@@ -522,8 +521,7 @@ def positions_from_delta(  # noqa: PLR0913
     xp = array_api_compat.array_namespace(ngal, delta, bias, vis, use_compat=False)
 
     # get default RNG if not given
-    if rng is None:
-        rng = glass.rng.rng_dispatcher(xp=xp)
```
paddyroddy, line 526: Are we still using `rng_dispatcher` at all?

paddyroddy (changes requested):

tests/core/test_array_api_utils.py
connoraird: Currently, we can have a mix of `rng` and `xp`  which will could in theory cause a failure because of mismatched array backends. This PR ensure that, if the `rng` provided does not match `xp`, we convert the returned `rng` arrays to the correct `xp` backend.

Here `rng` is the NumPy generator wrapped by `array-api-strict` but in theory we could have any combination of `rng` and `xp`

glass/galaxies.py
connoraird: Yes

glass/galaxies.py
connoraird: Can do, wasn't sure if it was still useful

glass/rng.py
connoraird: Change it to what? An alternative is to keep UnifiedGenerator as it was and give this new protocol class a different name?

glass/rng.py
connoraird: The example I used was from cosmology api. Happy to deduplicate but I feel like it's important to have the docstrings in the protocol as that's where we are defining the signature. Is there something similar to doxygen `copydoc`?

glass/rng.py
connoraird: I don't know why we couldn't before TBH.

glass/points.py
connoraird: We are using it when we know for a fact `rng` is `None` (i.e. a new instance) so we don't need to do the loop of `Generator` ->  `default_rng` -> `Generator`. Currently only in one place.

glass/rng.py
connoraird: I've updated the name of the protocol

tests/core/test_array_api_utils.py
paddyroddy: What is the advantage of asserting that `rng` is `glass.rng.Generator`? Is that not inevitable?

glass/galaxies.py
paddyroddy: Well I don't mind either way, but you had removed it some places and not others

glass/rng.py
paddyroddy: Will have a look, but yes the alternative was some long-winded `UnifiedGenerator | glass.jax.Generator | ...` etc.

tests/core/test_array_api_utils.py
connoraird: Good point. I've updated to check the returned arrays are as expected.

connoraird:
@paddyroddy I've made some significant changes after discovering a bug when adding the new tests with JAX.

glass/rng.py
paddyroddy: This is nicer, feel the old way wasn't super readable

glass/points.py
paddyroddy: Given that `rng_dispatcher` is only `default_rng(seed=SEED, xp=xp)`, is it still worth having?

paddyroddy:
> @paddyroddy I've made some significant changes after discovering a bug when adding the new tests with JAX.

Okay, will have a look

glass/rng.py
paddyroddy: So I did some reading and I think we should copy the one line top of the docstring in the `Protocol`, but not the rest. I've pushed this for simplicity.

glass/rng.py
```diff
@@ -63,19 +111,23 @@ def rng_dispatcher(*, xp: ModuleType) -> UnifiedGenerator:
...
 
-    This class wraps NumPy's random number generator and returns arrays compatible
-    with the provided backend.
+    This class wraps random number generators which match the glass UnifiedGenerator
+    type and returns arrays compatible with the provided backend.
 
     """
 
-    __slots__ = ("np", "rng", "xp")
+    default_dtype: DTypeLike
```
paddyroddy, line 67: We could do this?

```suggestion
    __slots__ = ("default_dtype", "rng", "xp")

    default_dtype: DTypeLike
```

glass/rng.py
```diff
@@ -88,17 +140,26 @@ def __init__(
...
-        self.np = numpy
-        self.rng = self.np.random.default_rng(seed=seed)
+        self.default_dtype = xp.float64
+
+        if rng is None:
+            if xp.__name__ == "jax.numpy":
+                import glass.jax  # noqa: PLC0415
+
+                self.rng = glass.jax.Generator(seed=seed)
+
```
paddyroddy:
```suggestion
```

tests/core/test_array_api_utils.py
```diff
@@ -44,6 +44,31 @@ def test_init() -> None:
     assert isinstance(rng, glass.rng.Generator)
 
 
+@pytest.mark.skipif(not HAVE_ARRAY_API_STRICT, reason="test requires array_api_strict")
```
paddyroddy, line 47: These tests are much better

glass/jax.py
```diff
@@ -201,6 +205,10 @@ def multinomial(
         dtype: DTypeLike = int,
     ) -> IntArray:
         """Draw samples from a multinomial distribution."""
+        # Ensure arrays are jax arrays
+        n = jnp.asarray(n)
+        pvals = jnp.asarray(pvals)
```
paddyroddy, line 210: Interesting that it's only these two methods that needed it

connoraird: In this case, it's needed because of the way we index `pvals` below `pvals[..., 0]`. I suppose `n` may not be necessary but it's a bit more complete.

glass/rng.py
connoraird: Looks good. What's the benefit of `__slots__`?

glass/jax.py
connoraird: Although, maybe I'm misremembering  as both instances are passing into `_shape` so that might be it.

glass/points.py
connoraird: Probably not

glass/rng.py
connoraird: I've had a search and seems `__slots__` making it faster

connoraird:
I've removed the Protocol class as I don't think it was actually doing anything. It definitely wasn't doing what I wanted which was restricting what RNGs could be passed to `Generator`

glass/jax.py
paddyroddy: Remind me why this change is needed now? Is it because we are now actually testing with JAX?

glass/rng.py
paddyroddy: Yeah, I'm not super familiar with it as a concept, but I've used it before, and it is what we had before so seems like a suitable change.

paddyroddy:
> I've removed the Protocol class as I don't think it was actually doing anything. It definitely wasn't doing what I wanted which was restricting what RNGs could be passed to `Generator`

Okay makes sense. Would be good if @ntessore could explain them to us a bit more 😅

paddyroddy:
Could have been a fluke, but the examples took more than 20 mins and failed. Looks like they usually take about 15 mins. Have re-run, otherwise we might have had a regression.

glass/rng.py
```diff
@@ -88,17 +88,25 @@ def __init__(
...
+            if xp.__name__ == "jax.numpy":
+                import glass.jax  # noqa: PLC0415
+
+                self.rng = glass.jax.Generator(seed=seed)
+            else:
+                import numpy as np  # noqa: PLC0415
+
+                self.rng = np.random.default_rng(seed=seed)
+        else:
+            self.rng = rng
```
paddyroddy, line 104: I think this logic is nice and clear now

paddyroddy (approved):
Looks good, see one comment on examples timing

paddyroddy:
Examples took 21:00

PR #1176, opened 2026-07-30
gh-1172: Allow regression tests to be run from forks

### LINKED ISSUE #1172: The regression tests don't work on PRs from forks
@connoraird looks like the regression tests don't work on PRs from forks.

_Originally posted by @paddyroddy in https://github.com/glass-dev/glass/issues/1137#issuecomment-5119644358_

paddyroddy:
# Description

I noticed in #1021 that the regression tests couldn't be run from forks. This PR attempts to address that by using the `refs` syntax that the GitHub CLI uses, e.g. `gh pr checkout`.

Tested from a fork https://github.com/glass-dev/glass/pull/1179.

Fixes: #1172

## Changelog entry

Fixed: The regression tests are now run from forks

## Checks

### DIFF
--- a/.github/workflows/regression-test.yaml
+++ b/.github/workflows/regression-test.yaml
@@ -26,6 +26,7 @@ on:
       - pyproject.toml
     branches:
       - main
+      - paddy/issue-1172
     types:
       - opened
       - ready_for_review
--- a/.github/workflows/regression-test.yaml
+++ b/.github/workflows/regression-test.yaml
@@ -47,8 +47,8 @@ jobs:
           persist-credentials: false
           ref: >-
             ${{ github.event_name == 'pull_request' &&
-            format('refs/pull/{0}/head', github.event.pull_request.number) ||
-            (inputs.head_ref || github.ref) }}
+            github.event.pull_request.head.sha || (inputs.head_ref ||
+            github.ref) }}
 
       - name: Set up uv
         # yamllint disable-line rule:line-length
--- a/.github/workflows/regression-test.yaml
+++ b/.github/workflows/regression-test.yaml
@@ -26,7 +26,6 @@ on:
       - pyproject.toml
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/regression-test.yaml
```diff
@@ -26,6 +26,7 @@ on:
       - pyproject.toml
     branches:
       - main
+      - paddy/issue-1172
```
connoraird: I guess, remove this before merging?

noxfile.py
```diff
@@ -273,7 +273,12 @@ def regression_tests(session: nox.Session) -> None:
     success_codes = [0, 5] if "-k" in session.posargs[2:] else [0]
 
     session.log(f"Comparing {before_revision} benchmark to revision {after_revision}")
-    session.install(f"git+{GLASS_REPO_URL}@{after_revision}")
+    if after_revision == "local":
+        session.log("Installing after-revision from local checkout")
+        session.install("--force-reinstall", ".")
```
connoraird, line 278: I this slightly unrelated? Does it allow running the regressions with locally checked out current branch? If it isn't essential to allow forks to run the reg tests, could you add a note about it in the PR description?

connoraird (approved):
Just a few minor comments

.github/workflows/regression-test.yaml
paddyroddy: Yes

noxfile.py
paddyroddy: So how I currently have it the `local` bit is essential. The `force-reinstall` is to avoid caching errors. I'm just looking into another approach.

paddyroddy: The checkout step gets the fork's code onto the runner via GitHub's `refs/pull/<n>/head` mirror; local tells `nox` to install from that local checkout instead of trying (and failing) to re-fetch the same commit from the wrong repo.

PR #1179, opened 2026-07-31
gh-1172: Test to see regression tests run from a fork

paddyroddy:
This is a temporary PR to check that the regression tests can now be run by a fork.


paddyroddy:
Okay, so it failed, but they did run.

PR #1138, opened 2026-06-17
gh-1111: add redshifts_from_bins()

ntessore:
# Description

Add a new function `glass.redshift_from_bins()` that allows sampling redshifts from existing tomographic bin labels and redshift distributions.

Closes: #1111

## Changelog entry

Added: A new function `glass.redshifts_from_bins()` to sample redshifts from existing tomographic bin labels and redshift distributions.

## Checks

- [x] Is your code passing linting?
- [x] Is your code passing tests?
- [x] Have you added additional tests (if required)?
- [x] Have you modified/extended the documentation (if required)?
- [x] Have you added a one-liner changelog entry above (if required)?

PR #1180, opened 2026-07-31
gh-930: Make the regressions pass if func is missing on `main`

### LINKED ISSUE #930: Can we separate out deviations from `main` from true regressions?
### What is the Topic of Your Question

Other

### Add Your Question Below

The regression tests sometimes fail because the code doesn't exist on `main`. So this is a "known"/"intended" failure. It would be good if we could separate this out, such that regression "failures" are easier to spot, and therefore not dismissed.

paddyroddy:
# Description

When a new function is added with a regression test, i.e. https://github.com/glass-dev/glass/pull/1138, the regression test will always fail in that PR. This isn't particularly helpful and could hide a real failure caused by the PR if people don't look. The idea behind this PR is to make sure that all regression tests are skipped if a given function/module doesn't exist. All future regression tests should follow the same format.

Fixes: #930

## Changelog entry

Fixed: Regression tests will no longer fail when a new function is added

## Checks

### DIFF
--- a/tests/regression/test_fields.py
+++ b/tests/regression/test_fields.py
@@ -26,6 +26,10 @@ glass_fields = pytest.importorskip(
 )
 
 
+@pytest.mark.skipif(
+    not hasattr(glass, "iternorm"),
+    reason="glass.iternorm not implemented",
+)
 @pytest.mark.stable
 def test_iternorm_no_size(
     benchmark: BenchmarkFixture,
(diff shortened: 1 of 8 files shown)


tests/regression/test_fields.py
connoraird, line 42: Does this need to be `glass_iternorm`? 

```suggestion
        generator = glass_iternorm(array_in)
```

Same question for the rest of the file?

connoraird, line 114:
```suggestion
        generator = glass_cls2cov(
```

connoraird (changes requested):
I think there's a few places you need to do `glass_*` instead of `glass.*`

tests/regression/test_fields.py
paddyroddy: No? We're not doing `pytest.importorskip`

paddyroddy: Same here

connoraird (approved):

PR #1204, opened 2026-09-16
gh-1203: Bump `array-api-extra` to `v0.11.1`

### LINKED ISSUE #1203: Bump to `array-api-extra` v0.11.1
### Add your issue here

https://github.com/data-apis/array-api-extra/releases#release-v0.11.1

paddyroddy:
# Description

This allows us to benefit from `deg2rad`, `rad2deg`, and `tril_indices`.

Closes: #1203

## Changelog entry

Changed: Bumped to `array-api-extra==0.11.1`.

## Checks


glass/lensing.py
```diff
@@ -762,4 +762,4 @@ def deflect(
 
     d = xp.atan2(sa * sg, st * ca - ct * sa * cg)
 
-    return lon - uxpx.degrees(d), uxpx.degrees(tp)
+    return lon - xpx.rad2deg(d), typing.cast("FloatArray", xpx.rad2deg(tp))
```
connoraird, line 765: Why do we need to type cast for only one of the returned values?

connoraird (approved):
Nice!, just one question

glass/lensing.py
paddyroddy: So my understanding is that [rad2deg outputs an internal `Array` type](https://github.com/data-apis/array-api-extra/blob/194545d2e2133db35702b9ebb7727fb0df4af31e/src/array_api_extra/_elementwise.py#L57). That conflicts we're out own `FloatArray` type. I think with `lon` we get `FloatArray - Array` which must resolve to `FloatArray` (the [internal `Array` type is `object`](https://github.com/data-apis/array-api-extra/blob/194545d2e2133db35702b9ebb7727fb0df4af31e/src/array_api_extra/_lib/_typing.py#L7)).

